PR #4, opened 2017-08-24
Handle Prometheus config with secrets

AlekSi:
Unfortunately, I had to copy a file from Prometheus and then patch it to remove special handling of secrets (passwords, etc.)

@arvenil @percona-csalguero Can you think of a better way?

@vadimtk Can we do it this way from the legal point of view: https://github.com/percona/pmm-managed/pull/4/files#diff-da7d8cce99d3a7aa636ac4f714f624fe ?


vadimtk:
I see you kept the original license too. I think it should be sufficient from licensing standpoint

PR #6, opened 2017-09-12
[WIP] PMM-1333 Added telemetry service

percona-csalguero:
Still needed:
1) If the config file for the telemetry service doesn't exists, the config uses the default values and generates an UUID. To be able to use the same UUID even if pmm-managed is restarted, we need to set a default location to save/retrieve the config.
2) For the build script (don't how this is being built) we need to add params to automatically update the version in the source code, based on git tags.


testdata/telemetry/config.yaml
```diff
@@ -0,0 +1,3 @@
+url: "http://v.percona.com"
+interval: 86400
+uuid: ""
```
AlekSi: You use config file so you can store UUID there?

cmd/pmm-managed/main.go
```diff
@@ -237,6 +240,48 @@ func runDebugServer(ctx context.Context) {
 	cancel()
 }
 
+func runTelemetryService(ctx context.Context) {
+	l := logrus.WithField("component", "TELEMETRY")
+
+	disabled := strings.ToLower(os.Getenv("DISABLE_TELEMETRY"))
+	if disabled == "true" || disabled == "1" {
```
AlekSi: `strconv.ParseBool()`

cmd/pmm-managed/main.go
```diff
@@ -237,6 +240,48 @@ func runDebugServer(ctx context.Context) {
...
+		l.Infof("Telemetry is disabled by DISABLE_TELEMETRY env var")
+		return
+	}
+
+	cfg, _ := telemetry.LoadConfig(*telemetryConfigF)
+
+	updated, err := telemetry.CheckConfig(cfg)
```
AlekSi: If LoadConfig can't load config and return `nil` for cfg, that line will panic. Let's not ignore LoadConfig error. Or combine LoadConfig and CheckConfig into single function – do we need to load invalid configs for something?

cmd/pmm-managed/main.go
```diff
@@ -237,6 +240,48 @@ func runDebugServer(ctx context.Context) {
...
+	// Using this env var for compatibility with the Toolkit
+	if telemetryEnvURL := os.Getenv("PERCONA_VERSION_CHECK_URL"); telemetryEnvURL != "" {
+		l.Infof("PERCONA_VERSION_CHECK_URL env var is set")
+		l.Infof("Using %s as the telemetry endpoint", telemetryEnvURL)
+		cfg.URL = telemetryEnvURL
+	}
+
+	svc, err := telemetry.NewService(cfg)
+	if err != nil {
+		l.Warnf("Cannot start telemetry: %s", err.Error())
```
AlekSi: No need to call `Error()` explicitly.

cmd/pmm-managed/main.go
```diff
@@ -237,6 +240,48 @@ func runDebugServer(ctx context.Context) {
...
+		return
+	}
+
+	l.Infof("Telemetry is enabled. Send data interval = %s seconds", cfg.Interval)
```
AlekSi: cfg.Interval is `int`, so `%s` will give you `%!s(int=100500)`. I would prefer to define `Interval` as time.Duration. yaml package should be able to work with it. Alternatively, change `%s` to `%d`.

cmd/pmm-managed/main.go
```diff
@@ -237,6 +240,48 @@ func runDebugServer(ctx context.Context) {
...
+	svc.Start()
+
+	<-ctx.Done()
+	svc.Stop()
+
```
AlekSi: Please remove that empty line.

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+				if err != nil {
+					s.lastStatus = fmt.Sprintf("%v error sending telemetry info: %s", time.Now(), err.Error())
+					continue
+				}
+				s.lastStatus = fmt.Sprintf("%v telemetry data sent.", time.Now())
+			case <-s.stopChan:
+				break LOOP
+			}
+		}
+		s.wg.Done()
```
AlekSi: Please move that to `defer`, so panics will invoke `Done()` too.

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+					continue
+				}
+				err = s.sendRequest(payload)
+				if err != nil {
+					s.lastStatus = fmt.Sprintf("%v error sending telemetry info: %s", time.Now(), err.Error())
+					continue
+				}
+				s.lastStatus = fmt.Sprintf("%v telemetry data sent.", time.Now())
+			case <-s.stopChan:
+				break LOOP
```
AlekSi: Once `s.wg.Done()` is moved to `defer`, we can just `return` there, and drop label.

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+				if err != nil {
+					s.lastStatus = fmt.Sprintf("%v cannot build payload for telemetry info: %s", time.Now(), err.Error())
+					continue
+				}
+				err = s.sendRequest(payload)
+				if err != nil {
+					s.lastStatus = fmt.Sprintf("%v error sending telemetry info: %s", time.Now(), err.Error())
+					continue
+				}
+				s.lastStatus = fmt.Sprintf("%v telemetry data sent.", time.Now())
```
AlekSi: Our logger should have a timestamp for each message already.

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+
+// Start sets a new ticker to collect data every Config.Interval seconds.
+func (s *Service) Start() {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if s.isRunning {
+		return
+	}
+	s.stopChan = make(chan bool)
+	s.isRunning = true
```
AlekSi: A lot of code there can be simplified with context. Something like:
```go
// Run runs Call home service until context is done.
func (s *Service) Run(ctx context.Context) {
  for {
    select {
      // select on ticker and ctx.Done()
    }

    // do work
  }
}
```

Chans, bool var, `Stop()` method can be dropped.

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+		data["OS"] = osType
+	}
+
+	data["PMM"] = Version
+
+	return data, nil
+}
+
+func (s *Service) makePayload(data map[string]interface{}) ([]byte, error) {
+	w := bytes.NewBuffer(nil)
```
AlekSi: `var w bytes.Buffer` is the preferred style.

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (s *Service) sendRequest(data []byte) error {
+	client := &http.Client{}
+	body := bytes.NewReader(data)
+
+	req, err := http.NewRequest("POST", s.Config.URL, body)
+	if err != nil {
+		return err
+	}
+	req.Header.Add("X-Percona-Toolkit-Tool", "pmm")
```
AlekSi: Should we also set `Content-Type`?

services/telemetry/base.go
```diff
@@ -0,0 +1,154 @@
...
+	resp, err := client.Do(req)
+	defer resp.Body.Close()
```
AlekSi: It is possible for `Do()` to return `(nil, err)` for very broken conditions. Check for `err != nil` first.

services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,265 @@
...
+	service.Stop()
+	isRunning = service.IsRunning()
+	assert.Equal(t, isRunning, false)
+}
+
+func TestServiceIntegration(t *testing.T) {
+
+	integrationTests := os.Getenv("INTEGRATION_TESTS")
+	if integrationTests == "" {
+		t.Skipf("Env var INTEGRATION_TESTS is not set. Skipping integration test")
```
AlekSi: Should we enable it on Travis CI?

AlekSi: Ok, nevermind.

services/telemetry/base.go
percona-csalguero: No need, but I can add it just for clarification

services/telemetry/base.go
percona-csalguero: I can change it if you want but I prefer to have a way to manually stop the service and to acknowledge it was really stopped.

services/telemetry/base.go
percona-csalguero: Yes, but this message is not being logged. It is stored internally for the GetLastStatus and it is used for testing

cmd/pmm-managed/main.go
```diff
@@ -237,6 +242,65 @@ func runDebugServer(ctx context.Context) {
...
+	l.Infof("Telemetry is enabled. Send data interval = %v", cfg.Interval)
+	svc.Start()
+
+	<-ctx.Done()
+	svc.Stop()
+}
+
+func getTelemetryUUID(consulAddress, defaultUUID string) (string, error) {
+	client, err := consulapi.NewClient(&consulapi.Config{Address: consulAddress})
```
AlekSi: I added consul service in `master`. Accept it as a parameter and use it, so we don't have several clients/connections in various places.

cmd/pmm-managed/main_test.go
```diff
@@ -0,0 +1,42 @@
+package main
```
AlekSi: Please add copyright header.

cmd/pmm-managed/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+
+	"golang.org/x/net/context"
+)
+
+func TestRunTelemetry(t *testing.T) {
+
```
AlekSi: Empty line.

cmd/pmm-managed/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+		break
+	case <-time.After(2100 * time.Millisecond):
+		t.Fatalf("Service didn't stop upon context cancellation")
+	}
+
+	w.Flush()
+	assert.Contains(t, b.String(), `msg="Telemetry is enabled. Send data interval = 24h0m0s" component=TELEMETRY`)
+
+	logrus.SetOutput(os.Stdout)
+
```
AlekSi: Empty line.

services/telemetry/base.go
```diff
@@ -0,0 +1,168 @@
+package telemetry
```
AlekSi: Copyright header.

services/telemetry/base.go
AlekSi: Context with sync.WaitGroup allows you to exactly that with less code and with the same interface – cancel the context to stop a running thing.

cmd/pmm-managed/main.go
```diff
@@ -237,6 +242,65 @@ func runDebugServer(ctx context.Context) {
 	cancel()
 }
 
+func runTelemetryService(ctx context.Context) {
+	l := logrus.WithField("component", "TELEMETRY")
+
+	disabled := strings.TrimSpace(strings.ToLower(os.Getenv("DISABLE_TELEMETRY")))
+	if disabled == "true" || disabled == "1" || disabled == "yes" {
```
AlekSi: Seriously, `strconv.ParseBool()`.

cmd/pmm-managed/main.go
```diff
@@ -34,15 +34,18 @@ import (
...
 	"github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/grpclog"
 
+	consulapi "github.com/hashicorp/consul/api"
```
AlekSi: I propose to group imports like this:
```go
import (
  // stdlib

  // third-party (including github.com/percona/something)

  // packages from that repository
)
```

So that import should go up.

services/telemetry/base.go
AlekSi: :ok:

services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,263 @@
+package telemetry
```
AlekSi: Copyright header.

services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+	time.Sleep(2100 * time.Millisecond)
+	assert.Equal(t, count, 3)
+	service.Stop()
+	isRunning = service.IsRunning()
+	assert.Equal(t, isRunning, false)
+}
+
+func TestServiceIntegration(t *testing.T) {
+
```
AlekSi: Empty line.

services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,263 @@
...
+	service, err := NewService(cfg)
+	require.NoError(t, err)
+	service.Start()
+	isRunning := service.IsRunning()
+	assert.Equal(t, isRunning, true)
+
+	time.Sleep(2100 * time.Millisecond)
+	assert.Contains(t, service.GetLastStatus(), "telemetry data sent")
+
```
AlekSi: Empty line.

services/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,263 @@
...
+		"pmm": "1.2.3",
+	}
+
+	b, err := svc.makePayload(m)
+	require.NoError(t, err)
+	// Don't remove \n at the end of the strings. They are needed by the API
+	// so I want to ensure makePayload adds them
+	assert.Contains(t, string(b), "ABCDEFG12345;OS;Kubuntu\n")
+	assert.Contains(t, string(b), "ABCDEFG12345;pmm;1.2.3\n")
+
```
AlekSi: Empty line.

services/telemetry/utils.go
```diff
@@ -0,0 +1,132 @@
...
+		return "", err
+	}
+
+	// Old telemetry IDs have only 32 chars in the table but UUIDs + "-" = 36
+	cleanUUID := strings.Replace(uuid.String(), "-", "", -1)
+	return cleanUUID, nil
+}
+
+func getOSNameAndVersion() (string, error) {
+
```
AlekSi: Empty line.

cmd/pmm-managed/main.go
percona-csalguero: I know

PR #7, opened 2017-09-12
PMM-1441 packaging needs

delgod:


AlekSi:
Please move packaging to https://github.com/percona/pmm-server-packaging. Pass `-prometheus-config` with the right value explicitly in unit.

PR #9, opened 2017-09-25
PMM-1507 Added OS detection from inside a docker container

percona-csalguero:


AlekSi:
It will not work if `dmesg` buffer was filled and init messages were overwritten.

AlekSi:
Closed in favor of #10.

PR #10, opened 2017-09-28
PMM-1507: Distribution detection

AlekSi:
That's for 1.3.2. Do not merge yet.


services/telemetry/base.go
```diff
@@ -106,3 +156,48 @@ func (s *Service) sendRequest(ctx context.Context, data []byte) error {
...
+// For both AMI and OVA this value is fixed by the environment variable and not autodetected.
+// If/when we decide to support installation with "normal" Linux package managers (apt, yum, etc.),
+// we could use the code that was there. See PM-1333 and PMM-1507 in both git logs and Jira for details.
+
+type pair struct {
+	re *regexp.Regexp
+	t  string
+}
+
+var procVersionRegexps = []pair{
```
delgod: rules can become outdated or improved. in this case, we receive situation when old PMM version reports the wrong OS.
I believe it is better to pass full output as is to our server and analyze it centrally because in this case we always will have right analyze.
Even more, OS version is very important for RedHat/Debian distributions.
if we can analyze this string on our side, we can have kernel->OS version matching.

AlekSi: > I believe it is better to pass full output as is to our server and analyze it
Sure, we will do it in the next version. Current server API is too limited for that.

PR #29, opened 2017-12-22
PMM-1729 Run RDS exporter

AlekSi:


cmd/pmm-managed/main.go
```diff
@@ -82,14 +82,17 @@ var (
 	dbUsernameF = flag.String("db-username", "pmm-managed", "Database username")
 	dbPasswordF = flag.String("db-password", "pmm-managed", "Database password")
 
-	agentMySQLdExporterF = flag.String("agent-mysqld-exporter", "/usr/local/percona/pmm-client/mysqld_exporter", "mysqld_exporter path")
-	agentQANBaseF        = flag.String("agent-qan-base", "/usr/local/percona/qan-agent", "qan-agent installation base path")
+	agentMySQLdExporterF    = flag.String("agent-mysqld-exporter", "/usr/local/percona/pmm-client/mysqld_exporter", "mysqld_exporter path")
+	agentRDSExporterF       = flag.String("agent-rds-exporter", "/usr/sbin/rds_exporter", "rds_exporter path")
```
arvenil: Why rds_exporter is under `/usr/sbin` but mysqld_exporter under `/usr/local/percona/pmm-client`?

services/rds/rds_exporter.go
```diff
@@ -0,0 +1,42 @@
...
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+package rds
+
+import (
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+type rdsExporterInstance struct {
```
arvenil: Why not embed https://github.com/percona/rds_exporter/blob/master/config/config.go#L10-L15 ?

services/rds/rds.go
```diff
@@ -149,6 +151,18 @@ func (svc *Service) ApplyPrometheusConfiguration(ctx context.Context, q *reform.
...
+		ScrapeInterval: "60s",
+		ScrapeTimeout:  "55s",
+		MetricsPath:    "/basic",
+	}
+	rdsEnhanced := &prometheus.ScrapeConfig{
+		JobName:        "rds-enhanced",
+		ScrapeInterval: "10s",
+		ScrapeTimeout:  "9s",
+		MetricsPath:    "/enhanced",
+	}
```
arvenil: Shouldn't this be in some configuration file?

cmd/pmm-managed/main.go
AlekSi: rds_exporter is not a part of pmm-client distribution package.

arvenil: This still doesn't explain why we are not pushing it under the same dir. Not that this is a big issue anyway.

services/rds/rds_exporter.go
AlekSi: It is simple enough to avoid an extra dependency.

services/rds/rds.go
AlekSi: An interval for basic metrics is fixed by AWS, so no configuration option is necessary.
Interval for enhanced metrics should be auto-adjustable, but that's another issue.

PR #34, opened 2018-01-29
PMM-1712 Improve external exporters

AlekSi:


services/prometheus/scrape_configs.go
```diff
@@ -97,18 +132,200 @@ func (svc *Service) putToConsul(scs []ScrapeConfig) error {
...
+// reachabilityCh is closed when this method returns.
+func (svc *Service) checkReachability(ctx context.Context, cfg *ScrapeConfig, targets []string, reachabilityCh chan<- ScrapeTargetReachability) {
+	// use ephemeral transport with small timeouts and disabled HTTP keep-alive
+	transport := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   5 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		TLSHandshakeTimeout: 5 * time.Second,
```
arvenil: The lower timeouts don't work?

services/prometheus/scrape_configs.go
```diff
@@ -97,18 +132,200 @@ func (svc *Service) putToConsul(scs []ScrapeConfig) error {
...
+func (svc *Service) getTargetsHealth(ctx context.Context) (map[string]map[string]Health, error) {
+	u := *svc.baseURL
+	u.Path = path.Join(u.Path, "api/v1/targets")
+	resp, err := svc.client.Get(u.String())
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	defer resp.Body.Close()
+
+	// Copied from vendor/github.com/prometheus/prometheus/web/api/v1/api.go to avoid a ton of dependencies
```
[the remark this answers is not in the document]
AlekSi: This package depends on a ton of service discovery mechanism. Kubernetes alone is a _huge_ dependency and a very tricky to update.

services/prometheus/scrape_configs.go
AlekSi: They should, I'm just being too cautious there.

arvenil: You can always try lower first. Especially it's a health-check. I would also add it as global, easy to modify and avoids duplicate timeout definitions.

AlekSi: But it is not duplicated anywhere.

PR #36, opened 2018-02-26
PMM-2155 Honor instance labels for RDS.

AlekSi:
Always use `instance` label values from exporters. See `honor_labels` [here](https://prometheus.io/docs/prometheus/1.8/configuration/configuration/#<scrape_config>).


AlekSi:
Added link to the description.

PR #57, opened 2018-09-25
PMM-2839 Add tablestats safety precaution.

AlekSi:
Do not pass -collect.info_schema.tablestats
to mysqld_exporter if there are more than 1000
tables.

### DIFF
--- a/services/rds/rds.go
+++ b/services/rds/rds.go
@@ -399,12 +399,14 @@ func (svc *Service) addMySQLdExporter(ctx context.Context, tx *reform.TX, servic
 		return errors.WithStack(err)
 	}
 
-	// check connection
+	// check connection and a number of tables
+	var tableCount int
 	dsn := agent.DSN(service)
 	db, err := sql.Open("mysql", dsn)
 	if err == nil {
-		err = db.PingContext(ctx)
+		err = db.QueryRowContext(ctx, "SELECT COUNT(*) FROM information_schema.tables").Scan(&tableCount)
 		db.Close()
+		agent.MySQLDisableTablestats = pointer.ToBool(tableCount > 1000)
 	}
 	if err != nil {
 		if err, ok := err.(*mysql.MySQLError); ok {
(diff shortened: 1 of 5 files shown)


services/rds/rds.go
```diff
@@ -399,12 +399,14 @@ func (svc *Service) addMySQLdExporter(ctx context.Context, tx *reform.TX, servic
...
-	// check connection
+	// check connection and a number of tables
+	var tableCount int
 	dsn := agent.DSN(service)
 	db, err := sql.Open("mysql", dsn)
 	if err == nil {
-		err = db.PingContext(ctx)
+		err = db.QueryRowContext(ctx, "SELECT COUNT(*) FROM information_schema.tables").Scan(&tableCount)
 		db.Close()
+		agent.MySQLDisableTablestats = pointer.ToBool(tableCount > 1000)
```
delgod: I don't know why we have so high limit, tablestat creates huge traffic and significant delays even on 200 tables.
anyway, feel free to merge!

AlekSi: That's the same limit as in pmm-admin.

Currently, we do not provide UI for setting this number, but we store disabling flag in the database when we add an instance, so our support can explain to a customer how to switch it.

arvenil:
@AlekSi My bad. This needs to be reworked. `tablestats` is an ugly trick (which I would love to get rid of). In pmm-admin `--disable-tablestats` actually disables a set of options, not just `-collect.info_schema.tablestats`.
https://github.com/percona/pmm-client/blob/dfe4f639f07bec01de8c87adfde8c8f1c8d7222f/pmm/plugin/mysql/metrics/metrics.go#L98-L105
```golang
"-collect.auto_increment.columns=",
"-collect.info_schema.tables=",
"-collect.info_schema.tablestats=",
"-collect.perf_schema.indexiowaits=",
"-collect.perf_schema.tableiowaits=",
"-collect.perf_schema.tablelocks=",
```

PR #61, opened 2018-10-09
PMM-3074 Draft API for MySQL and PostgreSQL.

BupycHuk:


api/mysql.proto
```diff
@@ -0,0 +1,92 @@
...
+
+message MySQLRemoveRequest {
+    MySQLInstanceID id = 1;
+}
+
+message MySQLRemoveResponse {
+}
+
+service MySQL {
+    rpc Discover(MySQLDiscoverRequest) returns (MySQLDiscoverResponse) {
```
AlekSi: We will not have that API in 1.16, let's drop it.

api/mysql.proto
```diff
@@ -0,0 +1,92 @@
...
+        option (google.api.http) = {
+            post: "/v0/mysql/discover"
+            body: "*"
+        };
+    }
+
+    rpc List(MySQLListRequest) returns (MySQLListResponse) {
+        option (google.api.http) = {
+            get: "/v0/mysql"
```
AlekSi: Let's use POST with a body for all new APIs, and not try to emulate REST. I will explain why in Slack.

So, something like `/v0/mysql/list`.

api/mysql.proto
```diff
@@ -0,0 +1,92 @@
...
+    rpc Add(MySQLAddRequest) returns (MySQLAddResponse) {
+        option (google.api.http) = {
+            post: "/v0/mysql"
+            body: "*"
+        };
+    }
+
+    rpc Remove(MySQLRemoveRequest) returns (MySQLRemoveResponse) {
+        option (google.api.http) = {
+            delete: "/v0/mysql"
```
AlekSi: POST with a body. `/v0/mysql/remove`.

api/mysql.proto
```diff
@@ -0,0 +1,92 @@
...
+
+    rpc List(MySQLListRequest) returns (MySQLListResponse) {
+        option (google.api.http) = {
+            get: "/v0/mysql"
+        };
+    }
+
+    rpc Add(MySQLAddRequest) returns (MySQLAddResponse) {
+        option (google.api.http) = {
+            post: "/v0/mysql"
```
AlekSi: `/v0/mysql/add`

api/postgresql.proto
```diff
@@ -0,0 +1,92 @@
...
+        };
+    }
+
+    rpc Remove(PostgreSQLRemoveRequest) returns (PostgreSQLRemoveResponse) {
+        option (google.api.http) = {
+            delete: "/v0/postgresql"
+            body: "*"
+        };
+    }
+}
```
AlekSi: Please do the same as for MySQL.

AlekSi:
Please click the button at https://github.com/percona/pmm-managed/pull/61#issuecomment-428146450.

api/mysql.proto
```diff
@@ -0,0 +1,77 @@
...
+
+message MySQLListRequest {
+}
+
+message MySQLListResponse {
+    repeated MySQLInstance instances = 1;
+}
+
+message MySQLAddRequest {
+    MySQLInstanceID id = 1;
```
AlekSi: We will need host and port, not region and name.

api/mysql.proto
```diff
@@ -0,0 +1,77 @@
...
+    string address = 4;
+    uint32 port = 5;
+    string engine = 6;
+    string engine_version = 7;
+}
+
+message MySQLInstanceID {
+    string region = 1;
+    string name = 2;
+}
```
AlekSi: That's RDS-specific.

api/mysql.proto
```diff
@@ -0,0 +1,77 @@
...
+
+package api;
+
+import "google/api/annotations.proto";
+
+message MySQLNode {
+    reserved 1, 2; // id and type
+
+    string region = 3;
+    string name = 4;
```
AlekSi: What are the region and name there?

api/mysql.proto
```diff
@@ -0,0 +1,77 @@
...
+
+message MySQLRemoveRequest {
+    MySQLInstanceID id = 1;
+}
+
+message MySQLRemoveResponse {
+}
+
+service MySQL {
+
```
AlekSi: Remove empty line.

api/mysql.proto
BupycHuk: we have host and port in `MySQLService`, should we duplicate it here?

api/mysql.proto
BupycHuk: got it

api/mysql.proto
```diff
@@ -0,0 +1,73 @@
...
+message MySQLAddRequest {
+    string address = 1;
+    uint32 port = 2;
+    string username = 3;
+    string password = 4;
+}
+
+message MySQLAddResponse {
+}
```
AlekSi: Let's return id.

api/mysql.proto
```diff
@@ -0,0 +1,73 @@
...
+
+message MySQLRemoveRequest {
+    string address = 1;
+    uint32 port = 2;
```
AlekSi: Let's use id instead.

api/mysql.proto
AlekSi: ^

api/mysql.proto
```diff
@@ -0,0 +1,73 @@
...
+        option (google.api.http) = {
+            post: "/v0/mysql"
+            body: "*"
+        };
+    }
+
+    rpc Remove(MySQLRemoveRequest) returns (MySQLRemoveResponse) {
+        option (google.api.http) = {
+            delete: "/v0/mysql"
+            body: "*"
```
AlekSi: Let's add id to the URL and drop body. Generator support that natively.

api/mysql.proto
BupycHuk: ok

api/mysql.proto
BupycHuk: ok

api/mysql.proto
BupycHuk: ok

PR #69, opened 2018-10-24
PMM-3061 Set default port

BupycHuk:
Set default port for MySQL and PostgreSQL if port is not presented in Add request.


services/postgresql/postgresql.go
```diff
@@ -75,6 +77,7 @@ func NewService(config *ServiceConfig) (*Service, error) {
 	} {
 		if *path == "" {
 			continue
+			const defaultPostgresPort = 5432
```
golangcibot: unreachable code

PR #75, opened 2018-10-30
PMM-3189 Add agentless logs into logs.zip

AlekSi:


models/agent.go
```diff
@@ -48,6 +48,14 @@ type Agent struct {
 	ID           int32     `reform:"id,pk"`
 	Type         AgentType `reform:"type"`
 	RunsOnNodeID int32     `reform:"runs_on_node_id"`
+
+	// TODO Does it really belong there? Remove when we have agent without one.
+	ListenPort *uint16 `reform:"listen_port"`
+}
+
+// NameForSupervisor is a name of agent for supervisor.
+func (a *Agent) NameForSupervisor() string {
```
BupycHuk: maybe it's better to use a function instead of a method?
```suggestion
func NameForSupervisor(type, port string) string {
```

services/logs/logs.go
```diff
@@ -164,54 +172,84 @@ func New(pmmVersion string, consul *consul.Client, rds *rds.Service, logs []Log)
...
-			if len(content) > 0 {
-				content = append(content, "\n\n"...)
+			if len(file.Data) > 0 {
+				file.Data = append(file.Data, "\n\n"...)
 			}
-			content = append(content, []byte(err.Error())...)
+			file.Data = append(file.Data, []byte(file.Err.Error())...)
+		}
+
+		if file.Name == "" {
```
BupycHuk: maybe it should be placed before `if file.Err != nil {` line ?

BupycHuk: because if file.Name == "" above operations don't make sense.

models/agent.go
AlekSi: Good point!

services/logs/logs.go
AlekSi: Actually, this line no longer has any sense. :)

AlekSi: Well, this is one corner case when it is needed…

services/mysql/mysql.go
```diff
@@ -283,8 +283,8 @@ func (svc *Service) addMySQLdExporter(ctx context.Context, tx *reform.TX, servic
 	return nil
 }
 
-func (svc *Service) mysqlExporterCfg(agent *models.MySQLdExporter, port uint16, dsn string) *servicelib.Config {
-	name := agent.NameForSupervisor()
+func (svc *Service) mysqlExporterCfg(agent *models.MySQLdExporter, dsn string) *servicelib.Config {
```
golangcibot: 286-322 lines are duplicate of `services/rds/rds.go:443-479`

services/rds/rds.go
```diff
@@ -440,8 +440,8 @@ func (svc *Service) addMySQLdExporter(ctx context.Context, tx *reform.TX, servic
 	return nil
 }
 
-func (svc *Service) mysqlExporterCfg(agent *models.MySQLdExporter, port uint16, dsn string) *servicelib.Config {
-	name := agent.NameForSupervisor()
+func (svc *Service) mysqlExporterCfg(agent *models.MySQLdExporter, dsn string) *servicelib.Config {
```
golangcibot: 443-479 lines are duplicate of `services/mysql/mysql.go:286-322`

PR #82, opened 2018-11-12
PMM-3158 Async protocol between pmm agent and pmm managed

BupycHuk:


services/agents/message_dispatcher.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"sync"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/percona/pmm/api/agent"
+)
+
+type messageDispatcher struct {
+	sync.RWMutex
```
AlekSi: Never embed `sync.Mutex` or `sync.RWMutex` – that makes `Lock()` / `Unlock()` methods a part of exported API.
```go
rw sync.RWMutex
```

services/agents/conn.go
```diff
@@ -29,16 +28,21 @@ import (
...
+	lastID            uint32
+	l                 *logrus.Entry
+	messageDispatcher *messageDispatcher
 }
 
 func NewConn(uuid string, stream agent.Agent_ConnectServer) *Conn {
-	return &Conn{
-		stream: stream,
-		l:      logrus.WithField("pmm-agent", uuid),
+	// Create goroutine to dispatch messages
```
AlekSi: We have two common styles for internal (i.e. not for method documentation, etc.) commands:
1.
```
sentence fragment, starting with a lower-case letter and not ending with a dot
```
2.
```
Several full sentences. Each starts with an upper-case letter. Each ends with a dot.
```

Please use the first one there.

services/agents/conn.go
```diff
@@ -52,19 +56,28 @@ func (c *Conn) SendAndRecv(toAgent agent.ServerMessagePayload) (*agent.AgentMess
...
+	agentMessage := c.messageDispatcher.WaitForMessage(id)
 	c.l.Debugf("Recv: %s.", agentMessage)
-	if agentMessage.Id != id {
-		return nil, status.Errorf(codes.Internal, fmt.Sprintf("expected message from agent with ID %d, got ID %d", id, agentMessage.Id))
-	}
+
 	return agentMessage, nil
 }
+func (c *Conn) startMessageDispatcher() {
+	for {
```
AlekSi:
```go
for c.stream.Context().Err() != nil
```

And then drop `select` below.

services/agents/conn.go
```diff
@@ -52,19 +56,28 @@ func (c *Conn) SendAndRecv(toAgent agent.ServerMessagePayload) (*agent.AgentMess
...
+		select {
+		case <-c.stream.Context().Done():
+			c.l.Debugln("Close connection: ", c.lastID)
+			return
+		default:
+			agentMessage, err := c.stream.Recv()
+			if err != nil {
+				errorStatus, ok := status.FromError(err)
+				if ok && errorStatus.Code() == codes.Canceled {
```
AlekSi: That check is too fragile. I think the error will be different if you `kill -9` or power cycle the peer.
What I propose:
* log warning for any error;
* we should signal the user of that type that connection is closed.

services/agents/conn.go
```diff
@@ -52,19 +56,28 @@ func (c *Conn) SendAndRecv(toAgent agent.ServerMessagePayload) (*agent.AgentMess
...
+					c.l.Debugln("Connection closed from other side")
+					return
+				}
+				c.l.Fatal(errors.Wrap(err, "failed to receive message from agent"))
```
AlekSi: That's a totally expected error, not a fatal one.

services/agents/message_dispatcher.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"sync"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/percona/pmm/api/agent"
+)
+
+type messageDispatcher struct {
```
AlekSi: This type handles only responses, so maybe `responseDispatcher`?

And do we need it to be a separate type? Let's see if we can combine it with `Conn` after we do pmm-agent part.

services/agents/message_dispatcher.go
```diff
@@ -0,0 +1,38 @@
...
+	return &messageDispatcher{
+		subscribers: make(map[uint32]chan *agent.AgentMessage),
+		l:           logrus.WithField("message dispatcher", uuid),
+	}
+}
+
+func (m *messageDispatcher) MessageReceived(message *agent.AgentMessage) {
+	m.Lock()
+	defer m.Unlock()
+	m.subscribers[message.Id] <- message
```
AlekSi: We should gracefully handle unexpected messages, not panic with sending to `nil`.

services/agents/message_dispatcher.go
```diff
@@ -0,0 +1,38 @@
...
+	delete(m.subscribers, message.Id)
+}
+
+func (m *messageDispatcher) WaitForMessage(id uint32) *agent.AgentMessage {
+	m.Lock()
+	agentChan := make(chan *agent.AgentMessage)
+	defer close(agentChan)
+	m.subscribers[id] = agentChan
```
AlekSi: We should handle a situation when there is already a subscriber for that ID.

services/agents/conn.go
```diff
@@ -52,19 +56,61 @@ func (c *Conn) SendAndRecv(toAgent agent.ServerMessagePayload) (*agent.AgentMess
...
+	}
+}
+
+func (c *Conn) emit(message *agent.AgentMessage) {
+	c.rw.Lock()
+	defer c.rw.Unlock()
+	if _, ok := c.subscribers[message.Id]; ok {
+		for _, subscriber := range c.subscribers[message.Id] {
+			go func() {
+				subscriber <- message
```
golangcibot: loop variable subscriber captured by func literal

AlekSi: It is not a "possible" race condition – it is a very real one: https://golang.org/doc/faq#closures_and_goroutines
Please add a unit test.

BupycHuk: I already fixed it

PR #90, opened 2018-12-04
PMM-3158 fix auth response

BupycHuk:
Added id to auth response


AlekSi:
We are dropping that message, see https://github.com/percona/pmm-agent/pull/8.

PR #92, opened 2018-12-10
PMM-3314 Async protocol

AlekSi:


services/agents/channel_test.go
```diff
@@ -16,48 +16,47 @@
...
-}
-
 func (s *testServer) Connect(stream agent.Agent_ConnectServer) error {
 	return s.connect(stream)
 }
 
 var _ agent.AgentServer = (*testServer)(nil)
 
-func setup(t *testing.T, connect func(agent.Agent_ConnectServer) error, expected ...error) (*Channel, *grpc.ClientConn, func(*testing.T)) {
+func setup(t *testing.T, connect func(*Channel) error, expected ...error) (agent.Agent_ConnectClient, *grpc.ClientConn, func(*testing.T)) {
```
golangcibot: `setup` - `expected` is unused (from `unparam`)

AlekSi:
@BupycHuk PTAL

services/agents/channel_test.go
```diff
@@ -76,188 +80,277 @@ func setup(t *testing.T, connect func(agent.Agent_ConnectServer) error, expected
...
-		// this message triggers "no subscriber for ID" error
-		err := stream.Send(&agent.ServerMessage{
-			Id: 111,
-			Payload: &agent.ServerMessage_QanData{
-				QanData: new(agent.QANDataResponse),
-			},
+func TestUnexpectedResponseFromAgent(t *testing.T) {
+	connect := func(ch *Channel) error { //nolint:unparam
+		// after receiving unexpected response, channel is closed
+		msg := ch.SendRequest(&agent.ServerMessage_Ping{
```
BupycHuk: why don't we return error here?

services/agents/channel_test.go
```diff
@@ -76,188 +80,277 @@ func setup(t *testing.T, connect func(agent.Agent_ConnectServer) error, expected
...
+		msg, err := stream.Recv()
+		require.NoError(t, err)
+		assert.Equal(t, i, msg.Id)
+		assert.NotNil(t, msg.GetQanData())
 	}
+
+	err := stream.CloseSend()
+	assert.NoError(t, err)
+
+	// check metrics
```
BupycHuk: maybe it's better to check metrics in separate test?

services/agents/channel_test.go
AlekSi: Two reasons. First, we can't really give a user a useful error there: was it send error? was it channel error, so message _maybe_ was sent, but not handled? or maybe it was handled, but we were not able to get a response? Second, we still need users to `Requests()` until it is closed, and not break earlier on error.

services/agents/channel_test.go
AlekSi: Maybe. I don't know yet. Probably when we will have more metrics.

PR #120, opened 2019-03-05
PMM-3620 Add NOT NULL where possible.

AlekSi:
After #119.


models/database.go
```diff
@@ -71,7 +71,22 @@ var databaseSchema = [][]string{
 			UNIQUE (address, region)
```
AlekSi: @DexterHD That's the biggest reason to use NULLs. We want data integrity on the database level (among other reasons, our users are database experts, so they _will_ try to use SQL instead of APIs 😆). And indexes in both MySQL and PostgreSQL handle NULLs as non-comparable values, to unique indexes work the way we want them to work.

I'm not sure that adding NOT NULL only to several columns worth it. Thoughts?

models/database.go
```diff
@@ -46,23 +46,23 @@ var databaseSchema = [][]string{
...
-			docker_container_id VARCHAR(255),
-			docker_container_name VARCHAR(255),
+			docker_container_id VARCHAR(255), -- NULL means "unknown"; non-NULL value must be unique
+			docker_container_name VARCHAR(255) NOT NULL,
 
 			-- RemoteAmazonRDS
-			-- instance is address
-			region VARCHAR(255),
+			-- RDS instance is stored in address
+			region VARCHAR(255), -- NULL means "not Remote"; non-NULL value must be unique in combination with instance/address
```
DexterHD: @AlekSi Btw I think it will be better to add comments inside the schema so you can see this info inside Table Info
```suggestion
			region VARCHAR(255) COMMENT 'NULL means "not Remote"; non-NULL value must be unique in combination with instance/address',
```

models/database.go
DexterHD: Maybe we should use `NULL` fields only in these cases where we need indexes for example? 
In other cases we can use empty fields and have proper comments in schema about what does empty fields mean. How do you think?

AlekSi:
Please squash whem merging.

PR #133, opened 2019-03-12
PMM-1416 gRPC API Improvements

DexterHD:


DexterHD:
Closing it as we meet with @AlekSi and decided that it isn't a suitable solution for us.

PR #135, opened 2019-03-12
PMM-3672 Update .travis.yml

rnovikovP:


DexterHD:
@AlekSi Could you tell to @rnovikovP, what exactly should be fixed (Maybe files or at least something that can help him to finish this), as for now it sounds like "You do it wrong, no matter how, but go and do it right..." 😊😕I  can't understand what is wrong for example. I pulled this branch, did `make env-up` and saw Grafana was working. This PR is about that I think, doesn't it?

AlekSi:
I don't expect Roma to fix test configuration – that's not his job. I also don't understand why he keeps trying to submit the same change over and over again without any comments.

We need to update test configuration in `testdata/grafana`. grafana.ini contains things specific to the old Grafana that are silently ignored by the new version. prometheus.yaml contains old URL. And Prometheus.json should be updated too.

rnovikovP:
@AlekSi  the fack is - we already use  Grafana 6  for building The Product 
My idea for PR  was  =- to update where I see we used a wrong version number 
For now - it's  much more clear what should be our process for updating Versions  (@adivinho  FYI)
I expect the same is needed for updates of Prometheus  and  etc..

I can see `grafana.ini` file but looks like the options we have there  (just several of them) are working  well with v 6.0
So I would propose to merge this PR to let  everyone else  see what versions do we use but don't close  issue in Jira to update other files if we need to

AlekSi:
See #163.

PR #142, opened 2019-03-19
PMM-3698 Add empty username and password for mongodb_exporter.

DexterHD:


services/agents/mongodb_test.go
```diff
@@ -54,4 +54,14 @@ func TestMongodbExporterConfig(t *testing.T) {
 	assert.Equal(t, expected.Args, actual.Args)
 	assert.Equal(t, expected.Env, actual.Env)
 	assert.Equal(t, expected, actual)
+
+	t.Run("CheckEmptyPassword", func(t *testing.T) {
```
AlekSi: But password is not empty!

DexterHD: @AlekSi Oh. This is a typo. There is no reason to check an empty password... But there is a reason to check empty credentials. So if a username is empty, credentials string must be empty too.

AlekSi: Yeah, I get it. Just pointed out wrong test name.

PR #145, opened 2019-03-22
PMM-3470 Change APIs.

AlekSi:


Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 # TODO switch to tags before PMM 2.0 release
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "PMM-3470-change-apis"
```
AlekSi: Should be reverted.

PR #161, opened 2019-04-05
PMM-3798 Register Node with address.

AlekSi:


Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 # TODO switch to tags before PMM 2.0 release
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "PMM-3798-node-address"
```
AlekSi: Must be reverted.

PR #162, opened 2019-04-05
PMM-3791 Logs fixes.

BupycHuk:


main.go
```diff
@@ -110,7 +110,7 @@ func addLogsHandler(mux *http.ServeMux, logs *logs.Logs) {
 		ctx, cancel := context.WithTimeout(req.Context(), 10*time.Second)
 		defer cancel()
 
-		filename := fmt.Sprintf("pmm-server_%s", time.Now().UTC().Format("2006-01-02_15-04"))
+		filename := fmt.Sprintf("pmm-server_%s.zip", time.Now().UTC().Format("2006-01-02_15-04"))
```
AlekSi: 🤦‍♂️ I was an idiot.

services/logs/logs.go
```diff
@@ -75,8 +72,7 @@ var defaultLogs = map[string]logInfo{
 
 	// core PMM components
 	"pmm-managed.log": {"/srv/logs/pmm-managed.log", "pmm-managed"},
-	"qan-api.log":     {"/srv/logs/qan-api.log", ""},
-	"qan-api2.log":    {"/var/log/qan-api2.log", ""},
+	"qan-api2.log":    {"/srv/logs/qan-api2.log", ""},
```
AlekSi: Second parameter should not be empty. Please find systemd unit name for qan-api2 and use it.

AlekSi: https://github.com/percona/pmm-server-packaging/blob/PMM-2.0/rhel/SPECS/percona-qan-api2.spec I guess it is `percona-qan-api2`.

@EvgeniyPatlan @hors Why we use `percona-qan-api2`, but `pmm-managed` there: https://github.com/percona/pmm-server-packaging/blob/PMM-2.0/rhel/SPECS/pmm-managed.spec ? Could we be more consistent?

PR #166, opened 2019-04-12
PMM-3632 Fix bugs.

BupycHuk:
Fixed not exist pmm-agent ID bug.
Fixed filter for services.


models/agent_helpers.go
```diff
@@ -181,6 +181,12 @@ func AgentAddExporter(q *reform.Querier, agentType AgentType, params *AddExporte
 		return nil, err
 	}
 
+	if params.PMMAgentID != "" {
```
DexterHD: I think it's wrong, isn't it? If there is no pmm-agent id, you can't add exporter at all, as all exporters must be connected to pmm-agent. So this function should return an error when this parameter is empty.

BupycHuk: As I know there might be external exporter with no pmm-agent id

DexterHD: Yeah, but we don't have it now. Also, the logic of it (how it should be created) can be different. I think we need separate function for adding external_exporter as we have a different function `AgentAddPmmAgent()`. But let's ask @AlekSi what does he think.

DexterHD: When I was refactoring this code, first of all, I create different functions for each exporter and only after that I made     `AgentAddExporter` when I have realized that they all work the same way. As of external_exporter, I think it's better to implement that and then look is it different and can we refactor it.

BupycHuk: Yeah, you are right

models/service_helpers.go
```diff
@@ -88,6 +88,21 @@ func ServicesForAgent(q *reform.Querier, agentID string) ([]*Service, error) {
 	return res, nil
 }
 
+// ServicesForNode returns all Services for Node with given ID provides insights.
```
AlekSi: That comment makes no sense.

services/inventory/grpc/services_server.go
```diff
@@ -39,7 +39,10 @@ func NewServicesServer(s *inventory.ServicesService) inventorypb.ServicesServer
 
 // ListServices returns a list of all Services.
```
AlekSi: Not really all anymore.

models/service_helpers.go
```diff
@@ -88,6 +88,21 @@ func ServicesForAgent(q *reform.Querier, agentID string) ([]*Service, error) {
...
+func ServicesForNode(q *reform.Querier, nodeID string) ([]*Service, error) {
+	tail := fmt.Sprintf("WHERE node_id = %s ORDER BY service_id", q.Placeholder(1)) //nolint:gosec
```
AlekSi: btw, @DexterHD @BupycHuk: what do you think about switching to plain placeholders? Like:
```go
structs, err := q.SelectAllFrom(ServiceTable, "WHERE node_id = $1 ORDER BY service_id", nodeID)
```

We used `q.Placeholder(1)` because we knew we are going to switch away from MySQL to either SQLite or PostgreSQL, and they have different placeholder syntax. We switched, and very unlikely to switch again, so we can make things clearer.

services/inventory/grpc/services_server.go
```diff
@@ -39,7 +39,10 @@ func NewServicesServer(s *inventory.ServicesService) inventorypb.ServicesServer
...
 func (s *servicesServer) ListServices(ctx context.Context, req *inventorypb.ListServicesRequest) (*inventorypb.ListServicesResponse, error) {
-	services, err := s.s.List(ctx)
+	filters := inventory.ServiceFilters{
+		NodeID: req.GetNodeId(),
```
AlekSi: In general, prefer fields to methods. `reg.NodeId`. Methods are useful _sometimes_ (for chains where `nil` is possible), but also send a signal that _something is complected there_, and that place is a simple one.

services/inventory/services.go
```diff
@@ -41,13 +41,24 @@ func NewServicesService(db *reform.DB, r registry) *ServicesService {
 	}
 }
 
+// ServiceFilters represents filters for agents list.
```
AlekSi: Not agents.

services/inventory/services.go
```diff
@@ -41,13 +41,24 @@ func NewServicesService(db *reform.DB, r registry) *ServicesService {
...
+type ServiceFilters struct {
+	// Return only Services that provide insights for that Node.
```
AlekSi: "insights" is a term for Prometheus metrics and QAN data. In other word, that's for Agents, not Services and Nodes.

AlekSi:
Oh, and Travis CI failed. `vendor/` is not in sync.

models/agent_helpers.go
AlekSi: > I think we need separate function for adding external_exporter as we have a different function AgentAddPmmAgent()

+1
But I also think that `AgentAddExporter` is not a great name – QAN data providers (built-in Agents for performance schema and slow log) are not exporters, and they are not "special" like pmm-agent or external exporters. Anyway, let's not change that now.

Speaking of external exporters, I think we will have them in pmm-managed (database, models, services, Inventory APIs) in 2.0 _mostly_ to get the data model right. We will expose them in Management API (if at all) and in `pmm-admin` in 2.x.

services/inventory/grpc/services_server.go
BupycHuk: we can't use field here because in proto file we used `oneof`

AlekSi: ok

PR #167, opened 2019-04-16
PMM-3843 Fixed problem with permissions for node_exporter.

BupycHuk:


services/agents/node.go
```diff
@@ -56,7 +56,7 @@ func nodeExporterConfig(node *models.Node, exporter *models.Agent) *agentpb.SetS
 			"--collector.mountstats",
 			"--collector.processes",
 			"--collector.qdisc",
-			"--collector.systemd",
+			//"--collector.systemd", disabled for now
```
AlekSi: Please add a link to the ticket.

PR #178, opened 2019-04-29
PMM-3921 MongoDB Management API.

DexterHD:


services/management/mongodb.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/percona/pmm-managed/models"
+
+	// FIXME Refactor, as service shouldn't depend on other service in one abstraction level.
+	// https://jira.percona.com/browse/PMM-3541
+	// See also main_test.go
+	"github.com/percona/pmm-managed/services/inventory"
+)
+
+// MongoDBService MongoDB Management Service.
+type MongoDBService struct {
```
golangcibot: U1000: type `MongoDBService` is unused (from `unused`)

main.go
```diff
@@ -163,7 +163,9 @@ func runGRPCServer(ctx context.Context, deps *serviceDependencies) {
 	mysqlSvc := management.NewMySQLService(deps.db, deps.agentsRegistry)
 	nodeSvc := management.NewNodeService(deps.db, deps.agentsRegistry)
 	serviceSvc := management.NewServiceService(deps.db, deps.agentsRegistry)
+	mongodbSvc := management.NewMongoDBService(deps.db, deps.agentsRegistry)
 
+	managementpb.RegisterMongoDBServer(gRPCServer, managementgrpc.NewManagementMongoDBServer(mongodbSvc))
```
AlekSi: Could we use the same order in this block and in the one above?

services/management/mongodb.go
```diff
@@ -0,0 +1,121 @@
...
+
+		invService, err := inventory.ToInventoryService(service)
+		if err != nil {
+			return err
+		}
+
+		res.Service = invService.(*inventorypb.MongoDBService)
+
+		if req.MongodbExporter {
+
```
AlekSi: Maybe you could start using https://github.com/mvdan/gofumpt to avoid those empty lines?

DexterHD: If it's a problem, I think it would be better to add it to our CI pipeline how do you think? Also, we should add this in confluence.

DexterHD: @AlekSi I mean it's an unpopular opinion basically. Only 199 stars. If it so important why the author doesn't add PR in official `gofmt` and why `gofmt` doesn't have the same features? I think it's because it's not necessary. It's just about tastes dispute.

AlekSi: > If it's a problem, I think it would be better to add it to our CI pipeline how do you think?

It is being added to the golangci-lint.

> it's an unpopular opinion basically. Only 199 stars.

Do you know that people _buy_ stars? :) That metric is hardly any good. That repo is also 1 month old.

> If it so important why the author doesn't add PR in official gofmt

What makes you think so? Did you do any research?

> why gofmt doesn't have the same features? I think it's because it's not necessary.

Again, did you do any research? There are open issues like 6996.

> It's just about tastes dispute.

Code formatting, in general, is a matter of taste. Still, gofmt disrupted that space. [gofmt's style is no one's favorite, yet gofmt is everyone's favorite](https://go-proverbs.github.io). The same can be told about many (most?) gofumpt and [go-critic](https://go-critic.github.io/overview) checks.

DexterHD: > What makes you think so? Did you do any research?
Because of this checks not exists in official `gofmt`. No, I didn't do any research.
> Again, did you do any research? There are open issues like 6996.
No sorry, I didn't.

All I want to say is: 
I don't want and I don't like to dispute about code style at all as it makes no sense. They (golang developers) give us `gofmt` to don't dispute about code style but developers always find some ways to add more code style rules and then dispute about code style. Our example shows it clearly. :) It's another rule like "Don't use space before if statement".

I mean for me it's nevermind do you use some spaces or you don't. I don't care, we have `gofmt` and in my opinion, if something not covered by `gofmt` (about codestyle) this doesn't make sense. When it will be implemented in `gofmt`, I will use it, but while is not, there a no sense to talk about it.

Linters it's about error checking. When linters checks some code style it's always about taste and for me, it doesn't make any sense. My opinion is: If some code style things not covers by standard `gofmt`, we shouldn't dispute about it.

PR #182, opened 2019-05-07
Add exhaustiveness checks on our "sum types".

AlekSi:
https://github.com/percona/pmm/pull/344


services/agents/registry.go
```diff
@@ -32,11 +33,17 @@ import (
 	"gopkg.in/reform.v1"
 
 	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/services/agents/channel"
 	"github.com/percona/pmm-managed/utils/logger"
 )
 
+const (
+	prometheusNamespace = "pmm_managed"
```
golangcibot: U1000: const `prometheusNamespace` is unused (from `unused`)

services/agents/registry.go
```diff
@@ -32,11 +33,17 @@ import (
...
+	prometheusSubsystem = "agents"
```
golangcibot: U1000: const `prometheusSubsystem` is unused (from `unused`)

services/agents/registry.go
```diff
@@ -144,49 +151,43 @@ func (r *Registry) Run(stream agentpb.Agent_ConnectServer) error {
...
-					},
+				agent.channel.SendResponse(&channel.Response{
+					ID:      req.ID,
+					Payload: new(agentpb.QANCollectResponse),
 				})
 
-			default:
-				l.Warnf("Unexpected request payload: %s.", msg)
+			case nil:
+				l.Warnf("Unexpected request: %v.", req)
```
BupycHuk: it makes pmm-managed not backward compatible.

AlekSi: You probably mean "forward compatible" – ability to work with newer pmm-agents. pmm-managed is already not forward compatible – if there is an unexpected type, the error would happen earlier, during protobuf decoding. And we don't need to make pmm-managed forward compatible. (We need it in a limited form for pmm-agent.)

PR #195, opened 2019-05-21
PMM-4047 Check connection to service.

BupycHuk:


testdata/prometheus/prometheus.yml
```diff
@@ -34,3 +34,105 @@ scrape_configs:
     - 127.0.0.1:7773
     labels:
       instance: pmm-server
+- job_name: mysqld_exporter_agent_id_4ab6cc14-10b8-4bcc-bba2-aab95e81299c_hr
```
DexterHD: It seems this all stuff should be reverted??

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
...
+	resp := pmmAgent.channel.SendRequest(request)
+	l.Infof("CheckConnection response: %+v.", resp)
+	checkConnectionResponse := resp.(*agentpb.CheckConnectionResponse)
+	if !checkConnectionResponse.Success {
+		return nil
+	}
+	return status.Error(codes.FailedPrecondition, checkConnectionResponse.Error)
+}
+
+func (r *Registry) get(ctx context.Context, pmmAgentID string) (*agentInfo, error) {
```
golangcibot: `(*Registry).get` - `ctx` is unused (from `unparam`)

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
...
+	}
+	return status.Error(codes.FailedPrecondition, checkConnectionResponse.Error)
+}
+
+func (r *Registry) Get(ctx context.Context, pmmAgentID string) (*agentInfo, error) {
+	r.rw.RLock()
+	pmmAgent := r.agents[pmmAgentID]
+	r.rw.RUnlock()
+	if pmmAgent == nil {
+		return nil, status.Error(codes.Unimplemented, fmt.Sprintf("pmm-agent with ID %q is not currently connected.", pmmAgentID))
```
DexterHD: I think, `codes.Unimplemented` should be changed to something else. Maybe to `codes.Internal`?

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
...
+	resp := pmmAgent.channel.SendRequest(request)
+	l.Infof("CheckConnection response: %+v.", resp)
+	checkConnectionResponse := resp.(*agentpb.CheckConnectionResponse)
+	if !checkConnectionResponse.Success {
+		return nil
+	}
+	return status.Error(codes.FailedPrecondition, checkConnectionResponse.Error)
+}
+
+func (r *Registry) get(pmmAgentID string) (*agentInfo, error) {
```
DexterHD: An interesting thing, I have something like this in Actions branch:

https://github.com/percona/pmm-managed/blob/bc909b2f8b472b8827f352662c0c19e3bd7f96da/services/agents/registry.go#L345-L354

Maybe we should merge those things somehow?

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
...
+	l := logger.Get(ctx)
+	start := time.Now()
+	defer func() {
+		if dur := time.Since(start); dur > time.Second {
+			l.Warnf("CheckConnectionToService took %s.", dur)
+		}
+	}()
+
+	pmmAgentID := pointer.GetString(agent.PMMAgentID)
+	pmmAgent, err := r.get(pmmAgentID)
```
AlekSi: Just `*agent.PMMAgentID`. It is ok panic and explode if caller passed nil agent, or agent without pmm-agent ID.

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
...
+	if err != nil {
+		l.Infof("CheckConnectionToService: pmm-agent with ID %q is not currently connected.", pmmAgentID)
+		return err
+	}
+	var request *agentpb.CheckConnectionRequest
```
AlekSi: Please move one line down, closer to switch where it is used.

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
...
+		}
+	default:
+		l.Panicf("unhandled Service type %s", service.ServiceType)
+	}
+
+	l.Infof("CheckConnectionRequest: %+v.", request)
+	resp := pmmAgent.channel.SendRequest(request)
+	l.Infof("CheckConnection response: %+v.", resp)
+	checkConnectionResponse := resp.(*agentpb.CheckConnectionResponse)
+	if !checkConnectionResponse.Success {
```
AlekSi: Inverted logic (remove `!`).

services/inventory/agents.go
```diff
@@ -325,6 +325,17 @@ func (as *AgentsService) AddMySQLdExporter(ctx context.Context, req *inventorypb
...
 			return err
 		}
+		if !req.SkipConnectionCheck {
+			service, err := models.FindServiceByID(tx.Querier, req.ServiceId)
+			if err != nil {
+				return err
+			}
+
+			err = as.r.CheckConnectionToService(ctx, service, row)
+			if err != nil {
```
AlekSi:
```go
if err = as.r.CheckConnectionToService(ctx, service, row); err != nil {
```

I'm not a fan of that syntax myself, but it is idiomatic, especially when only a single value (error) is returned.

services/inventory/agents_test.go
```diff
@@ -73,6 +74,7 @@ func TestAgents(t *testing.T) {
 
 		as.r.(*mockRegistry).On("IsConnected", "/agent_id/00000000-0000-4000-8000-000000000001").Return(true)
 		as.r.(*mockRegistry).On("SendSetStateRequest", ctx, "/agent_id/00000000-0000-4000-8000-000000000001")
+		as.r.(*mockRegistry).On("CheckConnectionToService", ctx, mock.Anything, mock.Anything).Return(nil)
```
AlekSi: Can we avoid `Anything`?

testdata/prometheus/prometheus.yml
AlekSi: Yes, it should be reverted

services/agents/registry.go
BupycHuk: @DexterHD let's merge it once some of us merge his PR. :)

services/agents/registry.go
```diff
@@ -456,6 +459,65 @@ func (r *Registry) SendSetStateRequest(ctx context.Context, pmmAgentID string) {
 	l.Infof("SetState response: %+v.", resp)
 }
 
+// CheckConnectionToService sends request to pmm-agent to check connection to service.
+func (r *Registry) CheckConnectionToService(ctx context.Context, service *models.Service, agent *models.Agent) error {
+	// TODO: extract to a separate struct to keep Single Responsibility principles.
+	l := logger.Get(ctx)
+	start := time.Now()
+	defer func() {
+		if dur := time.Since(start); dur > time.Second {
```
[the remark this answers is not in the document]
AlekSi: k, let's change default to 1 second everywhere. Edited comment above.

services/inventory/agents_test.go
BupycHuk: unfortunately, we can change it with only AnythingTypeOf(). 
Params required here are *models.Service and *models.Agent. Both of them have CreatedAt and UpdatedAt fields with current time value. 
Any idea how to handle it?

services/agents/registry.go
BupycHuk: I think returning error with agent looks better here than just agent. It's not obvious that pmmAgent might be a nil.

services/inventory/agents_test.go
```diff
@@ -73,6 +75,9 @@ func TestAgents(t *testing.T) {
 
 		as.r.(*mockRegistry).On("IsConnected", "/agent_id/00000000-0000-4000-8000-000000000001").Return(true)
 		as.r.(*mockRegistry).On("SendSetStateRequest", ctx, "/agent_id/00000000-0000-4000-8000-000000000001")
+		as.r.(*mockRegistry).On("CheckConnectionToService", ctx,
+			mock.AnythingOfType(reflect.TypeOf(&models.Service{}).Name()),
+			mock.AnythingOfType(reflect.TypeOf(&models.Agent{}).Name())).Return(nil)
```
AlekSi: 👍

PR #198, opened 2019-05-24
Expose more version information.

AlekSi:


AlekSi:
@DexterHD @BupycHuk PTAL

PR #201, opened 2019-05-29
PMM-4127 Restored missed vmstat metrics.

adivinho:


services/agents/node_test.go
```diff
@@ -41,16 +41,19 @@ func TestNodeExporterConfig(t *testing.T) {
...
 					"|Icmp6?_(OutEchoReps|OutEchos|InEchos|InEchoReps|InAddrMaskReps|InAddrMasks|OutAddrMaskReps|OutAddrMasks|InTimestampReps|InTimestamps" +
 					"|OutTimestampReps|OutTimestamps|OutErrors|InDestUnreachs|OutDestUnreachs|InTimeExcds|InRedirects|OutRedirects|InMsgs|OutMsgs)" +
 					"|IcmpMsg_(InType3|OutType3)|Ip(6|Ext)_(InOctets|OutOctets)|Ip_Forwarding|TcpExt_(Listen.*|Syncookies.*|TCPTimeouts))$",
-				"--collector.processes",
-				"--collector.qdisc",
-				"--collector.wifi",
-				"--web.listen-address=:{{ .listen_port }}",
+				"--collector.vmstat.fields=^(pg(steal_(kswapd|direct)|refill|alloc)_(movable|normal|dma3?2?)" +
+					"|nr_(dirty.*|slab.*|vmscan.*|isolated.*|free.*|shmem.*|i?n?active.*|anon_transparent_.*|writeback.*|unstable" +
+					"|unevictable|mlock|mapped|bounce|page_table_pages|kernel_stack)|drop_slab|slabs_scanned|pgd?e?activate)$",
```
AlekSi: Flags in `node.go` are grouped (and each group is sorted), but then they are all sorted for pmm-agent to receive a stable list and not restart agent only if an order of flags changed in pmm-managed because of refactoring. This test checks already sorted list, that's why it fails.

PR #230, opened 2019-07-23
PMM-4351 Split QAN protocol.

BupycHuk:
Should be merged together with https://github.com/percona/pmm-agent/pull/93.


services/agents/deps.go
```diff
@@ -34,5 +34,5 @@ type prometheusService interface {
 // qanClient is a subset of methods of qan.Client used by this package.
 // We use it instead of real type for testing and to avoid dependency cycle.
 type qanClient interface {
-	Collect(ctx context.Context, req *qanpb.CollectRequest) error
+	Collect(ctx context.Context, req *agentpb.CollectRequest) error
```
golangcibot: U1000: func `qanClient.Collect` is unused (from `unused`)

BupycHuk: it's used here https://github.com/percona/pmm-managed/pull/230/files#diff-7a78a0bf5b95767beb8c6dac1f0eb334R182

services/agents/deps.go
```diff
@@ -19,20 +19,17 @@ package agents
...
 type prometheusService interface {
 	UpdateConfiguration()
 }
 
 // qanClient is a subset of methods of qan.Client used by this package.
-// We use it instead of real type for testing and to avoid dependency cycle.
+// We use it instead of real type to avoid dependency cycle.
 type qanClient interface {
-	Collect(ctx context.Context, req *qanpb.CollectRequest) error
+	Collect(ctx context.Context, metricsBuckets []*agentpb.MetricsBucket) error
```
golangcibot: U1000: func `qanClient.Collect` is unused (from `unused`)

services/qan/client.go
```diff
@@ -125,19 +127,54 @@ func (c *Client) Collect(ctx context.Context, req *qanpb.CollectRequest) error {
...
+			Username:             m.Common.Username,
+			ClientHost:           m.Common.ClientHost,
+			ServiceType:          string(service.ServiceType),
+			AgentId:              m.Common.AgentId,
+			AgentType:            m.Common.AgentType,
+			PeriodStartUnixSecs:  m.Common.PeriodStartUnixSecs,
+			PeriodLengthSecs:     m.Common.PeriodLengthSecs,
+			Example:              m.Common.Example,
+			ExampleFormat:        qanpb.ExampleFormat(m.Common.ExampleFormat),
+			IsTruncated:          m.Common.IsTruncated,
```
golangcibot: m.Common.IsTruncated undefined (type *agentpb.MetricsBucket_Common has no field or method IsTruncated) (from `typecheck`)

services/qan/client.go
```diff
@@ -125,19 +127,54 @@ func (c *Client) Collect(ctx context.Context, req *qanpb.CollectRequest) error {
...
+			ExampleType:          qanpb.ExampleType(m.Common.ExampleType),
+			ExampleMetrics:       m.Common.ExampleMetrics,
```
golangcibot: m.Common.ExampleMetrics undefined (type *agentpb.MetricsBucket_Common has no field or method ExampleMetrics) (from `typecheck`)

services/qan/client.go
```diff
@@ -77,32 +78,33 @@ func mergeLabels(node *models.Node, service *models.Service, agent *models.Agent
...
 	// https://jira.percona.com/browse/PMM-3894
 
 	start := time.Now()
 	defer func() {
 		if dur := time.Since(start); dur > time.Second {
-			c.l.Warnf("Collect for %d buckets took %s.", len(req.MetricsBucket), dur)
+			c.l.Warnf("Collect for %d buckets took %s.", len(metricsBuckets), dur)
 		}
 	}()
+	metricBuckets := make([]*qanpb.MetricsBucket, len(metricsBuckets))
```
AlekSi: First, it is a bad name: `metricBuckets` and `metricsBuckets` are too close to each other.

Second, there is a logical mistake: we make `metricBuckets` of the same length as `metricsBuckets`, but then we can skip (`continue`) some buckets below. That will result in nil values at the end of the slice.

I think we should set capacity, not length there, and `append` to the slice.

Please also add a test for that.

services/qan/client.go
```diff
@@ -125,19 +127,52 @@ func (c *Client) Collect(ctx context.Context, req *qanpb.CollectRequest) error {
...
+			Username:             m.Common.Username,
+			ClientHost:           m.Common.ClientHost,
+			ServiceType:          string(service.ServiceType),
+			AgentId:              m.Common.AgentId,
+			AgentType:            m.Common.AgentType,
+			PeriodStartUnixSecs:  m.Common.PeriodStartUnixSecs,
+			PeriodLengthSecs:     m.Common.PeriodLengthSecs,
+			Example:              m.Common.Example,
+			ExampleFormat:        qanpb.ExampleFormat(m.Common.ExampleFormat),
+			ExampleType:          qanpb.ExampleType(m.Common.ExampleType),
```
AlekSi: That code will break if one enum will change a value, but another will not. We need an honest switch for both.

services/qan/client.go
```diff
@@ -125,19 +127,52 @@ func (c *Client) Collect(ctx context.Context, req *qanpb.CollectRequest) error {
...
-			"replication_set": &m.ReplicationSet,
-			"cluster":         &m.Cluster,
-			"service_type":    &m.ServiceType,
-			"environment":     &m.Environment,
-			"az":              &m.Az,
-			"region":          &m.Region,
-			"node_model":      &m.NodeModel,
-			"container_name":  &m.ContainerName,
-			"agent_id":        &m.AgentId,
+			"service_name":    &mb.ServiceName,
```
AlekSi: We should also skip that one. https://jira.percona.com/browse/PMM-4402

services/qan/client_test.go
```diff
@@ -17,9 +17,323 @@
...
+					MRowsSentSum:     4079,
+					MRowsExaminedCnt: 1,
+					MRowsExaminedSum: 4079,
+					MFullScanCnt:     1,
+					MFullScanSum:     1,
+					MNoIndexUsedCnt:  1,
+					MNoIndexUsedSum:  1,
+				},
+			}}
+			c.AssertCalled(t, "Collect", ctx, expectedRequest)
```
AlekSi: Why not
```go
defer c.AssertExpectations(t)
```
at the beginning?

BupycHuk: Because it generates better message.

services/qan/client.go
AlekSi: > Please also add a test for that.

https://github.com/percona/pmm-managed/pull/230/commits/cfd30e24ce0810a1bfc971a965cf793c4105e393

BupycHuk: sorry, missed that

PR #235, opened 2019-07-24
PMM-3643 Add collect query parameter to Prometheus configuration file generator.

DexterHD:
@AlekSi Should I do something like this? There is a lot of work should be done before we can use this but is this the right direction?

CC @rnovikovP


services/prometheus/prometheus.go
```diff
@@ -174,13 +174,13 @@ func (svc *Service) marshalConfig() ([]byte, error) {
...
 				for _, node := range nodes {
-					scfg, err := scrapeConfigForNodeExporter(s.HR, node, agent)
+					scfgs, err := scrapeConfigForNodeExporter(&s, node, agent)
 					if err != nil {
 						svc.l.Warnf("Failed to add %s %q, skipping: %s.", agent.AgentType, agent.AgentID, err)
 						continue
 					}
-					if scfg != nil {
-						cfg.ScrapeConfigs = append(cfg.ScrapeConfigs, scfg)
+					if scfgs != nil {
```
AlekSi: That check is not needed for a slice.

services/prometheus/scrape_configs.go
```diff
@@ -142,7 +143,7 @@ func scrapeConfigForStandardExporter(interval time.Duration, node *models.Node,
 	}
 
 	cfg := &config.ScrapeConfig{
-		JobName:        jobName(agent),
+		JobName:        fmt.Sprintf("%s_%s", jobName(agent), interval),
```
AlekSi: I would let `jobName` function return a job name, not a part of it. Add `interval` as a parameter to it.

services/prometheus/scrape_configs.go
```diff
@@ -174,68 +175,146 @@ func scrapeConfigForStandardExporter(interval time.Duration, node *models.Node,
...
+		"filesystem",
+		"loadavg",
+		"meminfo",
+		"netdev",
+		"netstat",
+		"stat",
+		"time",
+		"vmstat",
+		"meminfo_numa",
+		"textfile(hr)",           // TODO: Implement it!
```
AlekSi: `textfile.hr`, see https://jira.percona.com/browse/PMM-3645. Parentheses would require escaping in both query parameters and in a shell.

That ticket also can be done independently, no need to wait for PMM-3645.

models/settings.go
```diff
@@ -20,13 +20,28 @@ import (
 	"time"
 )
 
+const (
+	MetricResolutionHigh   = time.Second
```
golangcibot: exported const MetricResolutionHigh should have comment (or a comment on this block) or be unexported (from `golint`)

services/prometheus/scrape_configs.go
```diff
@@ -174,68 +175,178 @@ func scrapeConfigForStandardExporter(interval time.Duration, node *models.Node,
 	return cfg, nil
 }
 
-func scrapeConfigForNodeExporter(interval time.Duration, node *models.Node, agent *models.Agent) (*config.ScrapeConfig, error) {
-	return scrapeConfigForStandardExporter(interval, node, nil, agent, []string{})
+func scrapeConfigForNodeExporter(s *models.MetricsResolutions, node *models.Node, agent *models.Agent) ([]*config.ScrapeConfig, error) {
```
BupycHuk: let's rename to scraperConfigsForNodeExporter

BupycHuk: and everywhere else where you changed return value from object to array

services/prometheus/scrape_configs_test.go
```diff
@@ -187,13 +303,19 @@ func TestScrapeConfig(t *testing.T) {
...
 			}
 
-			_, err := scrapeConfigsForMySQLdExporter(s, node, service, agent)
+			_, err := scraperConfigsForMySQLdExporter(s, node, service, agent)
 			require.EqualError(t, err, "failed to decode custom labels: unexpected end of JSON input")
 		})
 	})
 
-	t.Run("scrapeConfigForMongoDBExporter", func(t *testing.T) {
+	t.Run("scraperConfigsForMongoDBExporter", func(t *testing.T) {
```
golangcibot: 311-393 lines are duplicate of `services/prometheus/scrape_configs_test.go:395-477` (from `dupl`)

services/prometheus/scrape_configs_test.go
```diff
@@ -250,13 +381,25 @@ func TestScrapeConfig(t *testing.T) {
...
+				LR: 60 * time.Second,
+			}
+
+			_, err := scraperConfigsForMongoDBExporter(res, node, service, agent)
 			require.EqualError(t, err, "failed to decode custom labels: unexpected end of JSON input")
 		})
 	})
 
-	t.Run("scrapeConfigForPostgresExporter", func(t *testing.T) {
+	t.Run("scraperConfigsForPostgresExporter", func(t *testing.T) {
```
golangcibot: 395-477 lines are duplicate of `services/prometheus/scrape_configs_test.go:479-561` (from `dupl`)

services/prometheus/scrape_configs_test.go
```diff
@@ -316,13 +465,25 @@ func TestScrapeConfig(t *testing.T) {
...
+				LR: 60 * time.Second,
+			}
+
+			_, err := scraperConfigsForPostgresExporter(res, node, service, agent)
 			require.EqualError(t, err, "failed to decode custom labels: unexpected end of JSON input")
 		})
 	})
 
-	t.Run("scrapeConfigForProxySQLExporter", func(t *testing.T) {
+	t.Run("scraperConfigsForProxySQLExporter", func(t *testing.T) {
```
golangcibot: 479-561 lines are duplicate of `services/prometheus/scrape_configs_test.go:311-393` (from `dupl`)

services/prometheus/scrape_configs.go
```diff
@@ -129,20 +130,20 @@ func mergeLabels(node *models.Node, service *models.Service, agent *models.Agent
...
-	return string(agent.AgentType) + strings.Replace(agent.AgentID, "/", "_", -1)
+func jobName(agent *models.Agent, interval time.Duration) string {
+	return fmt.Sprintf("%s%s_%s", agent.AgentType, strings.Replace(agent.AgentID, "/", "_", -1), interval)
 }
 
-// scrapeConfigForStandardExporter returns scrape config for standard exporter: /metrics endpoint, high resolution.
+// scraperConfigForStandardExporter returns scrape config for standard exporter: /metrics endpoint, high resolution.
 // If listen port is not known yet, it returns (nil, nil).
-func scrapeConfigForStandardExporter(interval time.Duration, node *models.Node, service *models.Service, agent *models.Agent, collect []string) (*config.ScrapeConfig, error) {
+func scraperConfigForStandardExporter(interval time.Duration, node *models.Node, service *models.Service, agent *models.Agent, collect []string) (*config.ScrapeConfig, error) {
```
golangcibot: line is 176 characters (from `lll`)

services/prometheus/scrape_configs.go
```diff
@@ -174,68 +175,178 @@ func scrapeConfigForStandardExporter(interval time.Duration, node *models.Node,
...
+		"loadavg",
+		"meminfo",
+		"netdev",
+		"netstat",
+		"stat",
+		"time",
+		"vmstat",
+		"meminfo_numa",
+		"textfile.hr",
+		"textfile.directory.hr",
```
AlekSi: Added `standard.process` (`prometheus.NewProcessCollector`) and `standard.go` (`prometheus.NewGoCollector`) to the Confluence page. When `collect[]` is present, and we filter collectors, we should enable them in exporter only when requested, and we should request them only for HR metrics in PMM. It is a mistake to send them for all resolutions.

services/prometheus/scrape_configs.go
AlekSi: I'm all for renaming `Config` to `Configs` but renaming `scrape` to `scraper` is a mistake (and possibly a typo) – https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config.
But it is not worth fixing now.

DexterHD: I thought that Nurlan suggest `scraper` because the current function name sounds strange in English: `scrape config for node exporter` (But we don't scrape config here. We get config which called `scrapeConfig`). In this case, I would prefer the name like `get scrape config for node exporter`.

AlekSi: https://golang.org/doc/effective_go.html#Getters

DexterHD: Yep, but `scrape` is a verb in the context of Prometheus. Anyway, it does not make a lot of sense to argue about this.

services/prometheus/scrape_configs_test.go
```diff
@@ -51,34 +57,108 @@ func TestScrapeConfig(t *testing.T) {
...
-							"agent_type":        "node_exporter",
-							"instance":          "/agent_id/75bb30d3-ef4a-4147-97a8-621a996611dd",
-							"node_id":           "/node_id/cc663f36-18ca-40a1-aea9-c6310bb4738d",
-							"node_name":         "node_name",
-						},
+			expected := []*config.ScrapeConfig{
+				{
+					JobName:        "node_exporter_agent_id_75bb30d3-ef4a-4147-97a8-621a996611dd_1s",
+					ScrapeInterval: model.Duration(s.HR),
+					ScrapeTimeout:  scrapeTimeout(s.HR),
```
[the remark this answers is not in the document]
DexterHD: My point here is we are testing `scrapeConfigForNodeExporter()`. It this case, we shouldn't assume that inside this function `scrapeTimeout(s.HR)` being used. If we want to test exactly `scrapeTimeout()` function, we should make a unit test for it.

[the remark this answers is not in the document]
AlekSi: Cool, then use `jobName` function too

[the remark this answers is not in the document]
DexterHD: I've just explained my decision. Moreover, I have used `scrapeTimeout()` also to not calculate those values by hand. Yep, cause we have a separate function `jobName` I don't see anything bad to use it too.

PR #245, opened 2019-08-13
PMM-4249 Add API QAN data retention.

askomorokhov:


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/pmm-managed?pullRequest=245) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/pmm-managed?pullRequest=245) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/pmm-managed?pullRequest=245) it.</sub>

models/settings.go
```diff
@@ -27,6 +27,11 @@ type MetricsResolutions struct {
 	LR time.Duration `json:"lr"`
 }
 
+// QAN contains query analytics configuration.
+type QAN struct {
+	DataRetention uint32 `json:"data_retention"` // number of days data retention
```
AlekSi: Please use `time.Duration`

services/server/server.go
```diff
@@ -98,6 +99,13 @@ func (s *Server) parseEnv(env []string) {
 			if err == nil {
 				s.envDisableTelemetry = b
 			}
+
+		case "QANAPI_DATA_RETENTION":
```
AlekSi: It should be `DATA_RETENTION`. Use `time.ParseDuration`. Unlike `METRICS_RESOLUTION`, there suffix is required.

models/settings.go
askomorokhov: @AlekSi, just to confirm. The minimum truncation of partitions is a day. So is it OK, from UX point of view, to specify data retention as a `time.Duration` where maximum is an hour. In this case user or UI need to multiply days by 24 and on api side we need to do this in a reverse order.

AlekSi: API will use standard type `google.protobuf.Duration` so we don't use several different types in the same API for durations. But there should be a validation that incoming value represents X whole days.
Datastore will use standard type `time.Duration` for the same reason. There also should be a validation.
In the future, that will allow us to reduce granularity (say, to hours instead of days) without breaking backward compatibility of API and stored data.

> time.Duration where maximum is an hour
`time.Hour` is a maximum predefined constant, but nothing prevents us from using `720*time.Hour` to represent 30 days.

services/server/server_test.go
```diff
@@ -20,46 +20,71 @@ import (
 	"testing"
 	"time"
 
+	"github.com/percona/pmm-managed/models"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

models/settings.go
```diff
@@ -27,18 +27,24 @@ type MetricsResolutions struct {
...
 
+// QAN contains query analytics configuration.
+type QAN struct {
+	DataRetention time.Duration `json:"data_retention"`
+}
+
 // Settings contains PMM Server settings.
 type Settings struct {
+	Updates struct {
+		AuthToken string `json:"auth_token"`
```
golangcibot: U1000: field `AuthToken` is unused (from `unused`)

AlekSi: Indeed

models/settings_helpers.go
```diff
@@ -44,6 +47,29 @@ func GetSettings(q reform.DBTX) (*Settings, error) {
...
+		if pair.dur.Truncate(time.Second) != pair.dur {
+			return status.Error(codes.InvalidArgument, pair.name+": should be a natural number of seconds")
+		}
+	}
+
+	if s.QAN.DataRetention < 24*time.Hour {
+		return status.Error(codes.InvalidArgument, "data_retention: minimal resolution is 24h")
+	}
+	if s.QAN.DataRetention.Truncate(24*time.Hour) != s.QAN.DataRetention {
+		return status.Error(codes.InvalidArgument, "data_retention: should be a natural number of days")
```
BupycHuk: why don't we just use days?

AlekSi: https://github.com/percona/pmm-managed/pull/245#discussion_r313765106

peterschwaller:
Dummy comment to trigger sync to Jira.  Please ignore.

PR #254, opened 2019-08-23
PMM-4534 Added initial TLS support for PostgreSQL

percona-csalguero:
It is initial support and not full support because in this step we added just 2 fields to signal ssl enabled= on/off and ssl mode = require/verify-full but in order to fully support verify-full wee need to handle certificates.
Even that, verify-full should work since the PostgreSQL can handle certificates in the DSN used when started the exporter.


models/database.go
```diff
@@ -111,6 +111,8 @@ var databaseSchema = [][]string{
 			username VARCHAR CHECK (username <> ''),
 			password VARCHAR CHECK (password <> ''),
 			metrics_url VARCHAR CHECK (metrics_url <> ''),
+			tls BOOLEAN,
+			tls_skip_verify BOOLEAN NULL,
```
AlekSi: It should be a pointer in model, or non-NULLable column. In reform, `nil == NULL`.

services/inventory/agents.go
```diff
@@ -413,11 +413,13 @@ func (as *AgentsService) AddPostgresExporter(ctx context.Context, req *inventory
...
-			Username:     req.Username,
-			Password:     req.Password,
-			CustomLabels: req.CustomLabels,
+			PMMAgentID:    req.PmmAgentId,
+			ServiceID:     req.ServiceId,
+			Username:      req.Username,
+			Password:      req.Password,
+			CustomLabels:  req.CustomLabels,
+			TLS:           req.Tls,
+			TLSSkipVerify: req.TlsSkipVerify,
```
AlekSi: The same should be done in Management API. Please search for `CreateAgentParams`.

models/database.go
```diff
@@ -111,6 +111,8 @@ var databaseSchema = [][]string{
 			username VARCHAR CHECK (username <> ''),
 			password VARCHAR CHECK (password <> ''),
 			metrics_url VARCHAR CHECK (metrics_url <> ''),
+			tls BOOLEAN,
```
BupycHuk: let's create new migration instead of changing existed one.

AlekSi: @percona-csalguero To do that, add a new slice element below. Also, make sure it is compatible with existing data. Something like this should work:
* add columns with DEFAULT false;
* remove default values.

models/database.go
```diff
@@ -119,6 +119,11 @@ var databaseSchema = [][]string{
 			CONSTRAINT runs_on_node_id_only_for_pmm_agent CHECK ((runs_on_node_id IS NULL) <> (agent_type='` + string(PMMAgentType) + `'))
 		)`,
 
+		`ALTER TABLE agents
```
BupycHuk: it's still a part of another migration. 
please take a look at this line, it creates new migration
https://github.com/percona/pmm-managed/pull/254/files#diff-087672b3a344779900ce93f4af23902dR161

AlekSi:
@percona-csalguero Please make changes as requested and re-request reviews: 
image

models/database.go
```diff
@@ -159,6 +159,13 @@ var databaseSchema = [][]string{
 		)`,
 		`INSERT INTO settings (settings) VALUES ('{}')`,
 	},
+
+	3: {
+		`ALTER TABLE agents 
+           ADD COLUMN tls BOOLEAN NOT NULL DEFAULT FALSE,
+           ADD COLUMN tls_skip_verify BOOLEAN NOT NULL DEFAULT TRUE
+		`,
```
AlekSi: The idea was to add another `ALTER TABLE` that would remove default values. That way we will migrate existing data, but also will not have `DEFAULT`s in database (as they can be problematic in corner-cases for reform).

percona-csalguero:
Made a new migration and updated tests.

AlekSi:
Will merge tomorrow morning.

services/converters.go
```diff
@@ -215,19 +216,22 @@ func ToAPIAgent(q *reform.Querier, agent *models.Agent) (inventorypb.Agent, erro
...
+			PmmAgentId:    pointer.GetString(agent.PMMAgentID),
+			ServiceId:     serviceID,
+			Username:      pointer.GetString(agent.Username),
+			Password:      pointer.GetString(agent.Password),
+			Disabled:      agent.Disabled,
+			Status:        inventorypb.AgentStatus(inventorypb.AgentStatus_value[agent.Status]),
+			ListenPort:    uint32(pointer.GetUint16(agent.ListenPort)),
+			CustomLabels:  labels,
+			Tls:           agent.TLS,
+			TlsSkipVerify: agent.TLSSkipVerify,
```
AlekSi: @percona-csalguero that one important

PR #267, opened 2019-09-11
PMM-4362 Hide pt-* actions for now.

AlekSi:


services/agents/registry.go
```diff
@@ -524,6 +524,7 @@ func (r *Registry) Collect(ch chan<- prom.Metric) {
 	r.mClockDrift.Collect(ch)
 }
 
+/*
```
BupycHuk: Let's remove instead of comment here too.

PR #275, opened 2019-09-14
PMM-3786 Cleanup APIs.

AlekSi:


services/agents/registry.go
```diff
@@ -458,30 +458,32 @@ func (r *Registry) CheckConnectionToService(ctx context.Context, service *models
 	switch service.ServiceType {
 	case models.MySQLServiceType:
 		request = &agentpb.CheckConnectionRequest{
-			Type: inventorypb.ServiceType_MYSQL_SERVICE,
-			Dsn:  agent.DSN(service, 2*time.Second, ""),
+			Type:    inventorypb.ServiceType_MYSQL_SERVICE,
+			Dsn:     agent.DSN(service, 2*time.Second, ""),
+			Timeout: ptypes.DurationProto(3 * time.Second),
```
BupycHuk: maybe we should extract this timeout and timeout above to consts, what do you think?

AlekSi: I don't think they will be constants for long, to be honest. For example, it is likely we will use higher values for remote services.

PR #294, opened 2019-11-06
PMM-4317 Upload SSH key.

BupycHuk:


AlekSi:
Build is broken.

services/server/server.go
```diff
@@ -493,6 +496,70 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
 	return res, nil
 }
 
+// SetSSHKey uploads public key.
+func (s *Server) SetSSHKey(ctx context.Context, in *serverpb.SetSSHKeyRequest) (*serverpb.SetSSHKeyResponse, error) {
+	//TODO: validate ssh key
```
AlekSi: Let's do it now. `ssh-keygen -l -f /tmp/XXX`, if ok – move to .ssh directory.

services/server/server.go
```diff
@@ -493,6 +496,70 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+	var settings *models.Settings
+	err := s.db.InTransaction(func(tx *reform.TX) error {
+		var e error
+		if settings, e = models.GetSettings(tx); e != nil {
+			return e
+		}
+
+		username := "admin"
+		usr, err := user.Lookup(username)
```
AlekSi: Note that there you start using `err` instead of `e`.
Most of our code looks like that:
```
	e := r.db.InTransaction(func(tx *reform.TX) error {
		// use err there
	}
```

services/server/server.go
```diff
@@ -493,6 +496,70 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+			return err
+		}
+		gid, err := strconv.Atoi(usr.Gid)
+		if err != nil {
+			return err
+		}
+
+		sshDirPath := path.Join(usr.HomeDir, ".ssh")
+		if _, err := os.Stat(sshDirPath); os.IsNotExist(err) {
+			err := os.Mkdir(sshDirPath, os.FileMode(0700))
```
AlekSi: There we introduce a _third_ `err`. :)

The whole think can be simplified by using `os.MkdirAll` – it works even if path exist.

services/server/server.go
```diff
@@ -493,6 +496,70 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+		}
+
+		keysPath := path.Join(sshDirPath, "authorized_keys")
+
+		file, err := os.OpenFile(keysPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(0600))
+		if err != nil {
+			return err
+		}
+
+		_, err = file.WriteString(in.SshKey)
```
AlekSi: Note that file is never closed there. That's a big problem.

Use `ioutil.WriteFile` to simplify the whole thing.

models/settings.go
```diff
@@ -41,6 +41,8 @@ type Settings struct {
 	MetricsResolutions MetricsResolutions `json:"metrics_resolutions"`
 
 	DataRetention time.Duration `json:"data_retention"`
+
+	SSHKey string `json:"ssh_key"`
```
AlekSi: If we want to support users overwriting that file without API, then there is no point to storing it in the database.

services/server/server.go
BupycHuk: yeah, I'll rename.
In most cases we
```
e := r.db.InTransaction(func(tx *reform.TX) error {
		// use err there
	}
```
but in this file in other method

```
err := r.db.InTransaction(func(tx *reform.TX) error {
		// use e there
	}
```
Which one I should use or it doesn't matter?

services/server/server.go
BupycHuk: MkdirAll creates home directory too, is it expected result? In this case we should change user owner for home directory too.

services/server/server.go
```diff
@@ -493,6 +514,67 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
 }
 
+func (s *Server) validateSSHKey(sshKey string) error {
+	tempFile := path.Join(os.TempDir(), "temp_keys")
+	err := ioutil.WriteFile(tempFile, []byte(sshKey), os.FileMode(0600))
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.Command("ssh-keygen", "-l", "-f", tempFile)
```
golangcibot: G204: Subprocess launched with variable (from `gosec`)

services/server/server.go
```diff
@@ -493,6 +514,67 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
 	return res, nil
 }
 
+func (s *Server) validateSSHKey(sshKey string) error {
+	tempFile := path.Join(os.TempDir(), "temp_keys")
```
AlekSi: Concurrent calls to this method will overwrite the same file. We should either use a mutex, or ioutil.TempFile.

services/server/server.go
AlekSi: We have to have a timeout there. The easiest ways seems to be to pass the request context with added timeout.

services/server/server.go
```diff
@@ -493,6 +514,67 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+	err := ioutil.WriteFile(tempFile, []byte(sshKey), os.FileMode(0600))
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.Command("ssh-keygen", "-l", "-f", tempFile)
+	pdeathsig.Set(cmd, unix.SIGKILL)
+
+	err = cmd.Start()
```
AlekSi: I guess cmd.Run() would be simpler.

services/server/server.go
AlekSi: In what case we do not have home directory?

services/server/server.go
AlekSi: The most common one wins :)

services/server/server.go
```diff
@@ -493,6 +514,67 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+	uid, err := strconv.Atoi(usr.Uid)
+	if err != nil {
+		return err
+	}
+	gid, err := strconv.Atoi(usr.Gid)
+	if err != nil {
+		return err
+	}
+	sshDirPath := path.Join(usr.HomeDir, ".ssh")
+	err = os.Mkdir(sshDirPath, os.FileMode(0700))
```
AlekSi: https://github.com/percona/pmm-managed/pull/294#discussion_r343569478

services/server/server.go
```diff
@@ -493,6 +514,67 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+	if err != nil && !os.IsExist(err) {
+		return err
+	}
+	err = os.Chown(sshDirPath, uid, gid)
+	if err != nil {
+		return err
+	}
+	keysPath := path.Join(sshDirPath, "authorized_keys")
+	err = ioutil.WriteFile(keysPath, []byte(sshKey), os.FileMode(0600))
```
AlekSi: We should protect from concurrent calls with mutex.

services/server/server.go
```diff
@@ -493,6 +514,67 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+	if err != nil {
+		return err
+	}
+	err = os.Chown(keysPath, uid, gid)
+	if err != nil {
```
AlekSi:
```go
if err = os.Chown(keysPath, uid, gid); err != nil
```
For relatively short simple lines like there and in other places.

services/server/server.go
BupycHuk: locally I haven't had a home directory for the some reason.

AlekSi: Do we have it in current OVF and AMI?

BupycHuk: No

services/server/server.go
BupycHuk: We've already protected it with mutex here https://github.com/percona/pmm-managed/pull/294/files#diff-6f891b9f28cfb10751fe182afa544a9cR428. Do we need extra one?

AlekSi: `envRW` protects `envXXX` fields. That line acquires shared read lock. We need an exclusive lock for `authorized_keys` file.

services/server/server.go
```diff
@@ -493,6 +514,66 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+		return err
+	}
+	if err = ioutil.WriteFile(tempFile.Name(), []byte(sshKey), os.FileMode(0600)); err != nil {
+		return err
+	}
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	cmd := exec.CommandContext(timeoutCtx, "ssh-keygen", "-l", "-f", tempFile.Name())
```
golangcibot: G204: Subprocess launched with variable (from `gosec`)

services/server/server.go
AlekSi: It still wasn't being closed.

services/server/server.go
```diff
@@ -517,33 +517,36 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
 func (s *Server) validateSSHKey(ctx context.Context, sshKey string) error {
 	tempFile, err := ioutil.TempFile("", "temp_keys_*")
 	if err != nil {
-		return err
+		return errors.WithStack(err)
 	}
+	tempFile.Close()                 //nolint:errcheck
+	defer os.Remove(tempFile.Name()) //nolint:errcheck
```
AlekSi: ioutil.TempFile returns open file. We should close it; otherwise, we will leak file descriptor.
We also did not removed that temporary file.

PR #301, opened 2019-11-21
PMM-4966 Filter sensitive information.

BupycHuk:


services/agents/mysql.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // mysqldExporterConfig returns desired configuration of mysqld_exporter process.
-func mysqldExporterConfig(service *models.Service, exporter *models.Agent) *agentpb.SetStateRequest_AgentProcess {
+func mysqldExporterConfig(service *models.Service, exporter *models.Agent, debug bool) *agentpb.SetStateRequest_AgentProcess {
```
golangcibot: Function 'mysqldExporterConfig' is too long (82 > 60) (from `funlen`)

main.go
```diff
@@ -511,7 +511,8 @@ func main() {
 
 	qanClient := getQANClient(ctx, db, *qanAPIAddrF)
 
-	agentsRegistry := agents.NewRegistry(db, prometheus, qanClient)
+	isDebugMode := *debugF || *traceF
```
AlekSi: In other places we use this:
```go
l.Logger.GetLevel() >= logrus.DebugLevel
```

This makes that check request-scoped if there is a request with a scope.

models/agent_model.go
```diff
@@ -298,6 +298,14 @@ func (s *Agent) IsMySQLTablestatsGroupEnabled() bool {
 	}
 }
 
+func (s *Agent) HideKeywords(debug bool) []string {
```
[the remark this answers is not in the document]
AlekSi: At very least, `bool` parameters are bad. If you don't want to use logrus logger level (I can see both ways), let's add "enum" constants.

main.go
BupycHuk: I haven't seen that. I did it this way because I found this https://github.com/percona/pmm-managed/pull/301/files#diff-7ddfb3e035b42cd70649cc33393fe32cR561. Using your way might be better.

services/agents/agents.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/percona/pmm-managed/models"
+)
+
+type debugValue bool
+
+var enableDebug = debugValue(true)
+var disableDebug = debugValue(false)
+
+func redactKeywords(s *models.Agent, debug debugValue) []string {
```
percona-csalguero: Why this function returns an array if as maximum, it will have only one element?

services/agents/mongodb.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // mongodbExporterConfig returns desired configuration of mongodb_exporter process.
-func mongodbExporterConfig(service *models.Service, exporter *models.Agent) *agentpb.SetStateRequest_AgentProcess {
+func mongodbExporterConfig(service *models.Service, exporter *models.Agent, debug debugValue) *agentpb.SetStateRequest_AgentProcess {
```
percona-csalguero: This is just a matter of preference. Instead of passing a boolean parameter in each function call I would prefer to have a way to set a value for `enableDebug`

services/agents/agents.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"github.com/AlekSi/pointer"
+
+	"github.com/percona/pmm-managed/models"
+)
+
+type debugValue bool
+
+var enableDebug = debugValue(true)
+var disableDebug = debugValue(false)
```
percona-csalguero: Do we need both? They are mutually exclusive.

services/agents/agents.go
BupycHuk: because in future it can be changed

services/agents/mysql.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // mysqldExporterConfig returns desired configuration of mysqld_exporter process.
-func mysqldExporterConfig(service *models.Service, exporter *models.Agent) *agentpb.SetStateRequest_AgentProcess {
+func mysqldExporterConfig(service *models.Service, exporter *models.Agent, redactMode redactMode) *agentpb.SetStateRequest_AgentProcess {
```
golangcibot: Function 'mysqldExporterConfig' is too long (85 > 60) (from `funlen`)

services/agents/agents.go
AlekSi: Added `AWSSecretKey` to the list.

PR #315, opened 2019-12-11
PMM-5008 Validation of node type in management API.

BupycHuk:


services/management/management.go
```diff
@@ -34,14 +34,29 @@ func nodeID(tx *reform.TX, nodeID, nodeName string, addNodeParams *managementpb.
 	}
 	switch {
 	case nodeID != "":
-		return nodeID, nil
+		node, err := models.FindNodeByID(tx.Querier, nodeID)
+		if err != nil {
+			return "", err
+		}
+		err = validateExistNodeType(node)
+		if err != nil {
```
AlekSi:
```go
if err = validateExistNodeType(node); err != nil
```

services/management/management.go
```diff
@@ -52,6 +67,15 @@ func nodeID(tx *reform.TX, nodeID, nodeName string, addNodeParams *managementpb.
 	}
 }
 
+func validateExistNodeType(node *models.Node) error {
```
AlekSi: "existing"

services/management/management.go
```diff
@@ -34,14 +34,27 @@ func nodeID(tx *reform.TX, nodeID, nodeName string, addNodeParams *managementpb.
...
 		if err != nil {
 			return "", err
 		}
-		return node.NodeID, nil
+		if err = validateExistingNodeType(node); err != nil {
+			return "", err
+		}
+		return node.NodeID, err
 	case addNodeParams != nil:
+		if addNodeParams.NodeType != inventorypb.NodeType_REMOTE_NODE {
```
percona-csalguero: According to the ticket, remote RDS nodes should also be accepted

AlekSi: @BupycHuk ^

BupycHuk: got it, will fix it right after DS.

AlekSi: @percona-csalguero please request changes for this PR. Otherwise, it lost for everybody:
* it is assigned to you, but you neither merged it, nor declined;
* ticket is also still assigned to you.

BupycHuk: why do we need RDS node type if we have separate API for RDS?

AlekSi: Good point. Do we have a test for that? :)

BupycHuk: tests for API for RDS or adding node with RDS Node Type?

AlekSi: image

BupycHuk: will check and add tests if there are no tests for API for RDS.

PR #317, opened 2019-12-18
PMM-1772 Use stable RDS instances in tests.

BupycHuk:


services/management/rds_test.go
```diff
@@ -121,13 +126,72 @@ func TestRDSService(t *testing.T) {
...
+					NodeModel:     "db.t2.micro",
+					Address:       "autotest-mysql-57.c3uoaol27cbb.us-west-2.rds.amazonaws.com",
+					Port:          3306,
+					Engine:        managementpb.DiscoverRDSEngine_DISCOVER_RDS_MYSQL,
+					EngineVersion: "5.7.22",
+				},
+			}, instances.RdsInstances)
 		})
+
+		type instance struct {
```
golangcibot: type `instance` is unused (from `unused`)

services/management/rds_test.go
```diff
@@ -121,13 +126,72 @@ func TestRDSService(t *testing.T) {
...
+				ctx, cancel := context.WithTimeout(ctx, awsDiscoverTimeout)
+				defer cancel()
+
+				instances, err := discoverRDSRegion(ctx, sess, tt.region)
+
+				assert.NoError(t, err)
+				require.Equal(t, len(instances), len(tt.instances), "Should have one instance")
+				for i, instance := range tt.instances {
+					assert.Equal(t, instance.az, pointer.GetString(instances[i].AvailabilityZone))
+					assert.Equal(t, instance.instanceID, pointer.GetString(instances[i].DBInstanceIdentifier))
```
AlekSi: I assume we compare instances this way to produce better diffs? If yes, I think a comment there will help

BupycHuk: we compare instances this way because there too much fields that we don't need to compare.

PR #323, opened 2019-12-30
PMM-5231 Wrong keys in rds_exporter config.

percona-csalguero:
The updateRDSExporterConfig function was receiving a service as a
parameter and it was using AWS credentials from that service, but we can
have more than one service so, we need to get credentials for each
service individually.

### DIFF
--- a/services/rds/rds.go
+++ b/services/rds/rds.go
@@ -503,8 +503,8 @@ func (svc *Service) updateRDSExporterConfig(tx *reform.TX) (*rdsExporterConfig, 
 	for _, n := range nodes {
 		node := n.(*models.RDSNode)
 
-		service := &models.RDSService{}
-		if e := tx.SelectOneTo(service, "WHERE node_id = ?", node.ID); e != nil {
+		var service models.RDSSeree
+		if e := tx.FindOneTo(&service, "node_id", node.ID); e != nil {
 			return nil, errors.WithStack(e)
 		}
 
--- a/services/rds/rds.go
+++ b/services/rds/rds.go
@@ -493,7 +493,7 @@ func (svc *Service) mysqlExporterCfg(agent *models.MySQLdExporter, dsn string) *
 	}
 }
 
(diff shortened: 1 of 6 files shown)


puneet0191:
@percona-csalguero I have tested this and shared the test log and finding with you, please ping me on slack for details. The configuration file keys are correctly mapped but we have issues in the rds_exporter log. 
I have tested feature build: https://github.com/Percona-Lab/pmm-submodules/pull/608

services/rds/rds.go
```diff
@@ -502,6 +502,12 @@ func (svc *Service) updateRDSExporterConfig(tx *reform.TX, service *models.RDSSe
 	}
 	for _, n := range nodes {
 		node := n.(*models.RDSNode)
+
+		service := &models.RDSService{}
```
AlekSi:
```suggestion
		var service models.RDSService
```

That's a preferred syntax for zero values in PMM code base. See also https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_use_a_consistent_declaration_style

services/rds/rds.go
```diff
@@ -502,6 +502,12 @@ func (svc *Service) updateRDSExporterConfig(tx *reform.TX, service *models.RDSSe
...
+		if e := tx.SelectOneTo(service, "WHERE node_id = ?", node.ID); e != nil {
```
AlekSi:
```go
tx.FindOneTo(&service, "node_id", node.ID)
```
We can and should avoid SQL strings with MySQL-specific placeholders in simplest cases like that one.

services/rds/rds.go
```diff
@@ -502,6 +502,12 @@ func (svc *Service) updateRDSExporterConfig(tx *reform.TX, service *models.RDSSe
 	}
 	for _, n := range nodes {
 		node := n.(*models.RDSNode)
+
+		var service models.RDSSeree
```
BupycHuk: `RDSSeree`?

AlekSi: @percona-csalguero please build it locally

AlekSi:
Now only qan-agent test fails, and it is unlikely we ever fix it.

PR #336, opened 2020-01-29
PMM-4493 Hide updates if they are disabled.

AlekSi:


services/server/server.go
```diff
@@ -256,6 +256,10 @@ func (s *Server) Readiness(ctx context.Context, req *serverpb.ReadinessRequest)
 
 // CheckUpdates checks PMM Server updates availability.
 func (s *Server) CheckUpdates(ctx context.Context, req *serverpb.CheckUpdatesRequest) (*serverpb.CheckUpdatesResponse, error) {
+	s.envRW.RLock()
+	updatesDisabled := s.envDisableUpdates
```
BupycHuk: Why do we check updates and return info about latest version if we want to hide them?

AlekSi: There are some tickets about checking for updates and letting users know about them, but not show update button because DISABLE_UPDATES is in effect. That would require API extension, but less changes on pmm-managed side – due to this change.

services/server/server.go
```diff
@@ -279,15 +283,23 @@ func (s *Server) CheckUpdates(ctx context.Context, req *serverpb.CheckUpdatesReq
 		UpdateAvailable: v.UpdateAvailable,
 		LatestNewsUrl:   v.LatestNewsURL,
 	}
+
+	if updatesDisabled {
+		res.UpdateAvailable = false
```
AlekSi: @BupycHuk basically, we will only replace `UpdateAvailable` with something like `CanUpdate` on that line.

BupycHuk: Yeah, I see that. So it's ok that we return latest version in response, right?

AlekSi: Yes, client depends (should depend) only on `UpdateAvailable` value and should not compare version strings.
Please check that, btw.

BupycHuk: Checked frontend side and yes it uses `UpdateAvailable`, but alt+click will run update even if `updateAvailable` is false.

AlekSi: That's okay, we protect from that in StartUpdate.

BupycHuk: LGTM then

PR #343, opened 2020-02-20
PMM-3387 Implement custom base Prometheus file

percona-csalguero:


services/prometheus/prometheus_test.go
```diff
@@ -518,3 +521,39 @@ scrape_configs:
...
+		db, svc, original := setup(t)
+		defer teardown(t, db, svc, original)
+
+		assert.NoError(t, svc.updateConfiguration())
+		tmpfile, err := ioutil.TempFile("", "mock_base_config.yml")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		defer os.Remove(tmpfile.Name())
```
golangcibot: Error return value of `os.Remove` is not checked (from `errcheck`)

services/prometheus/prometheus.go
```diff
@@ -264,32 +268,56 @@ func (svc *Service) addScrapeConfigs(cfg *config.Config, q *reform.Querier, s *m
...
+	cfg := &config.Config{
+		GlobalConfig: config.GlobalConfig{
+			// these values will be overwriten if they exit in the base config gfile
+			ScrapeInterval:     model.Duration(s.LR),
+			ScrapeTimeout:      scrapeTimeout(s.LR),
+			EvaluationInterval: model.Duration(s.LR),
+		},
+	}
+
+	buf, err := ioutil.ReadFile(filename)
```
golangcibot: G304: Potential file inclusion via variable (from `gosec`)

services/prometheus/prometheus_test.go
```diff
@@ -518,3 +521,39 @@ scrape_configs:
 		assert.Equal(t, expected, string(actual), "actual:\n%s", actual)
 	})
 }
+
+func TestBasePrometheusConfig(t *testing.T) {
```
golangcibot: func `TestBasePrometheusConfig` is unused (from `unused`)

services/prometheus/prometheus_test.go
```diff
@@ -518,3 +521,39 @@ scrape_configs:
...
+  evaluation_interval: 9m
+rule_files:
+- /srv/prometheus/rules/fake_first.rules.yml`)
+
+		assert.NoError(t, ioutil.WriteFile(basePrometheusConfig, cfg, os.ModePerm))
+
+		newcfg, err := svc.marshalConfig()
+		assert.NoError(t, err)
+
+		assert.True(t, bytes.Contains(newcfg, cfg))
```
BupycHuk: @AlekSi is it ok that base config file overrides our global configs?

services/prometheus/prometheus_test.go
```diff
@@ -518,3 +521,39 @@ scrape_configs:
...
+		// Check that the default rule files dir was also included in the config
+		assert.True(t, bytes.Contains(newcfg, []byte(defaultPrometheusRulesDir)))
+
+		basePrometheusConfig = oldBasePrometheusConfig
```
BupycHuk: I think we also should have `assert.Equal` to check the result.

services/prometheus/prometheus.go
```diff
@@ -47,9 +47,13 @@ import (
...
 )
 
-const updateBatchDelay = 3 * time.Second
+const (
+	updateBatchDelay          = 3 * time.Second
+	defaultPrometheusRulesDir = "/srv/prometheus/rules/*.rules.yml"
+)
 
 var checkFailedRE = regexp.MustCompile(`FAILED: parsing YAML file \S+: (.+)\n`)
+var basePrometheusConfig = "/srv/prometheus/prometheus.base.yml"
```
AlekSi: Why use a global variable instead of `Service` field?

services/prometheus/prometheus.go
```diff
@@ -264,32 +268,56 @@ func (svc *Service) addScrapeConfigs(cfg *config.Config, q *reform.Querier, s *m
...
+			// these values will be overwriten if they exit in the base config gfile
+			ScrapeInterval:     model.Duration(s.LR),
+			ScrapeTimeout:      scrapeTimeout(s.LR),
+			EvaluationInterval: model.Duration(s.LR),
+		},
+	}
+
+	buf, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return cfg
```
AlekSi: We should log errors other than `os.IsNotExist` (i.g. file exists, but can't be read due to permissions, etc.)

services/prometheus/prometheus.go
```diff
@@ -264,32 +268,56 @@ func (svc *Service) addScrapeConfigs(cfg *config.Config, q *reform.Querier, s *m
...
+			return true
+		}
+	}
+	return false
+}
+
 // marshalConfig marshals Prometheus configuration.
 func (svc *Service) marshalConfig() ([]byte, error) {
-	var cfg *config.Config
+	cfg := &config.Config{}
```
AlekSi: Why? We don't use that initialised value.

services/prometheus/prometheus.go
```diff
@@ -264,32 +268,56 @@ func (svc *Service) addScrapeConfigs(cfg *config.Config, q *reform.Querier, s *m
...
-				"/srv/prometheus/rules/*.rules.yml",
-			},
-			ScrapeConfigs: []*config.ScrapeConfig{
-				scrapeConfigForPrometheus(s.HR),
-				scrapeConfigForGrafana(s.MR),
-				scrapeConfigForPMMManaged(s.MR),
-				scrapeConfigForQANAPI2(s.MR),
-			},
+		cfg = svc.baseConfig(basePrometheusConfig, s)
+		if !inSlice(cfg.RuleFiles, defaultPrometheusRulesDir) {
```
AlekSi: Why do that extra check? If this path is already in the base file, we should add it twice.

services/prometheus/prometheus.go
```diff
@@ -264,32 +268,56 @@ func (svc *Service) addScrapeConfigs(cfg *config.Config, q *reform.Querier, s *m
...
+			cfg.RuleFiles = append(cfg.RuleFiles, defaultPrometheusRulesDir)
 		}
+		cfg.ScrapeConfigs = append(cfg.ScrapeConfigs, scrapeConfigForPrometheus(s.HR))
+		cfg.ScrapeConfigs = append(cfg.ScrapeConfigs, scrapeConfigForGrafana(s.MR))
+		cfg.ScrapeConfigs = append(cfg.ScrapeConfigs, scrapeConfigForPMMManaged(s.MR))
+		cfg.ScrapeConfigs = append(cfg.ScrapeConfigs, scrapeConfigForQANAPI2(s.MR))
 
 		if settings.AlertManagerURL != "" {
```
AlekSi: We should also add to Alertmanager configurations, not set it.

services/prometheus/prometheus_test.go
AlekSi: Yes, because our GlobalConfig settings are always present; base file is the only way to change them.

PR #344, opened 2020-02-20
PMM-5317 Skip unavailable metrics scrape on macOS

artemgavrilov:


services/prometheus/scrape_configs.go
```diff
@@ -241,43 +241,52 @@ func scrapeConfigForRDSExporter(intervalName string, interval time.Duration, hos
...
+			"standard.go",
+			"standard.process",
+			"stat",
+			"textfile.hr",
+			"vmstat",
+		)
+	}
+
+	hr, err = scrapeConfigForStandardExporter("hr", s.HR, params, append(hrCollect,
 		"cpu",
```
BupycHuk: Do only 7 collectors work on MacOS?

artemgavrilov: Yep, only 7 from specified in that file. `textfile.*` can work on macOS by itself, but it's disabled.
https://github.com/percona/pmm-managed/blob/PMM-2.0/services/agents/node.go#L52

AlekSi: But where we enable CPU collector on macOS, for example?

AlekSi: Ah, it is enabled by default.

PR #348, opened 2020-02-25
PMM-5464 Update prometheus dependencies

artemgavrilov:
Update transitive Prometheus dependency `github.com/prometheus/common`

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/715

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -21,4 +21,4 @@ required = [
 # to prevent unexpected downgrades; see https://github.com/percona/exporter_shared/releases/tag/v0.6.0
 [[constraint]]
   name = "github.com/percona/exporter_shared"
-  version = "0.6.0"
+  version = "0.7.2"
--- a/services/prometheus/internal/common/config/http_config.go
+++ b/services/prometheus/internal/common/config/http_config.go
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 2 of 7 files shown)


services/prometheus/internal/common/config/http_config.go
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-// Original file: https://github.com/prometheus/common/blob/v0.9.1/config/http_config.go
+// Original file: https://github.com/prometheus/common/blob/v0.8.0/config/http_config.go
```
artemgavrilov: This file in both versions is identical, but Prometheus v1.16.0 depends on common v0.8.0.
So comment was updated to match consistent dependencies state.

artemgavrilov:
Readme updated

Gopkg.toml
```diff
@@ -21,4 +21,4 @@ required = [
 # to prevent unexpected downgrades; see https://github.com/percona/exporter_shared/releases/tag/v0.6.0
 [[constraint]]
   name = "github.com/percona/exporter_shared"
-  version = "0.6.0"
+  version = "0.7.2"
```
BupycHuk: Why did you update exporter_shared version?

artemgavrilov: It has transitive dependency on `github.com/prometheus/common`. So firstly I updated `prometheus/common` in `exporter_shared` and then I updated `exporter_shared` in `mongodb_exporter`.

PR #351, opened 2020-03-05
PMM-5112 Filter by type.

BupycHuk:


models/agent_helpers.go
```diff
@@ -101,12 +107,20 @@ func FindAgentsByIDs(q *reform.Querier, ids []string) ([]*Agent, error) {
...
-	structs, err := q.FindAllFrom(AgentTable, "node_id", nodeID)
+	conditions := []string{
+		"node_id = $1",
+	}
+	args := []interface{}{nodeID}
+	if agentType != nil {
+		conditions = append(conditions, "agent_type = $2")
+		args = append(args, agentType)
+	}
+	structs, err := q.SelectAllFrom(AgentTable, fmt.Sprintf("WHERE %s ORDER BY agent_id", strings.Join(conditions, " AND ")), args...)
```
golangcibot: G201: SQL string formatting (from `gosec`)

models/agent_helpers.go
```diff
@@ -120,12 +134,20 @@ func FindAgentsForNode(q *reform.Querier, nodeID string) ([]*Agent, error) {
...
-	structs, err := q.FindAllFrom(AgentTable, "service_id", serviceID)
+	conditions := []string{
+		"service_id = $1",
+	}
+	args := []interface{}{serviceID}
+	if agentType != nil {
+		conditions = append(conditions, "agent_type = $2")
+		args = append(args, agentType)
+	}
+	structs, err := q.SelectAllFrom(AgentTable, fmt.Sprintf("WHERE %s ORDER BY agent_id", strings.Join(conditions, " AND ")), args...)
```
golangcibot: G201: SQL string formatting (from `gosec`)

models/agent_helpers.go
```diff
@@ -139,12 +161,20 @@ func FindAgentsForService(q *reform.Querier, serviceID string) ([]*Agent, error)
...
-	structs, err := q.SelectAllFrom(AgentTable, "WHERE pmm_agent_id = $1 ORDER BY agent_id", pmmAgentID)
+	conditions := []string{
+		"pmm_agent_id = $1",
+	}
+	args := []interface{}{pmmAgentID}
+	if agentType != nil {
+		conditions = append(conditions, "agent_type = $2")
+		args = append(args, agentType)
+	}
+	structs, err := q.SelectAllFrom(AgentTable, fmt.Sprintf("WHERE %s ORDER BY agent_id", strings.Join(conditions, " AND ")), args...)
```
golangcibot: G201: SQL string formatting (from `gosec`)

services/inventory/grpc/agents_server.go
```diff
@@ -19,7 +19,7 @@ package grpc
 import (
 	"context"
 	"fmt"
-
+	"github.com/percona/pmm-managed/models"
```
golangcibot: File is not `goimports`-ed with -local github.com/percona/pmm-managed (from `goimports`)
```suggestion

```

models/agent_helpers.go
```diff
@@ -45,8 +45,14 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
...
-func FindAllAgents(q *reform.Querier) ([]*Agent, error) {
-	structs, err := q.SelectAllFrom(AgentTable, "ORDER BY agent_id")
+func FindAllAgents(q *reform.Querier, agentType *AgentType) ([]*Agent, error) {
+	var conditions []string
+	var args []interface{}
+	if agentType != nil {
+		conditions = append(conditions, "WHERE agent_type = $1")
+		args = append(args, agentType)
+	}
+	structs, err := q.SelectAllFrom(AgentTable, fmt.Sprintf("%s ORDER BY agent_id", strings.Join(conditions, " AND ")), args...)
```
askomorokhov: `conditions = append(conditions, "WHERE agent_type = $1")` in couple with `strings.Join(conditions, " AND ")` looks strange - cannot be multiple `WHERE` in query.

BupycHuk: No, it's not possible, at least for now. Made it this way to have the same structure with queries in other helpers.

models/agent_helpers.go
askomorokhov: @BupycHuk, what is your opinion about to form this part of the query in go "text/template"?
IMHO it can be more readable.

BupycHuk: using text/template here is overengineering, IMHO

models/agent_helpers.go
```diff
@@ -45,8 +45,14 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
 }
 
 // FindAllAgents returns all Agents.
-func FindAllAgents(q *reform.Querier) ([]*Agent, error) {
-	structs, err := q.SelectAllFrom(AgentTable, "ORDER BY agent_id")
+func FindAllAgents(q *reform.Querier, agentType *AgentType) ([]*Agent, error) {
```
askomorokhov: Maybe better FindAgentsByType ?

BupycHuk: no, because type must be nil

models/node_helpers.go
```diff
@@ -81,8 +81,14 @@ func checkUniqueNodeInstanceRegion(q *reform.Querier, instance, region string) e
 }
 
 // FindAllNodes returns all Nodes.
-func FindAllNodes(q *reform.Querier) ([]*Node, error) {
-	structs, err := q.SelectAllFrom(NodeTable, "ORDER BY node_id")
+func FindAllNodes(q *reform.Querier, nodeType *NodeType) ([]*Node, error) {
+	var conditions []string
+	var args []interface{}
+	if nodeType != nil {
+		conditions = append(conditions, "WHERE node_type = $1")
```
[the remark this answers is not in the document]
percona-csalguero: I understand that if we remove the WHERE from the conditions array, later we will need an if to check if we should add the WHERE clause or not but, maybe we don't need an array here. Just a string would do the work

models/agent_helpers.go
```diff
@@ -120,12 +134,20 @@ func FindAgentsForNode(q *reform.Querier, nodeID string) ([]*Agent, error) {
 }
 
 // FindAgentsForService returns all Agents providing insights for given Service.
-func FindAgentsForService(q *reform.Querier, serviceID string) ([]*Agent, error) {
+func FindAgentsForService(q *reform.Querier, serviceID string, agentType *AgentType) ([]*Agent, error) {
```
askomorokhov: `FindAllAgents`, `FindAgentsForService`, `FindAgentsForNode`, `FindAgentsRunningByPMMAgent` - looks almost the same. Do we really need such a zoo of the methods - or it is time to join them in a single one?

BupycHuk: looks like we can join them

services/inventory/grpc/services_server.go
```diff
@@ -36,10 +36,26 @@ func NewServicesServer(s *inventory.ServicesService) inventorypb.ServicesServer
...
+	inventorypb.ServiceType_POSTGRESQL_SERVICE: models.PostgreSQLServiceType,
+	inventorypb.ServiceType_PROXYSQL_SERVICE:   models.ProxySQLServiceType,
+}
+
+func serviceType(serviceType inventorypb.ServiceType) *models.ServiceType {
+	if serviceType == inventorypb.ServiceType_SERVICE_TYPE_INVALID {
+		return nil
+	}
+	result := serviceTypes[serviceType]
+	return &result
```
percona-csalguero: ServiceType is an int. 
If the function is called with an invalid parameter, what `return &result` would return?
The same way we return a nil for `serviceType == inventorypb.ServiceType_SERVICE_TYPE_INVALID`, if the function parameter is not valid, we should return a nil, maybe something like:
```
result, ok := serviceTypes[serviceType]
if !ok {
    return nil
}
return &result
```

BupycHuk: grpc should protect us from getting service types not declared in our proto files.
So if we missed any service type defined in proto files it's better to have panic message instead of silently going forward, IMHO.

models/agent_helpers.go
```diff
@@ -44,9 +44,37 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
...
+	}
+	if filters.ServiceID != "" {
+		conditions = append(conditions, fmt.Sprintf("service_id = %s", q.Placeholder(idx)))
+		args = append(args, filters.ServiceID)
+		idx++
+	}
+	if filters.PMMAgentID != "" {
+		conditions = append(conditions, fmt.Sprintf("pmm_agent_id = %s", q.Placeholder(idx)))
+		args = append(args, filters.PMMAgentID)
+		idx++
```
golangcibot: ineffectual assignment to `idx` (from `ineffassign`)

models/agent_helpers.go
```diff
@@ -44,9 +44,37 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
...
+	}
+
+	var whereClause string
+	if len(conditions) != 0 {
+		whereClause = fmt.Sprintf("WHERE %s", strings.Join(conditions, " AND "))
```
golangcibot: G201: SQL string formatting (from `gosec`)

models/node_helpers.go
```diff
@@ -80,9 +80,19 @@ func checkUniqueNodeInstanceRegion(q *reform.Querier, instance, region string) e
...
+func FindNodes(q *reform.Querier, filters NodeFilters) ([]*Node, error) {
+	var conditions []string
+	var args []interface{}
+	if filters.NodeType != nil {
+		conditions = append(conditions, "node_type = $1")
+		args = append(args, filters.NodeType)
+	}
+	var whereClause string
+	if len(conditions) != 0 {
+		whereClause = fmt.Sprintf("WHERE %s", strings.Join(conditions, " AND "))
```
golangcibot: G201: SQL string formatting (from `gosec`)

models/service_helpers.go
```diff
@@ -55,9 +55,26 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
...
+		idx++
+	}
+	if filters.NodeID != "" {
+		conditions = append(conditions, fmt.Sprintf("node_id = %s", q.Placeholder(idx)))
+		args = append(args, filters.NodeID)
+		idx++
+	}
+	var whereClause string
+	if len(conditions) != 0 {
+		whereClause = fmt.Sprintf("WHERE %s", strings.Join(conditions, " AND "))
```
golangcibot: G201: SQL string formatting (from `gosec`)

models/service_helpers.go
```diff
@@ -55,9 +55,26 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
...
+	idx := 1
+	if filters.ServiceType != nil {
+		conditions = append(conditions, fmt.Sprintf("service_type = %s", q.Placeholder(idx)))
+		args = append(args, filters.ServiceType)
+		idx++
+	}
+	if filters.NodeID != "" {
+		conditions = append(conditions, fmt.Sprintf("node_id = %s", q.Placeholder(idx)))
+		args = append(args, filters.NodeID)
+		idx++
```
golangcibot: ineffectual assignment to `idx` (from `ineffassign`)

models/node_helpers.go
```diff
@@ -80,9 +80,19 @@ func checkUniqueNodeInstanceRegion(q *reform.Querier, instance, region string) e
...
+	}
+	structs, err := q.SelectAllFrom(NodeTable, fmt.Sprintf("%s ORDER BY node_id", whereClause), args...)
```
askomorokhov: I see it looks like follow a general pattern, but here it a bit overcomplication.
Maybe it could be simplified somehow like this? 
```suggestion
	var args []interface{}
	tail := "ORDER BY node_id"
	if filters.NodeType != nil {
	        tail = "WHERE node_type = $1 " + tail
	        args = []interface{}{filters.NodeType}
	}
	structs, err := q.SelectAllFrom(NodeTable, tail, args...)
```

models/node_helpers.go
```diff
@@ -80,9 +80,19 @@ func checkUniqueNodeInstanceRegion(q *reform.Querier, instance, region string) e
 	}
 }
 
-// FindAllNodes returns all Nodes.
-func FindAllNodes(q *reform.Querier) ([]*Node, error) {
-	structs, err := q.SelectAllFrom(NodeTable, "ORDER BY node_id")
+// FindNodes returns Nodes by filters.
+func FindNodes(q *reform.Querier, filters NodeFilters) ([]*Node, error) {
```
askomorokhov: `filters NodeFilters` can be simplified to `node *NodeType`?

models/agent_helpers.go
```diff
@@ -44,9 +44,45 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
...
+		}
+		conditions = append(conditions, fmt.Sprintf("pmm_agent_id = %s", q.Placeholder(idx)))
+		args = append(args, filters.PMMAgentID)
+	}
+
+	var whereClause string
+	if len(conditions) != 0 {
+		whereClause = fmt.Sprintf("WHERE %s", strings.Join(conditions, " AND "))
+	}
+	structs, err := q.SelectAllFrom(AgentTable, fmt.Sprintf("%s ORDER BY agent_id", whereClause), args...)
```
askomorokhov: 👍

Makefile.devcontainer
```diff
@@ -25,7 +25,7 @@ install-race:                   ## Install pmm-managed and pmm-managed-init bina
 	go build -race -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-init ./cmd/pmm-managed-init
 
 PMM_TEST_PACKAGES ?= ./...
-PMM_TEST_FLAGS ?= -timeout=30s
+PMM_TEST_FLAGS ?= -timeout=60s -count 1
```
AlekSi: I don't think we should pass tests cache all the time – it us actually useful.

models/agent_helpers.go
```diff
@@ -419,3 +399,15 @@ func RemoveAgent(q *reform.Querier, id string, mode RemoveMode) (*Agent, error)
 
 	return a, nil
 }
+
+// AgentFilters represents filters for agents list.
+type AgentFilters struct {
```
AlekSi: Let's move definition before the first usage.

models/agent_helpers.go
```diff
@@ -44,9 +44,45 @@ func checkUniqueAgentID(q *reform.Querier, id string) error {
...
+	}
+	if filters.ServiceID != "" {
+		if _, err := FindServiceByID(q, filters.ServiceID); err != nil {
+			return nil, err
+		}
+		conditions = append(conditions, fmt.Sprintf("service_id = %s", q.Placeholder(idx)))
+		args = append(args, filters.ServiceID)
+		idx++
+	}
+	if filters.PMMAgentID != "" {
```
AlekSi: Is there a reason for different order of those `if`s and fields in AgentFilters?

models/node_helpers.go
```diff
@@ -80,9 +80,15 @@ func checkUniqueNodeInstanceRegion(q *reform.Querier, instance, region string) e
 	}
 }
 
-// FindAllNodes returns all Nodes.
-func FindAllNodes(q *reform.Querier) ([]*Node, error) {
-	structs, err := q.SelectAllFrom(NodeTable, "ORDER BY node_id")
+// FindNodes returns Nodes by filters.
+func FindNodes(q *reform.Querier, nodeType *NodeType) ([]*Node, error) {
```
AlekSi: NodeFilters instead of NodeType?

services/inventory/grpc/agents_server.go
```diff
@@ -34,12 +35,35 @@ func NewAgentsServer(s *inventory.AgentsService) inventorypb.AgentsServer {
 	return &agentsServer{s}
 }
 
+var agentTypes = map[inventorypb.AgentType]models.AgentType{
```
AlekSi: Don't we have that mapping already?

models/node_helpers.go
AlekSi: But then NodeFilters is not used?

models/node_helpers.go
BupycHuk: I did that way, but for now we have only one filter and it doesn't look like we will have other ones. So decided to use NodeType instead of NodeFilters

AlekSi: But then NodeFilters is not used anywhere?

services/inventory/grpc/agents_server.go
BupycHuk: No, we don't. Map we've made for PMM-5003 is different and we can't use it here.

PR #353, opened 2020-03-06
PMM-5540 Prometheus configuration

artemgavrilov:
- https://jira.percona.com/browse/PMM-5540

Use https://github.com/percona/promconfig for Prometheus configurations structures

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/761


services/prometheus/scrape_configs_test.go
```diff
@@ -301,16 +298,16 @@ func TestScrapeConfig(t *testing.T) {
...
 						Password: "/agent_id/75bb30d3-ef4a-4147-97a8-621a996611dd",
 					},
 				},
-				ServiceDiscoveryConfig: sd_config.ServiceDiscoveryConfig{
-					StaticConfigs: []*targetgroup.Group{{
-						Targets: []model.LabelSet{{"__address__": "4.5.6.7:12345"}},
-						Labels: model.LabelSet{
+				ServiceDiscoveryConfig: config.ServiceDiscoveryConfig{
+					StaticConfigs: []*config.Group{{
+						Targets: []string{ "4.5.6.7:12345"},
```
golangcibot: File is not `goimports`-ed with -local github.com/percona/pmm-managed (from `goimports`)
```suggestion
						Targets: []string{"4.5.6.7:12345"},
```

services/grafana/auth_server.go
```diff
@@ -131,7 +131,7 @@ func (s *AuthServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	if err := extractOriginalRequest(req); err != nil {
 		s.l.Warnf("Failed to parse request: %s.", err)
-		rw.WriteHeader(400)
+		rw.WriteHeader(http.StatusBadRequest)
```
artemgavrilov: @AlekSi But `http.StatusBadRequest` is same 400 `int` but defined as constant. Nothing really changed, or I missed something?

services/prometheus/internal/config/global.go
```diff
@@ -0,0 +1,18 @@
+package config
+
+import "github.com/prometheus/common/model"
```
artemgavrilov: The only import left from prometheus because of custom durations format. Can be removed if we copy parsing function from `prometheus/common`, however, we have transitive depenendency on that package anyway. So I think it's ok for now to use this type.

services/grafana/auth_server.go
AlekSi: Is 400 less explicit than `http.StatusBadRequest`, 200 – less than `http.StatusOK` in the context of HTTP protocol responses? Can't say I necessary agree with that in general (for example, I remember the difference between 301/302/303/307/308 by codes, but constant names are… not very descriptive 😄), but that's a different topic.

But in that particular case 400/200 is much explicit since nginx configuration uses numbers (and can't use named constants), and that code is tightly coupled with it.

So let's not use named constants for status codes there.

artemgavrilov: Ok, I get it. Will mute linter)

AlekSi:
https://github.com/percona/promconfig has arrived

services/prometheus/scrape_configs.go
```diff
@@ -25,40 +25,33 @@ import (
...
 	switch {
 	case interval <= 2*time.Second:
-		return model.Duration(time.Second)
+		return config.Duration(time.Second)
 	case interval <= 10*time.Second:
-		return model.Duration(interval - time.Second)
+		return config.Duration(interval - time.Second)
 	default:
-		return model.Duration(10 * time.Second)
+		return config.Duration(10 * time.Second)
```
golangcibot: mnd: Magic number: 10, in <argument> detected (from `gomnd`)

services/prometheus/scrape_configs.go
```diff
@@ -25,40 +25,34 @@ import (
...
-	"github.com/percona/pmm-managed/services/prometheus/internal/prometheus/config"
-	sd_config "github.com/percona/pmm-managed/services/prometheus/internal/prometheus/discovery/config"
-	"github.com/percona/pmm-managed/services/prometheus/internal/prometheus/discovery/targetgroup"
 )
 
-const addressLabel = model.LabelName(model.AddressLabel)
-
 // scrapeTimeout returns default scrape timeout for given scrape interval.
-func scrapeTimeout(interval time.Duration) model.Duration {
+// nolint:gomnd
```
AlekSi: > Use //nolint instead of // nolint because machine-readable comments should have no space by Go convention.

From https://github.com/golangci/golangci-lint#nolint

artemgavrilov: Fixed

PR #360, opened 2020-03-13
PMM-1962 Implemented Grafana health check

percona-csalguero:


services/grafana/client.go
```diff
@@ -352,6 +353,52 @@ func (c *Client) findAnnotations(ctx context.Context, from, to time.Time) ([]ann
...
 }
 
+type grafanaHealthResponse struct {
+	Commit   string `json:"commit"`
+	Database string `json:"database"`
+	Version  string `json:"version"`
+}
+
+// Check calls Grafana API to check its status
+func (c Client) Check(ctx context.Context) error {
```
golangcibot: `(Client).Check` - `ctx` is unused (from `unparam`)

services/server/server.go
```diff
@@ -255,11 +258,27 @@ func (s *Server) Version(ctx context.Context, req *serverpb.VersionRequest) (*se
 // Readiness returns an error when some PMM Server component is not ready yet or is being restarted.
 // It can be used as for Docker health check or Kubernetes readiness probe.
 func (s *Server) Readiness(ctx context.Context, req *serverpb.ReadinessRequest) (*serverpb.ReadinessResponse, error) {
-	// TODO https://jira.percona.com/browse/PMM-1962
 
-	if err := s.prometheus.Check(ctx); err != nil {
+	errs := make([]error, 0)
```
BupycHuk: It looks too complicated.
We don't have to return all errors. We can return only first error.

services/server/server_test.go
```diff
@@ -35,6 +35,10 @@ import (
 	"github.com/percona/pmm-managed/utils/tests"
 )
 
+type checkerMock struct{}
```
BupycHuk: It's better to generate mock using mockery like here https://github.com/percona/pmm-managed/pull/360/files#diff-60d6255f64d5cf44c23f4f2361cc1a30R29

services/server/server.go
percona-csalguero: I think we should return all errors at once to make debugging easier and faster otherwise, supose both, prometheus and grafana are failing, you would need to run the check twice to get both errors.

BupycHuk: We usually need it for pmm-update and healthcheck, so user don't fix it manually. Am I right, @AlekSi

AlekSi: In general, I agree with @percona-csalguero.

But in that particular case we should expose less information, not more – that API is available without authentication:
```proto
message ReadinessRequest {}

message ReadinessResponse {
  // This probe is available without authentication, so it should not contain any data.
}
```

So our ideal error message could be "Problems with: Prometheus, Grafana.", not more. And error message like "not ready" or "error" is better than more details.

services/server/server.go
```diff
@@ -255,11 +258,27 @@ func (s *Server) Version(ctx context.Context, req *serverpb.VersionRequest) (*se
...
+
+	var err error
+	// concat the errors so the client will receive all the errors at once
+	for _, e := range errs {
+		if e != nil {
+			if err == nil {
+				err = e
+				continue
+			}
+			err = fmt.Errorf("%s, %w", err, e)
```
AlekSi: We definitely don't want to start using `%w` – it is not compatible with pkg/errors, and migration is not trivial.

services/server/server_test.go
```diff
@@ -41,9 +41,11 @@ func TestServer(t *testing.T) {
 		require.NoError(t, sqlDB.Close())
 	}()
 
+	checker := new(MockChecker)
```
AlekSi: But why we mock it? It should work in devcontainer?

services/grafana/client.go
```diff
@@ -352,6 +353,52 @@ func (c *Client) findAnnotations(ctx context.Context, from, to time.Time) ([]ann
...
+		return fmt.Errorf("cannot read Grafana Health API response")
+	}
+
+	var status grafanaHealthResponse
+	if err := json.Unmarshal(buf, &status); err != nil {
+		return fmt.Errorf("cannot parse Grafana Health API response: %s", err)
+	}
+
+	if strings.ToLower(status.Database) != "ok" {
+		return fmt.Errorf("grafana is up but the database is not ok. Database status is %s", status.Database)
```
AlekSi: For this error message, and above, we should log error, but not return it.

services/server/server.go
```diff
@@ -255,11 +258,20 @@ func (s *Server) Version(ctx context.Context, req *serverpb.VersionRequest) (*se
...
-		return nil, err
+		fs = append(fs, "Prometheus")
+	}
+
+	if err := s.grafanaClient.Check(ctx); err != nil {
+		fs = append(fs, "Grafana")
 	}
+
+	if len(fs) > 0 {
+		return nil, fmt.Errorf("Failing services: %s", strings.Join(fs, ", "))
```
golangcibot: ST1005: error strings should not be capitalized (from `stylecheck`)

BupycHuk:
@percona-csalguero please fix travis ci build.
You forgot to add line here https://github.com/percona/pmm-managed/pull/360/files#diff-60d6255f64d5cf44c23f4f2361cc1a30R30

services/server/deps.go
```diff
@@ -29,12 +29,19 @@ import (
 //go:generate mockery -name=prometheusService -case=snake -inpkg -testonly
 //go:generate mockery -name=supervisordService -case=snake -inpkg -testonly
 //go:generate mockery -name=telemetryService -case=snake -inpkg -testonly
+//go:generate mockery -name=Checker -case=snake -inpkg -testonly
```
BupycHuk: let's keep the same order that we have below and make it internal

services/server/server_test.go
percona-csalguero: The mock here is because of:
1. We have other mocks so, to keep it consistent
2. To avoid dependencies in tests.

services/server/deps.go
```diff
@@ -26,15 +26,22 @@ import (
 	"github.com/percona/pmm-managed/models"
 )
 
+//go:generate mockery -name=Checker -case=snake -inpkg -testonly
```
BupycHuk: checker

services/server/server_test.go
```diff
@@ -41,9 +41,12 @@ func TestServer(t *testing.T) {
 		require.NoError(t, sqlDB.Close())
 	}()
 
+	// use a mockup to avoid impor test to fail since this is the server pkg and not server_test
```
BupycHuk: import

services/grafana/client.go
BupycHuk: why?

AlekSi: For the same reason as above:
> But in that particular case we should expose less information, not more – that API is available without authentication

Status can contain too much.

BupycHuk: Yeah, but shouldn't we return error?

AlekSi: We should log "grafana is up but the database is not ok. Database status is %s", but not return that error message; instead we should return something like "not ok" without any details.

services/grafana/client.go
```diff
@@ -352,6 +353,26 @@ func (c *Client) findAnnotations(ctx context.Context, from, to time.Time) ([]ann
...
+// Check calls Grafana API to check its status
+func (c Client) Check(ctx context.Context) error {
+	var status grafanaHealthResponse
+	if err := c.do(ctx, "GET", "/api/health", http.Header{}, nil, &status); err != nil {
+		return fmt.Errorf("error calling Grafana API: %s", err)
+	}
+
+	if strings.ToLower(status.Database) != "ok" {
+		logrus.Errorf("grafana is up but the database is not ok. Database status is %s", status.Database)
+	}
```
AlekSi: See https://github.com/percona/pmm-managed/pull/360#discussion_r395740362

We should not return details, by we still should return the fact that Grafana is not ok.

Probably it is easier to return bool.

services/grafana/client.go
```diff
@@ -352,6 +353,26 @@ func (c *Client) findAnnotations(ctx context.Context, from, to time.Time) ([]ann
...
+	Commit   string `json:"commit"`
+	Database string `json:"database"`
+	Version  string `json:"version"`
+}
+
+// Check calls Grafana API to check its status
+func (c Client) Check(ctx context.Context) error {
+	var status grafanaHealthResponse
+	if err := c.do(ctx, "GET", "/api/health", http.Header{}, nil, &status); err != nil {
+		return fmt.Errorf("error calling Grafana API: %s", err)
```
AlekSi: No, we should not return internal details.

services/grafana/client.go
```diff
@@ -352,6 +353,26 @@ func (c *Client) findAnnotations(ctx context.Context, from, to time.Time) ([]ann
...
+type grafanaHealthResponse struct {
+	Commit   string `json:"commit"`
+	Database string `json:"database"`
+	Version  string `json:"version"`
+}
+
+// Check calls Grafana API to check its status
+func (c Client) Check(ctx context.Context) error {
+	var status grafanaHealthResponse
+	if err := c.do(ctx, "GET", "/api/health", http.Header{}, nil, &status); err != nil {
```
AlekSi: Just `nil` instead of `http.Header{}`.

services/server/deps.go
```diff
@@ -26,15 +26,22 @@ import (
...
+//go:generate mockery -name=checker -case=snake -inpkg -testonly
 //go:generate mockery -name=prometheusService -case=snake -inpkg -testonly
 //go:generate mockery -name=supervisordService -case=snake -inpkg -testonly
 //go:generate mockery -name=telemetryService -case=snake -inpkg -testonly
 
+// Checker interface wraps all services that implements the Check method to report the
+// service health for the Readiness check
+type checker interface {
+	Check(ctx context.Context) error
+}
```
AlekSi: Sorry, I think that both interface name and method name are too generic: we already have several checkers like AWSInstanceChecker, and we will have more _checks_ and _checkers_ with a coming feature.

Something like `IsReady(context.Context) bool`?

AlekSi: But more importantly, deps.go is for mocking dependencies, not for generic mocks and interfaces.
I assume you want to have `type grafanaClient interface` there like all other dependencies. And you want to accept Grafana client (a real one or mock), not any checker, just like you want to accept Prometheus service (a real one or mock).

services/grafana/client.go
BupycHuk: Yeah, we shouldn't expose error to users, but the grafana checker should return an error, not ignore it. So we can format our message in the API layer.

services/server/deps.go
percona-csalguero: The idea here was to be able to have a common interface for all services implementing checks. 
Can we name it something like ServiceChecker and keep it here?
If you take a look a couple of lines below, now Prometheus service also implements the `checker` interface. 
We discussed with Nurlan that in the future we might want to add check for other services as well so we implemented checker interface here.

BupycHuk: @AlekSi so we can provide list of checkers as field of struct you suggested here https://github.com/percona/pmm-managed/pull/360#discussion_r396991827
to loop over them and return a message with non-ready components. In this case we should add extra method to interface `ServiceName() string`
What do you think about that?

BupycHuk: @AlekSi ^

services/server/server.go
```diff
@@ -82,24 +83,35 @@ type pmmUpdateAuth struct {
...
+		db:                in.DB,
+		prometheus:        in.Prometheus,
+		supervisord:       in.Supervisord,
+		telemetryService:  in.TelemetryService,
+		checker:           in.Checker,
 		l:                 logrus.WithField("component", "server"),
 		pmmUpdateAuthFile: path,
-		alertManagerFile:  alertManagerFile,
+		alertManagerFile:  in.AlertManagerFile,
+		grafanaClient:     in.GrafanaClient,
```
BupycHuk, line 114: Maybe it's better to move grafanaClient a few lines above with other services.

services/server/server.go
```diff
@@ -62,6 +62,7 @@ type Server struct {
 
 	pmmUpdateAuthFileM sync.Mutex
 	pmmUpdateAuthFile  string
+	grafanaClient      serviceChecker
```
BupycHuk, line 65: let's move it to other services

services/server/server.go
BupycHuk: here and in other places

services/server/deps.go
AlekSi: I think it will be easier for me to make a PR to show what I mean. :)

In the meantime, please rename `Check` method to `CheckHealth` everywhere – word "check" becomes overloaded with the introduction of check.percona.con, Security Threat Tool checks, etc.

AlekSi: LGTM with recent changes.

PR #363, opened 2020-03-20
PMM-5042 Add socket connection  to mysql service.

askomorokhov:
https://jira.percona.com/browse/PMM-5042

Deps:
- https://github.com/Percona-Lab/pmm-api-tests/pull/86
- https://github.com/percona/pmm/pull/506

FB:
 - https://github.com/Percona-Lab/pmm-submodules/pull/764


utils/validators/mysql.go
```diff
@@ -0,0 +1,39 @@
...
+package validators
+
+import (
+	"github.com/pkg/errors"
+)
+
+// ValidateMySQLConnectionOptions validates MySQL connection options.
+func ValidateMySQLConnectionOptions(socket, host *string, port *uint16) error {
+	if (host == nil || port == nil) && socket == nil {
+		return errors.New("Address (with port) or socket is required.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

utils/validators/mysql.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	if (host != nil || port != nil) && socket != nil {
+		return errors.New("Setting both address (with port) and socket in once is disallowed.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

utils/validators/mysql.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	if host != nil && port == nil {
+		return errors.New("Port is required.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

utils/validators/mysql.go
```diff
@@ -0,0 +1,50 @@
...
+
+		if port == nil {
+			return status.Error(codes.InvalidArgument, "port is required with non empty host")
+		}
+
+		if int(*port) <= 0 {
+			return status.Errorf(codes.InvalidArgument, "invalid field Port: value '%d' must be greater than '0'", port)
+		}
+
+		if int(*port) > 65535 {
```
golangcibot: mnd: Magic number: 65535, in <condition> detected (from `gomnd`)

models/service_helpers.go
```diff
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/percona/pmm-managed/utils/validators"
```
BupycHuk: please move it below as third block

utils/validators/mysql.go
```diff
@@ -0,0 +1,50 @@
...
+// Package validators contains settings validators.
+package validators
+
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ValidateMySQLConnectionOptions validates MySQL connection options.
+func ValidateMySQLConnectionOptions(socket, host *string, port *uint16) error {
```
BupycHuk: It's called `ValidateMySQLConnectionOptions`, but it validates all services, not only MySQL.

models/service_helpers.go
```diff
@@ -179,6 +182,10 @@ func AddNewService(q *reform.Querier, serviceType ServiceType, params *AddDBMSSe
 		return nil, err
 	}
 
+	if err := validators.ValidateMySQLConnectionOptions(params.Socket, params.Address, params.Port); err != nil {
```
BupycHuk: Should we validate it twice? in API layer and here? 
Maybe it's better to move it above to not make unnecessary DB queries.

models/database.go
```diff
@@ -235,6 +235,11 @@ var databaseSchema = [][]string{
 			AND settings->'metrics_resolutions'->>'mr' = '5000000000'
 			AND settings->'metrics_resolutions'->>'lr' = '60000000000';`,
 	},
+
+	11: {
+		`ALTER TABLE services
+			ADD COLUMN socket VARCHAR CHECK (socket <> '')`,
```
BupycHuk: let's add test for this check

BupycHuk: Maybe we can write better check? like 
```
CHECK (
				(CASE WHEN socket IS NULL THEN 0 ELSE 1 END) +
  				(CASE WHEN address IS NULL THEN 0 ELSE 1 END)  = 1 )
```

askomorokhov: Added.

utils/validators/mysql.go
askomorokhov: Left name as is - according to the conversation

models/service_helpers.go
askomorokhov: Fixed.

BupycHuk:
@askomorokhov looks like you forgot to push changes here

models/database_test.go
```diff
@@ -152,7 +152,27 @@ func TestDatabaseChecks(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// Try to insert both address and socket
+		_, err = db.Exec(
+			"INSERT INTO services (service_id, service_type, service_name, node_id, environment, cluster, replication_set, address, port, socket, created_at, updated_at) "+
+				"VALUES ('/service_id/1', 'mysql', 'name', '/node_id/1', '', '', '', '10.10.10.10', 3306, '/var/run/mysqld/mysqld.sock', $1, $2)",
+			now, now,
+		)
+		require.Error(t, err, `pq: new row for relation "services" violates check constraint "address_socket_check"`)
+
```
golangcibot: unnecessary trailing newline (from `whitespace`)

BupycHuk: let's try with both empty

BupycHuk: can we also add the check that if address is not empty then port is also shouldn't be empty

models/dsn_helpers_test.go
```diff
@@ -105,7 +107,7 @@ func TestFindDSNByServiceID(t *testing.T) {
 
 		dsn, err := models.FindDSNByServiceIDandPMMAgentID(q, "S2", "PA1", "test")
 		require.NoError(t, err)
-		expected := "pmm-user@tcp(127.0.0.1:0)/test?timeout=1s"
```
BupycHuk: it was a bug :)

models/dsn_helpers_test.go
```diff
@@ -105,7 +107,7 @@ func TestFindDSNByServiceID(t *testing.T) {
...
+		expected := "pmm-user@tcp(127.0.0.1:3306)/test?timeout=1s"
```
BupycHuk: check for dsn for services with socket?

models/database_test.go
askomorokhov: Added

models/dsn_helpers_test.go
askomorokhov: Added.

models/agent_model_test.go
```diff
@@ -73,6 +73,28 @@ func TestAgent(t *testing.T) {
...
+		agent := &Agent{
+			Username: pointer.ToString("username"),
+			Password: pointer.ToString("s3cur3 p@$$w0r4."),
+		}
+		service := &Service{
+			Socket: pointer.ToString("/var/run/mysqld/mysqld.sock"),
+		}
+		for typ, expected := range map[AgentType]string{
+			MySQLdExporterType:          "username:s3cur3 p@$$w0r4.@unix(/var/run/mysqld/mysqld.sock)/database?timeout=1s",
+			ProxySQLExporterType:        "username:s3cur3 p@$$w0r4.@unix(/var/run/mysqld/mysqld.sock)/database?timeout=1s",
```
BupycHuk: Does Proxy SQL support sockets?

services/management/service_test.go
```diff
@@ -106,6 +106,31 @@ func TestServiceService(t *testing.T) {
 			tests.AssertGRPCError(t, status.New(codes.InvalidArgument, `wrong service type`), err)
 		})
 
+		t.Run("MySQL Conflict socket and address", func(t *testing.T) {
```
BupycHuk: in other tests we were testing Remove, but here we test adding and it looks weird that we test models package in management package.

models/agent_model_test.go
askomorokhov: Yes, they should.

askomorokhov: https://github.com/sysown/proxysql/blob/63f09e6889464b61b1a0a6ad36ad082f6c0f5154/etc/proxysql.cnf#L54

services/management/service_test.go
askomorokhov: And what is your suggestion?

BupycHuk: Extract it from Remove part and maybe remove it since we have api tests

askomorokhov: Moved to service_helpers_test.go

models/agent_model.go
```diff
@@ -165,8 +166,12 @@ func (s *Agent) DSN(service *Service, dialTimeout time.Duration, database string
 		cfg := mysql.NewConfig()
 		cfg.User = username
 		cfg.Passwd = password
-		cfg.Net = "tcp"
-		cfg.Addr = net.JoinHostPort(host, strconv.Itoa(int(port)))
+		cfg.Net = "unix"
+		cfg.Addr = socket
+		if cfg.Addr == "" {
```
BupycHuk: maybe it's better to rewrite it like `if socket == "" {`
It'll improve readability

models/agent_model.go
```diff
@@ -189,8 +194,12 @@ func (s *Agent) DSN(service *Service, dialTimeout time.Duration, database string
...
```
BupycHuk:
```suggestion
		if socket == "" {
```

models/agent_model.go
BupycHuk:
```suggestion
		if socket == "" {
```

utils/validators/mysql.go
```diff
@@ -0,0 +1,45 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// Package validators contains settings validators.
```
AlekSi, line 17: This file has nothing to do with settings.

models/database.go
```diff
@@ -235,6 +235,21 @@ var databaseSchema = [][]string{
...
+			ADD COLUMN socket VARCHAR CONSTRAINT address_socket_check CHECK (
+				(address IS NOT NULL AND socket IS NULL) OR (address IS NULL AND socket IS NOT NULL)
+			);`,
+
+		`ALTER TABLE services
+			ADD CONSTRAINT address_port_check CHECK (
+				(address IS NULL AND port IS NULL) OR (address IS NOT NULL AND port IS NOT NULL)
+			),
+			ADD CONSTRAINT port_check CHECK (
+				port IS NULL OR (port > 0 AND port < 65535)
```
AlekSi, line 250: It _can_ be 65535. It should be `< 65536` or `<= 65535`.

BupycHuk:
@AlekSi We moved it there because we validate API inputs there like we did for inventory list endpoints and settings endpoints. Also, we have checks on the database layer and it makes no sense to add an extra check on pre-database layer.
Before we call `models.AddNewService` method we also call `nodeID` function which creates new node in transaction. So what's the reason to do so much operations just to get error that we got wrong input from the user. 
The difference between these validations and those in AddNewService that we don't have to check these ones in transaction.

PR #368, opened 2020-04-17
PMM-5695 Fix error messages.

BupycHuk:


models/service_helpers.go
```diff
@@ -55,6 +55,27 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
 	}
 }
 
+func validateDBConnectionOptions(socket, host *string, port *uint16) error {
+	if host == nil && socket == nil {
+		return status.Error(codes.InvalidArgument, "Neither socket nor address passed")
```
askomorokhov, line 60: Do we need a dot at the end of the sentence?

models/service_helpers.go
```diff
@@ -55,6 +55,27 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
...
+	}
+
+	if host != nil {
+		if socket != nil {
+			return status.Error(codes.InvalidArgument, "Socket and address cannot be specified together")
+		}
+
+		if port == nil {
+			return status.Errorf(codes.InvalidArgument, "Invalid field Port: value '%d' must be greater than '0'", port)
```
askomorokhov, line 69: Do we need to inherit this error message from the API?
Or need to change it to our own  - more human-readable?

models/service_helpers.go
```diff
@@ -55,6 +55,27 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
...
+		}
+	}
```
askomorokhov, line 71: Do we need to validate port range here?

models/service_helpers.go
percona-csalguero, line 60: Error messages shouldn't be capitalized

models/service_helpers.go
```diff
@@ -55,6 +55,27 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
...
 }
 
+func validateDBConnectionOptions(socket, host *string, port *uint16) error {
+	if host == nil && socket == nil {
+		return status.Error(codes.InvalidArgument, "Neither socket nor address passed")
+	}
+
+	if host != nil {
+		if socket != nil {
+			return status.Error(codes.InvalidArgument, "Socket and address cannot be specified together")
```
percona-csalguero, line 65: Same here. Error messages shouldn't be capitalized.

models/service_helpers.go
AlekSi: They should be for gRPC errors: https://jira.percona.com/browse/PMM-4695

models/service_helpers.go
BupycHuk: No, because uint16 range is between 0 and 65535

services/inventory/services_test.go
```diff
@@ -154,7 +154,7 @@ func TestServices(t *testing.T) {
 			Address:     pointer.ToString("127.0.0.1"),
 			Socket:      pointer.ToString("/var/run/mysqld/mysqld.sock"),
 		})
-		require.EqualError(t, err, "rpc error: code = InvalidArgument desc = setting both address and socket in once is disallowed")
+		require.EqualError(t, err, "rpc error: code = InvalidArgument desc = Socket and address cannot be specified together.")
```
AlekSi, line 157: I wonder why we don't use `tests.AssertGRPCError` there.

models/service_helpers.go
AlekSi: In theory, we could check that port is not 0 (`pointer.GetUint16(port) == 0`), but I think it is not possible there.

models/service_helpers.go
```diff
@@ -55,6 +55,27 @@ func checkServiceUniqueName(q *reform.Querier, name string) error {
 	}
 }
 
+func validateDBConnectionOptions(socket, host *string, port *uint16) error {
+	if host == nil && socket == nil {
+		return status.Error(codes.InvalidArgument, "Neither socket nor address passed.")
+	}
+
+	if host != nil {
```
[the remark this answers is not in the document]
BupycHuk: no, it will ruin all our logic

PR #371, opened 2020-04-20
SAAS-20 Data querying

artemgavrilov:
- https://jira.percona.com/browse/SAAS-20


services/agents/registry.go
```diff
@@ -768,6 +768,109 @@ func (r *Registry) StartMongoDBExplainAction(ctx context.Context, id, pmmAgentID
 	return nil
 }
 
+// StartMySQLQueryShowAction starts mySQL show query action on pmm-agent.
```
AlekSi, line 771: MySQL, not mySQL. The same thing below.

services/agents/registry.go
```diff
@@ -768,6 +768,109 @@ func (r *Registry) StartMongoDBExplainAction(ctx context.Context, id, pmmAgentID
...
+	agent, err := r.get(pmmAgentID)
+	if err != nil {
+		return err
+	}
+
+	agent.channel.SendRequest(aRequest)
+	return nil
+}
+
+// StartPostgreSQLQueryShowAction starts postgreSQL show query action on pmm-agent.
```
AlekSi, line 813: PostgreSQL, not postgreSQL. The same thing below.

services/checks/deps.go
```diff
@@ -0,0 +1,31 @@
...
+
+// registryService is a subset of methods of agents.Registry used by this package.
+// We use it instead of real type for testing and to avoid dependency cycle.
+type registryService interface {
+	StartMySQLQueryShowAction(ctx context.Context, id, pmmAgentID, dsn, query string) error
+	StartMySQLQuerySelectAction(ctx context.Context, id, pmmAgentID, dsn, query string) error
+	StartPostgreSQLQueryShowAction(ctx context.Context, id, pmmAgentID, dsn string) error
+	StartPostgreSQLQuerySelectAction(ctx context.Context, id, pmmAgentID, dsn, query string) error
+	StartMongoDBQueryGetParameterAction(ctx context.Context, id, pmmAgentID, dsn string) error
+}
```
AlekSi, line 32: Once more action there

services/checks/checks.go
```diff
@@ -94,32 +114,25 @@ func New(pmmVersion string) *Service {
...
+		if f := os.Getenv(envCheckFile); f != "" {
+			s.l.Warnf("Use local test checks file: %s.", f)
+			if err := s.loadLocalChecks(f); err != nil {
+				s.l.Errorf("Failed to load local checks file: %s.", err)
+			}
+		} else {
+			if err := s.downloadChecks(ctx); err != nil {
+				s.l.Errorf("Failed to download checks, %s.", err)
+			}
 		}
```
AlekSi, line 132: I would extract this block to separate function named something like `updateChecks`

services/checks/checks.go
```diff
@@ -94,32 +114,25 @@ func New(pmmVersion string) *Service {
...
-		}
-
-		s.updateChecks(checks)
-		return
-	}
-
 	ticker := time.NewTicker(s.interval)
 	defer ticker.Stop()
 
+	go s.checkResults(ctx)
```
AlekSi, line 120: I think we should do it synchronously/blocking while we can, and we can now. That will greatly simplify implementation.

services/checks/checks.go
```diff
@@ -131,13 +144,302 @@ func (s *Service) Run(ctx context.Context) {
...
+		if err != nil {
+			s.l.Errorf("Failed to prepare action result for agent %s: %s.", pmmAgentID, err)
+			continue
+		}
+		dsn := agent.DSN(services[*agent.ServiceID], 2*time.Second, "")
+
+		for _, c := range checks {
+			switch c.Type {
+			case check.MySQLShow:
+				if err := s.registry.StartMySQLQueryShowAction(context.Background(), r.ID, pmmAgentID, dsn, c.Query); err != nil {
```
AlekSi, line 266: And then wait for result synchronously, in the same goroutine. There no need to go async yet.

services/checks/checks.go
```diff
@@ -131,13 +144,302 @@ func (s *Service) Run(ctx context.Context) {
...
+	}
+
+	for _, agent := range agents {
+		pmmAgentID := *agent.PMMAgentID
+		r, err := models.CreateActionResult(s.db.Querier, pmmAgentID)
+		if err != nil {
+			s.l.Errorf("Failed to prepare action result for agent %s: %s.", pmmAgentID, err)
+			continue
+		}
+		dsn := agent.DSN(services[*agent.ServiceID], 2*time.Second, "")
```
AlekSi, line 261: `2*time.Second` should be a named constant shared by all actions there

services/checks/checks.go
```diff
@@ -131,13 +144,302 @@ func (s *Service) Run(ctx context.Context) {
...
+			default:
+				s.l.Errorf("Unknown mySQL check type: %s.", c.Type)
+				continue
+			}
+
+			res = append(res, checkTask{
+				resultID:   r.ID,
+				pmmAgentID: pmmAgentID,
+				serviceID:  *agent.ServiceID,
+				check:      &c})
```
AlekSi, line 284: Move `})` to the next line

services/agents/registry.go
artemgavrilov: Fixed

services/agents/registry.go
artemgavrilov: Fixed

services/checks/checks.go
artemgavrilov: Done, extracted to `grabChecks` function, as `updateChecks` func already exists and updatates slice filed where checks stored.

services/checks/checks.go
artemgavrilov: Done, check results until all results are ready or next checks run started (by timeout for now)

services/checks/checks.go
artemgavrilov: Done, use single `timeout` for all external connections (5 seconds)

services/checks/checks.go

services/agents/registry.go
AlekSi: No :) I will fix it

artemgavrilov: 🙈

services/checks/checks.go
```diff
@@ -131,13 +130,301 @@ func (s *Service) Run(ctx context.Context) {
...
-	defer s.m.RUnlock()
+	s.cm.Lock()
+	defer s.cm.Unlock()
 
 	r := make([]check.Check, 0, len(s.checks))
 	return append(r, s.checks...)
 }
 
+func (s *Service) processTasks(ctx context.Context, tasks []task) {
+	ticker := time.NewTicker(time.Minute)
```
AlekSi, line 141: We should stop it in `defer`.

services/checks/checks.go
```diff
@@ -131,13 +130,301 @@ func (s *Service) Run(ctx context.Context) {
...
-	s.m.RLock()
-	defer s.m.RUnlock()
+	s.cm.Lock()
+	defer s.cm.Unlock()
 
 	r := make([]check.Check, 0, len(s.checks))
 	return append(r, s.checks...)
 }
 
+func (s *Service) processTasks(ctx context.Context, tasks []task) {
```
AlekSi, line 140: Please start adding comments to private methods too. Once you start doing that, you may come up with better names and even ideas how to refactor it better.

In particular, I think this function should be simplified even more – just run query, wait for result (polling for result can be a separate function), execute script. Do not wait for several results in parallel, etc. Yes, it will be slow, but it will be simpler. Once we have a simple/linear/blocking implementation, we will make it more efficient.

services/checks/checks.go
AlekSi: And that value is too high: currently, all actions are limited to 10 seconds: https://github.com/percona/pmm-agent/blob/a2844d35b8e1a2fb25ef41f80d2258ab99a07721/client/client.go#L95

services/checks/checks.go
```diff
@@ -131,13 +130,301 @@ func (s *Service) Run(ctx context.Context) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), resultsTimeout)
+	defer cancel()
+
+	s.processTasks(ctx, tasks)
+}
+
+func (s *Service) executeMySQLChecks(checks []check.Check) ([]task, error) {
+	var res []task
+
+	agents, services, err := s.findAgentsAndServices(models.MySQLdExporterType)
```
AlekSi, line 218: `models.FindPMMAgentsForService`, then pick the first one. We should not use exporters there.

services/checks/checks.go
artemgavrilov: Set same value, 10 seconds for result awaiting

services/checks/checks.go
artemgavrilov: Oh, thanks, seems I finaly get a proper way to find agents.

artemgavrilov:
Made refactoring, simplified process flow.

services/checks/checks.go
artemgavrilov: Comments added, the whole service simplified

services/checks/checks.go
AlekSi: Changed it to 15 seconds to cascade timeouts

services/checks/checks.go
```diff
@@ -94,30 +106,16 @@ func New(pmmVersion string) *Service {
...
-		return
-	}
-
 	ticker := time.NewTicker(s.interval)
 	defer ticker.Stop()
 
 	for {
-		if err := s.downloadChecks(ctx); err != nil {
-			s.l.Errorf("Failed to download checks: %v", err)
+		// FIXME do that only if STT is enabled in settings https://jira.percona.com/browse/SAAS-30
```
AlekSi, line 113: We should not forget about that.

services/checks/checks.go
```diff
@@ -131,13 +129,308 @@ func (s *Service) Run(ctx context.Context) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		if !res.Done {
+			continue
+		}
+
+		// FIXME we still need to delete old result - they may never be done: https://jira.percona.com/browse/PMM-5840
+		if err = s.db.Delete(res); err != nil {
```
AlekSi, line 161: PTAL

AlekSi:
Waiting for CI just in case…

PR #372, opened 2020-04-21
SAAS-30 Execute checks

AlekSi:


services/alertmanager/alert.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 // Severity defines alert severity.
+// TODO Move to saas package: https://jira.percona.com/browse/SAAS-84
 type Severity string
```
AlekSi, line 31: @artemgavrilov ^

services/alertmanager/registry.go
```diff
@@ -76,14 +70,14 @@ func (r *registry) RemovePrefix(prefix string) {
...
-func (r *registry) Collect() ammodels.PostableAlerts {
+func (r *Registry) Collect() ammodels.PostableAlerts {
 	r.rw.RLock()
 	defer r.rw.RUnlock()
 
 	var res ammodels.PostableAlerts
 	now := time.Now()
 	for id, t := range r.times {
-		if t.After(now) {
+		if t.Before(now) {
```
artemgavrilov, line 80: Love it)

AlekSi: LOL. I remember agonising on that line, changing it several times.

@atymchuk that explains disappearing alerts

atymchuk: Cool, then the fix is not insanely difficult :)

services/alertmanager/alert.go
```diff
@@ -60,78 +47,9 @@ func (ap *AlertParams) validate() error {
 		return errors.New("empty Description")
 	}
 
-	if ap.Severity == "" {
-		ap.Severity = Info
+	if ap.Severity < check.Emergency || ap.Severity > check.Debug {
```
AlekSi, line 50: I think we should verify that only when we check results, so we don't have this check in two places.

services/checks/checks.go
```diff
@@ -110,12 +115,22 @@ func (s *Service) Run(ctx context.Context) {
 	defer ticker.Stop()
 
 	for {
-		// FIXME do that only if STT is enabled in settings https://jira.percona.com/browse/SAAS-30
-		if true {
+		var sttEnabled bool
```
AlekSi, line 123: Talked with Roma – yes, we should just delay the first check for one minute.

services/checks/checks.go
```diff
@@ -200,38 +215,32 @@ func (s *Service) executeMySQLChecks(ctx context.Context, checks []check.Check)
...
 	for _, target := range targets {
-		r, err := models.CreateActionResult(s.db.Querier, target.agentID)
+		r, err := models.CreateActionResult(s.db.Querier, target.agent.AgentID)
 		if err != nil {
-			s.l.Errorf("Failed to prepare action result for agent %s: %s.", target.agentID, err)
+			s.l.Errorf("Failed to prepare action result for agent %s: %s.", target.agent.AgentID, err)
 			continue
 		}
 
 		for _, c := range checks {
```
AlekSi, line 229: What I don't particularly like in that code that there we have two sets of checks: one is passed as an explicit parameter, the other is available as implicit parameters s.checks. I think we should get rid of `s.checks` completely and always values as parameters.

services/checks/checks.go
```diff
@@ -316,26 +319,82 @@ func (s *Service) executeMongoDBChecks(ctx context.Context, checks []check.Check
...
+}
+
 // target contains required info about check target
 type target struct {
-	agentID   string
-	serviceID string
-	dsn       string
+	agent   *models.Agent
+	service *models.Service
+	node    *models.Node
```
AlekSi, line 401: I think both `agent` and `node` are not actually used – they were added for demo.

services/alertmanager/alert.go
artemgavrilov: In fact that function is unused at all.

services/checks/checks.go
artemgavrilov: The idea with `s.checks` is that when for some reason we can't download new checks form server we still able to run previously downloaded stuff.

services/checks/checks.go
artemgavrilov: Timeout added

services/checks/checks.go
artemgavrilov: All that stuff needed to create labels set. Refactored, now targen contains labels set instead of node/service/agent.

artemgavrilov:
Resolving conflicts

services/alertmanager/alert.go
AlekSi: The whole file is not used: 59fe895223e541a6e82a4e9d9aef469a550bd815

services/checks/checks.go
```diff
@@ -166,55 +204,69 @@ func (s *Service) waitForResult(ctx context.Context, resultID string) (*models.A
...
-		return res, nil
+		return out, nil
 	}
 }
 
 // executeChecks runs all available checks for all reachable services.
 func (s *Service) executeChecks(ctx context.Context) {
+	s.l.Info("Executing checks...")
+
 	mySQLChecks, postgreSQLChecks, mongoDBChecks := s.groupChecksByDB(s.checks)
```
AlekSi, line 220: That's why state is bad :) We are not holding a lock there.
Will fix.

artemgavrilov: oops, yeah, checks should be accessed via `s.Checks` method

PR #374, opened 2020-04-23
PMM-5779 Add external types.

BupycHuk:


BupycHuk:
@AlekSi nothing will fail until we start to use these constants, it was done to be able to work on task simultaneously. Or do you mean updating PMM-2.0? So we use service type or agent type in API only for remove endpoints and I'll update PR to handle them.

AlekSi:
I guess my main concern is that we are moving from exhaustive checks for types (where we handle all cases in switches and maps contain all types) to non-exhaustive. And it is very easy to forget to add them later.

PR #375, opened 2020-04-24
PMM-5819 Added fields to control STT enable/disable

percona-csalguero:


models/settings_helpers.go
```diff
@@ -109,6 +113,9 @@ func UpdateSettings(q reform.DBTX, params *ChangeSettingsParams) (*Settings, err
 		settings.AlertManagerURL = ""
 	}
 
+	// because of validation, both variables cannot be true at the same time
+	settings.EnableSTT = params.EnableSTT && !params.DisableSTT
```
BupycHuk, line 117: settings via API can be passed independently, So, both values can be `false`. Please update like we did for telemetry

AlekSi, line 117: Let's do it in a more explicit and "dumb" way like for telemetry above. If something needs a comment explaining the logic, then it probably could be simpler. :)

models/settings_helpers.go
```diff
@@ -186,6 +193,13 @@ func ValidateSettings(params *ChangeSettingsParams) error {
 		}
 	}
 
+	if params.DisableSTT && params.EnableSTT {
+		return fmt.Errorf("enable STT and disable STT cannot be both true")
+	}
+	if params.EnableSTT && !params.EnableTelemetry {
+		return fmt.Errorf("cannot enable STT while telemetry is disabled")
```
AlekSi, line 200: You are confusing the current state and desired state there. We should not allow user to enable STT and disable cemetery in the same request (so `!params.EnableTelemetry` should be `params.DisableTelemetry`).

And then in other place, where we save, we should check that STT can't be enabled when telemetry is disabled.

models/settings_helpers.go
```diff
@@ -186,6 +193,13 @@ func ValidateSettings(params *ChangeSettingsParams) error {
 		}
 	}
 
+	if params.DisableSTT && params.EnableSTT {
+		return fmt.Errorf("enable STT and disable STT cannot be both true")
+	}
+	if params.EnableSTT && !params.EnableTelemetry {
```
BupycHuk, line 199: settings via API can be passed independently, So, params.EnableTelemetry may be false even if it was enabled in previous request. Please don't use `!` for validation

services/server/server.go
```diff
@@ -460,6 +461,12 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverp
 		return status.Error(codes.FailedPrecondition, "Data retention for queries is set via DATA_RETENTION environment variable.")
 	}
 
+	if req.DisableStt && req.EnableStt {
+		return status.Error(codes.InvalidArgument, "Enable STT and disable STT cannot be both true")
+	}
+	if req.EnableStt && !req.EnableTelemetry {
```
BupycHuk, line 467: let's check that `req.EnableStt` is not passed with `req.DisableTelemetry` and `req.DisableStt` is not passed with `req.EnableTelemetry`. Do not forget about `s.envSettings.DisableTelemetry`

models/settings.go
```diff
@@ -48,6 +48,9 @@ type Settings struct {
 
 	// not url.URL to keep username and password
 	AlertManagerURL string `json:"alert_manager_url"`
+
+	// Enable/Disable Security Threat Tool
+	EnableSTT bool `json:"enable_stt"`
```
BupycHuk, line 53: `STTEnabled`

models/settings_helpers.go
BupycHuk: We also need validate using params and saved settings.
if `settings.Telemetry.Disabled` is true `params.EnableSTT` can't be true.
if `settings.STTEnabled` is true `params.DisableTelemetry` can't be true.

AlekSi:
@percona-csalguero Please merge with the current PMM-2.0 branch

BupycHuk:
@percona-csalguero please resolve merge conflicts

BupycHuk:
@percona-csalguero Travis CI tests are failed

models/settings_helpers.go
```diff
@@ -189,6 +208,27 @@ func ValidateSettings(params *ChangeSettingsParams) error {
 	return nil
 }
 
+func ValidateSettingsConflicts(params *ChangeSettingsParams, settings *Settings) error {
```
AlekSi, line 211: It should not be exported

services/server/server_test.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	"github.com/percona/pmm/api/serverpb"
 	"github.com/stretchr/testify/assert"
+	mock "github.com/stretchr/testify/mock"
```
AlekSi, line 27: No need to add alias there.

models/settings_helpers.go
```diff
@@ -65,6 +66,11 @@ type ChangeSettingsParams struct {
 	// not url.URL to keep username and password
 	AlertManagerURL       string
 	RemoveAlertManagerURL bool
+
+	// Enable Secutiry Threat Tool
+	EnableSTT bool
+	// Disable Secutiry Threat Tool
+	DisableSTT bool
```
artemgavrilov, line 73: Why we need two fields for STT config? To change default behavior in the future?

AlekSi:
There is merge conflict

models/settings_helpers.go
AlekSi: In gRPC/JSON request, we can't differentiate between `{"stt": false}`, `{"stt": null}`, and `{}`. So we have to use something else if zero value is useful. The biggest problem is boolean flags, so we use command verbs in requests: enable_stt, disable_stt. Both can't be present, but both can be absent.

See also other resolved discussions in that PR and in API tests

artemgavrilov: Oh, it makes sense, thanks!

percona-csalguero: Because of go's default values. The same way we have them for telemetry, we need fields to explicitly enable/disable a function. 
Suppose we only use enable. If the request comes with the default values, it would be false and we would be disabling the feature.

AlekSi: Oh, and another reason: that API is somewhat special because there we expect users to send only what they want to change, not the whole state/model. Request `{}` is valid request for that API.

artemgavrilov: 💜

BupycHuk:
@artemgavrilov could you check why downloadCheck tests are failing?

artemgavrilov:
@BupycHuk here is PR fixes test, but we have another problem on SAAS side and it will be fixed only tomorrow. 😕

BupycHuk:
@artemgavrilov why did you close this PR?

artemgavrilov:
@BupycHuk oops, accidentally, I don't know how it happened. Probably I hit some shortcut. Sorry.

AlekSi:
SaaS side is fixed

services/server/server.go
```diff
@@ -440,6 +441,13 @@ func (s *Server) validateChangeSettingsRequest(ctx context.Context, req *serverp
 		}
 	}
 
+	if req.DisableStt && req.EnableStt {
+		return status.Error(codes.InvalidArgument, "Enable STT and disable STT cannot be both true")
+	}
+	if req.EnableStt && req.DisableTelemetry {
+		return status.Errorf(codes.InvalidArgument, "Cannot enable STT while disabling telemetry")
+	}
+
```
AlekSi, line 450: @percona-csalguero @BupycHuk why we validate it there and in models?

percona-csalguero: You are right. I am going to remove it from here.

models/settings_helpers_test.go
```diff
@@ -153,60 +153,70 @@ func TestSettings(t *testing.T) {
...
 			_, err = models.UpdateSettings(sqlDB, &models.ChangeSettingsParams{
-				EnableSTT: true,
+				EnableSTT:        true,
+				DisableTelemetry: true,
 			})
-			assert.EqualError(t, err, `cannot enable STT while telemetry is disabled`)
+			assert.EqualError(t, err, `Cannot enable STT while disabling telemetry.`)
 
-			_, err = models.UpdateSettings(sqlDB, &models.ChangeSettingsParams{
+			// enable both
```
BupycHuk, line 184: Maybe it's better to disable them before enabling? Because they are already enabled

AlekSi: Nope, STT is disabled at this point

PR #376, opened 2020-04-24
PMM-5776 Add external service to inventory.

askomorokhov:
https://jira.percona.com/browse/PMM-5776

refs:
 - https://github.com/percona/pmm/pull/517
 - https://github.com/percona/pmm-admin/pull/100


models/database.go
```diff
@@ -260,6 +260,16 @@ var databaseSchema = [][]string{
...
 			ALTER COLUMN rds_enhanced_metrics_disabled DROP DEFAULT`,
 	},
+
+	13: {
+		`ALTER TABLE services
+			DROP CONSTRAINT address_socket_check`,
+
+		`ALTER TABLE services
+			ADD CONSTRAINT address_socket_check CHECK (
+				(address IS NOT NULL AND socket IS NULL) OR (address IS NULL AND socket IS NOT NULL) OR (address IS NULL AND socket IS NULL)
```
BupycHuk, line 270: maybe for the third condition we should add service type, what do you think?

askomorokhov: Well, I think, it is not a very good solution to duplicate full validation in DB.
Such a concept can lead to:
-  the creation of validation stored procedures as constraints - and keeping them in sync it with models;
-  overcomplicate migration without the right instruments to apply/handle them.

And I guess soon the external services will also have host/socket connection to store.

Do you have any strong pros arguments about this?

AlekSi:
@askomorokhov Please merge with PMM-2.0

PR #378, opened 2020-04-29
SAAS-84 Update SAAS dependency

artemgavrilov:
That PR should fix the problem with crashed download checks test.


AlekSi:
Merging it to fix tests and unblock other work.

PR #379, opened 2020-04-30
PMM-5840 POC

percona-csalguero:
This is just my idea about how can we implement this.


internal/pprocess/pprocess.go
```diff
@@ -0,0 +1,10 @@
+// Package pprocess stands for Periodic Process
+package pprocess
+
+import "context"
+
+// PProcess defined the periodic process interface
+type PProcess interface {
```
AlekSi, line 7: As a rule of thumb, do not introduce a generic interface for a specific problem unless you have 3 similar problems that could use a generic interface.

internal/pprocess/cleaner/cleaner.go
```diff
@@ -0,0 +1,89 @@
...
+func (p *Cleaner) Stop() error {
+	if !p.isRunning() {
+		return fmt.Errorf("cleaner process is not running")
+	}
+
+	p.stopChan <- struct{}{}
+	return nil
+}
+
+func (p *Cleaner) run(ctx context.Context) {
```
BupycHuk, line 52: I think we can remove `Start`, `Stop` methods and use only this one.

BupycHuk: If we need it to be run in background we are able to call it in separate goroutine.
And if we need to stop it we can stop it by cancelling context.

internal/pprocess/cleaner/cleaner.go
```diff
@@ -0,0 +1,89 @@
...
+	defer p.setRunning(false)
+
+	// Semaphore to prevent having 2 queries running at the same time.
```
BupycHuk, line 55: It won't happen since you already check that only one goroutine can be running

internal/pprocess/cleaner/cleaner.go
```diff
@@ -0,0 +1,89 @@
...
+	if !p.isRunning() {
+		return fmt.Errorf("cleaner process is not running")
+	}
+
+	p.stopChan <- struct{}{}
+	return nil
+}
+
+func (p *Cleaner) run(ctx context.Context) {
+	defer p.setRunning(false)
```
BupycHuk, line 53: this can be done by semaphore

internal/pprocess/cleaner/cleaner.go
```diff
@@ -0,0 +1,89 @@
...
+		db:       db,
+		ticker:   ticker,
+		stopChan: make(chan struct{}),
+		lock:     &sync.Mutex{},
+	}
+}
+
+// Start starts the background process cleaner.
+func (p *Cleaner) Start(ctx context.Context) error {
+	if p.isRunning() {
```
BupycHuk, line 35: it makes possible to have race condition

internal/pprocess/cleaner/cleaner.go
BupycHuk: current implementation may cause race condition.

internal/pprocess/cleaner/cleaner.go
```diff
@@ -0,0 +1,89 @@
...
+	// It should never happen but if for some reason the clean query takes too much time
+	// and we receive a ticker tick while the query is still running, we should prevent that
+	sem := make(chan struct{}, 1)
+	sem <- struct{}{}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-p.stopChan:
```
BupycHuk, line 65: so this channel can be removed.

BupycHuk:
@percona-csalguero could you remove this PR if it's not actual anymore

PR #380, opened 2020-04-30
PMM-5840 Old action results cleaner

percona-csalguero:


utils/clean/clean.go
```diff
@@ -0,0 +1,64 @@
...
+	return &CleanResults{db: db}
+}
+
+// Run starts the clean process.
+func (c *CleanResults) Run(ctx context.Context, interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	l := logrus.WithField("component", "cleaner")
+
+	for {
+		// if this query runs every minute, it will clean results older than 1 minute,
```
BupycHuk, line 46: Maybe it's better to have 2 separate variables for them?

utils/clean/clean.go
```diff
@@ -0,0 +1,64 @@
...
+		// if it runs every 10 minutes, it will clean results older than 10 minutes.
+		olderThan := models.Now().Add(-1 * interval)
+		if err := models.CleanupOldResults(c.db.Querier, olderThan); err != nil {
+			l.Error(err)
+		}
+
+		select {
+		case <-ctx.Done():
+			ticker.Stop()
+			select {
```
BupycHuk, line 56: any reason to do this?

utils/clean/clean.go
```diff
@@ -0,0 +1,64 @@
...
+		// if this query runs every minute, it will clean results older than 1 minute,
+		// if it runs every 10 minutes, it will clean results older than 10 minutes.
+		olderThan := models.Now().Add(-1 * interval)
+		if err := models.CleanupOldResults(c.db.Querier, olderThan); err != nil {
+			l.Error(err)
+		}
+
+		select {
+		case <-ctx.Done():
+			ticker.Stop()
```
BupycHuk, line 55: just add it as defer right after https://github.com/percona/pmm-managed/pull/380/files#diff-331e36915d67ae729a2bdc6e16409abaR42

utils/clean/clean.go
percona-csalguero: To ensure that if context if being canceled at the same time the ticker tics, the channel is empty before returning.

BupycHuk: I don't think that it's required here.

utils/clean/clean.go
BupycHuk: It's not addressed nor answered yet

utils/clean/clean.go
BupycHuk: this one too

utils/clean/clean.go
percona-csalguero: If we add a second parameter to change the deletion range we would end with 2 scenarios:
**First scenario:** run every 2 minutes, delete results older than 4 minutes
```
00:00 x deleted on first run
00:01 x deleted on first run
00:02 x deleted on first run
00:03 x deleted on second run
00:04 x deleted on second run
00:05
00:06 First run <- ticker starts here
00:07
00:08 Second run
00:09
00:10
00:11
00:12
00:13
00:14
00:15
```
**Second scenario:** run every 4 minutes, delete results older than 2 minutes
```
00:00 x deleted on first run
00:01 x deleted on first run
00:02 x deleted on first run
00:03 x deleted on first run
00:04 x deleted on second run
00:05 x deleted on second run
00:06 x deleted on second run <-- First run - ticker starts here
00:07 x deleted on second run
00:08 
00:09
00:10 Second run
00:11
00:12
00:13
00:14
00:15
```

**Current implementation:** run every 2 minutes, delete results older than 2 minutes
```
00:00 x deleted on first run
00:01 x deleted on first run
00:02 x deleted on first run
00:03 x deleted on first run
00:04 x deleted on second run
00:05 x deleted on second run
00:06 <-- First run - ticker starts here
00:07 
00:08 <- Second run 
00:09
00:10 
00:11
00:12
00:13
00:14
00:15
```
The current implementation keeps results clean enough and it is easier to understand how it works. 
If we add another parameter, I would also add a validation to ensure we are in the first scenario.

BupycHuk: @AlekSi what do you think?

BupycHuk: I think we will run it every 1 hour and remove everything older than 24 hours.

AlekSi: 1 hour / 24 hours values are too high.

I think the sliding window approach (run every X minutes, delete rows older than Y minutes) is better overall for storage consumption.

I also think X and Y should be named constants until we really have a usecase for them being configurable.

main.go
```diff
@@ -626,5 +629,11 @@ func main() {
 		runDebugServer(ctx)
 	}()
 
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		cleaner.Run(ctx, 10*time.Minute, 30*time.Minute)
```
askomorokhov, line 635: I'd define these intervals as constants.

BupycHuk: +1

AlekSi: I proposed to do the same:

> I also think X and Y should be _named_ constants until we really have a usecase for them being configurable.

utils/clean/clean.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Run starts the clean process.
+func (c *CleanResults) Run(ctx context.Context, interval time.Duration, olderThan time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	l := logrus.WithField("component", "cleaner")
+
+	for {
+		olderThanTs := models.Now().Add(-1 * olderThan)
```
AlekSi, line 47: It can be just `-olderThan`.

percona-csalguero: models.Now() is time.Time and olderThan is time.Duration

AlekSi: It has nothing to do with types. `-1 * olderThan` can be replaced by `-olderThan`

PR #388, opened 2020-05-13
PMM-5924 Create working dir for alertmanager.

askomorokhov:
https://jira.percona.com/browse/PMM-5924

FB: Percona-Lab/pmm-submodules#843

### DIFF
--- a/services/alertmanager/alertmanager.go
+++ b/services/alertmanager/alertmanager.go
@@ -21,7 +21,10 @@ import (
 	"context"
 	"io/ioutil"
 	"os"
+	"os/user"
+	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	httptransport "github.com/go-openapi/runtime/client"
@@ -60,6 +63,7 @@ func (svc *Service) Run(ctx context.Context) {
 	svc.l.Info("Starting...")
 	defer svc.l.Info("Done.")
 
+	svc.createDataDir()
 	svc.generateBaseConfig()
 
 	t := time.NewTicker(resendInterval)
@@ -77,19 +81,49 @@ func (svc *Service) Run(ctx context.Context) {
 	}
 }
 
+// createDataDir creates alert manager directories if not exists in the persistent volume.
+func (svc *Service) createDataDir() {
+	group, err := user.Lookup("pmm")
+	if err != nil {
+		svc.l.Errorf("error looking up pmm user user info %v.", err)
+		return
+	}
+
+	uid, _ := strconv.Atoi(group.Uid)
+	gid, _ := strconv.Atoi(group.Gid)
+
+	stat, err := os.Stat(alertmanagerDataDir)
+	if os.IsNotExist(err) {
+		svc.l.Infof("Creating %q", alertmanagerDataDir)
+		if err := os.MkdirAll(alertmanagerDataDir, 0775); err != nil {
+			svc.l.Errorf("Cannot create datadir for AlertManager %v.", err)
+			return
+		}
+		if err := os.Chown(alertmanagerDataDir, uid, gid); err != nil {
+			svc.l.Errorf("Cannot chown datadir for AlertManager %v.", err)
+			return
+		}
+	}
+
+	if stat.Mode().Perm() != 0775 {
+		if err := os.Chmod(alertmanagerDataDir, 0775); err != nil {
+			svc.l.Errorf("Cannot chmod datadir for AlertManager %v.", err)
+		}
+	}
+
+	systat := stat.Sys().(*syscall.Stat_t)
+	if systat.Uid != uint32(uid) || systat.Gid != uint32(gid) {
+		if err := os.Chown(alertmanagerDataDir, uid, gid); err != nil {
+			svc.l.Errorf("Cannot chown datadir for AlertManager %v.", err)
+		}
+	}
+}
+
 // generateBaseConfig generates /srv/alertmanager/alertmanager.base.yml if it is not present.
 //
 // TODO That's a temporary measure until we start generating /etc/alertmanager.yml
 // using /srv/alertmanager/alertmanager.base.yml as a base. See supervisord config.
 func (svc *Service) generateBaseConfig() {
-	// create alert manager directories if not exists in the persistent volume.
-	if _, err := os.Stat(alertmanagerDataDir); os.IsNotExist(err) {
-		svc.l.Infof("Creating %q", alertmanagerDataDir)
-		if err := os.MkdirAll(alertmanagerDataDir, os.ModePerm); err != nil {
-			svc.l.Errorf("Cannot create datadir for AlertManager %v.", err)
-			os.Exit(1)
-		}
-	}
 
 	_, err := os.Stat(path)
 	svc.l.Debugf("%s status: %v", path, err)
--- a/services/alertmanager/alertmanager.go
+++ b/services/alertmanager/alertmanager.go
@@ -82,33 +82,16 @@ func (svc *Service) Run(ctx context.Context) {
 
 // createDataDir creates alert manager directories if not exists in the persistent volume.
 func (svc *Service) createDataDir() {
-	prometheusDirStat, err := os.Stat(prometheusDir)
-	if err != nil {
-		svc.l.Errorf("Cannot get stat of %q: %v.", prometheusDir, err)
+	// try to create alertmanager data dir if not exists.
+	if err := os.MkdirAll(alertmanagerDataDir, 0775); err != nil {
+		svc.l.Errorf("Cannot create datadir for AlertManager %v.", err)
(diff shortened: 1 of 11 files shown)


askomorokhov:
> Can we do it there?
> 
> https://github.com/percona/pmm-managed/blob/74af6b1238a0a0f3935cfa85af5c2ed9f14a1b0a/services/alertmanager/alertmanager.go#L76-L100

@AlekSi, yes, we can. Moved.

services/alertmanager/alertmanager.go
```diff
@@ -78,7 +82,15 @@ func (svc *Service) Run(ctx context.Context) {
 // TODO That's a temporary measure until we start generating /etc/alertmanager.yml
 // using /srv/alertmanager/alertmanager.base.yml as a base. See supervisord config.
 func (svc *Service) generateBaseConfig() {
-	const path = "/srv/alertmanager/alertmanager.base.yml"
+	// create alert manager directories if not exists in the persistent volume.
+	if _, err := os.Stat(alertmanagerDataDir); os.IsNotExist(err) {
+		svc.l.Infof("Creating %q", alertmanagerDataDir)
+		if err := os.MkdirAll(alertmanagerDataDir, os.ModePerm); err != nil {
```
AlekSi, line 88: That should be `0775` (rwxrwxr-x), not os.ModePerm. That's just not how it works – os.ModePerm is a bit mask, not a value.

We also should also set owner and group to the same values as `/srv/prometheus` has.

services/alertmanager/alertmanager.go
```diff
@@ -78,7 +82,15 @@ func (svc *Service) Run(ctx context.Context) {
 // TODO That's a temporary measure until we start generating /etc/alertmanager.yml
 // using /srv/alertmanager/alertmanager.base.yml as a base. See supervisord config.
```
AlekSi, line 83: Comment should be updated: /srv/alertmanager/alertmanager.base.yml generation is a temporary measure, /srv/alertmanager/data creation is not. Maybe that should be two different methods for clarity.

services/alertmanager/alertmanager.go
```diff
@@ -78,7 +82,15 @@ func (svc *Service) Run(ctx context.Context) {
...
 func (svc *Service) generateBaseConfig() {
-	const path = "/srv/alertmanager/alertmanager.base.yml"
+	// create alert manager directories if not exists in the persistent volume.
+	if _, err := os.Stat(alertmanagerDataDir); os.IsNotExist(err) {
+		svc.l.Infof("Creating %q", alertmanagerDataDir)
+		if err := os.MkdirAll(alertmanagerDataDir, os.ModePerm); err != nil {
+			svc.l.Errorf("Cannot create datadir for AlertManager %v.", err)
+			os.Exit(1)
```
AlekSi, line 90: Absolutely not. It is too late for that there. The only thing we could do is to log a error and return.

services/alertmanager/alertmanager.go
```diff
@@ -78,7 +82,15 @@ func (svc *Service) Run(ctx context.Context) {
 // TODO That's a temporary measure until we start generating /etc/alertmanager.yml
 // using /srv/alertmanager/alertmanager.base.yml as a base. See supervisord config.
 func (svc *Service) generateBaseConfig() {
-	const path = "/srv/alertmanager/alertmanager.base.yml"
+	// create alert manager directories if not exists in the persistent volume.
+	if _, err := os.Stat(alertmanagerDataDir); os.IsNotExist(err) {
```
AlekSi, line 86: It may exist but have different permissions – we should handle that.

services/alertmanager/alertmanager.go
askomorokhov: Moved to a separate method.

services/alertmanager/alertmanager.go
askomorokhov: it's strange, but with os.ModePerm it worked better :)

AlekSi: os.ModePerm == 0777 == rwxrwxrwx. We should not allow everyone to write to that directory

services/alertmanager/alertmanager.go
```diff
@@ -73,12 +81,50 @@ func (svc *Service) Run(ctx context.Context) {
 	}
 }
 
+// createDataDir creates alert manager directories if not exists in the persistent volume.
+func (svc *Service) createDataDir() {
+	group, err := user.Lookup("pmm")
```
AlekSi, line 86: Nope. We should stat `/srv/prometheus` and use the same uid and gid. Remember, that file is on the volume.

services/alertmanager/alertmanager.go
```diff
@@ -73,12 +81,50 @@ func (svc *Service) Run(ctx context.Context) {
...
+	if err != nil {
+		svc.l.Errorf("error looking up pmm user user info %v.", err)
+		return
+	}
+
+	uid, _ := strconv.Atoi(group.Uid)
+	gid, _ := strconv.Atoi(group.Gid)
+
+	stat, err := os.Stat(alertmanagerDataDir)
+	if os.IsNotExist(err) {
```
AlekSi, line 96: I think we should do something like that
```
if err == nil {
  return
}

// do everything: mkdir, chown, chmod, etc.

File: services/alertmanager/alertmanager.go (hunk 5 above)
Comment (askomorokhov) [reply services/alertmanager/alertmanager.go]: Added.

File: services/alertmanager/alertmanager.go (hunk 6 above)
Comment (askomorokhov) [reply services/alertmanager/alertmanager.go:100]: stat, err := os.Stat(alertmanagerDataDir)
if err == nil {
  return
}
In such case we cannot handle https://github.com/percona/pmm-managed/pull/388#discussion_r425170314

Comment (BupycHuk) [reply services/alertmanager/alertmanager.go:100]: I think @AlekSi means that if err != nil it's allways not exist error based on condition above.

File: services/alertmanager/alertmanager.go [hunk 7]
```diff
@@ -73,12 +80,60 @@ func (svc *Service) Run(ctx context.Context) {
...
+	// Check and fix permissions.
+	if alertmanagerDataDirStat.Mode()&os.ModePerm != os.FileMode(0775) {
+		if err := os.Chmod(alertmanagerDataDir, 0775); err != nil {
+			svc.l.Errorf("Cannot chmod datadir for AlertManager %v.", err)
+		}
+	}
+
+	alertmanagerDataDirSysStat := alertmanagerDataDirStat.Sys().(*syscall.Stat_t)
+	aUID, aGID := int(alertmanagerDataDirSysStat.Uid), int(alertmanagerDataDirSysStat.Gid)
+	// Chown user and grop as Prometheus has if they are not same.
```
BupycHuk, line 123: group

services/alertmanager/alertmanager.go
BupycHuk: Do we really need this `if`? MkdirAll does nothing if directory is exists and doesn't return error.
So we can move check from https://github.com/percona/pmm-managed/pull/388/files#diff-a8171d88a4e24ab2acd4b471b52230faR124 to https://github.com/percona/pmm-managed/pull/388/files#diff-a8171d88a4e24ab2acd4b471b52230faR107

services/alertmanager/alertmanager.go
```diff
@@ -73,12 +80,52 @@ func (svc *Service) Run(ctx context.Context) {
 	}
 }
 
+// createDataDir creates alert manager directories if not exists in the persistent volume.
```
AlekSi, line 83: "Alertmanager", single word

services/alertmanager/alertmanager.go
```diff
@@ -73,12 +80,52 @@ func (svc *Service) Run(ctx context.Context) {
...
+func (svc *Service) createDataDir() {
+	// try to create alertmanager data dir if not exists.
+	if err := os.MkdirAll(alertmanagerDataDir, 0775); err != nil {
+		svc.l.Errorf("Cannot create datadir for AlertManager %v.", err)
```
AlekSi, line 87: "Alertmanager", not "AlertManager". In other places too.

services/alertmanager/alertmanager.go
```diff
@@ -73,12 +80,52 @@ func (svc *Service) Run(ctx context.Context) {
...
+	}
+}
+
 // generateBaseConfig generates /srv/alertmanager/alertmanager.base.yml if it is not present.
 //
 // TODO That's a temporary measure until we start generating /etc/alertmanager.yml
 // using /srv/alertmanager/alertmanager.base.yml as a base. See supervisord config.
 func (svc *Service) generateBaseConfig() {
-	const path = "/srv/alertmanager/alertmanager.base.yml"
+
```
AlekSi, line 128: Empty line

PR #389, opened 2020-05-13
SAAS-110 Refactoring and tests

artemgavrilov:


services/alert_params.go
```diff
@@ -0,0 +1,24 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package services
+
+// AlertParams defines alertmanager alert parameters.
+type AlertParams struct {
```
AlekSi, line 20: So this is kinda model, but not stored in PostgreSQL. I'm not sure if we want to move it into models or keep there.
Or we can take a third-way and remove the whole struct, passing 3 parameters to methods.
@BupycHuk thoughts?

BupycHuk: I don't think that it should be moved somewhere else. We have similar structure like server.Params and scrapeConfigParams, so I think we can have extra one structure for params

AlekSi: > We have similar structure like server.Params and scrapeConfigParams

Both are used only by a single service. That's unlike that `AlertParams`.

AlekSi: > Or we can take a third-way and remove the whole struct, passing 3 parameters to methods.

@artemgavrilov let's do that ^. We can use unexported alertmanager.alertParams if needed, but should not be exported.

artemgavrilov: Done

services/checks/checks_test.go
```diff
@@ -47,6 +51,63 @@ func TestDownloadChecks(t *testing.T) {
...
+	assert.Equal(t, "good_check_mongo", c3.Name)
+	assert.Equal(t, uint32(1), c3.Version)
+	assert.Empty(t, c3.Query)
+}
+
+func TestCollectChecks(t *testing.T) {
+	t.Run("collect local checks", func(t *testing.T) {
+		err := os.Setenv("PERCONA_TEST_CHECKS_FILE", "../../testdata/checks/checks.yml")
+		require.NoError(t, err)
+		defer os.Unsetenv("PERCONA_TEST_CHECKS_FILE")
```
BupycHuk, line 83: **[golangci-lint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
Error return value of `os.Unsetenv` is not checked (errcheck)

services/checks/checks.go
```diff
@@ -408,8 +430,7 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
 	alertsIDs := make([]string, len(results))
 	for i, result := range results {
 		id := alertsPrefix + hashID(target.serviceID+"/"+result.Summary)
-		alert := makeAlert(id, check.Name, target, &result)
-		s.alertsRegistry.Add(id, 0, alert)
+		s.createAlert(id, check.Name, target, &result)
```
BupycHuk, line 433: **[golangci-lint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
G601: Implicit memory aliasing in for loop. (gosec)

services/checks/checks_test.go
artemgavrilov: Seems that reviewdog uses Nurlans token

artemgavrilov: @BupycHuk ^

services/checks/checks.go
```diff
@@ -165,13 +166,31 @@ func (s *Service) Run(ctx context.Context) {
 	}
 }
 
-// getChecks returns available checks.
-func (s *Service) getChecks() []check.Check {
+// getChecks returns available MySQL checks.
```
AlekSi, line 169:
```suggestion
// getMySQLChecks returns available MySQL checks.
```

The same thing for methods below.
I wonder why golangci-lint is silent about that.

services/checks/checks.go
```diff
@@ -406,10 +428,9 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
 	l.Debugf("Results: %+v", results)
 
 	alertsIDs := make([]string, len(results))
-	for i, result := range results {
+	for i, result := range results { //nolint:gosec
```
AlekSi, line 431: I wonder if we know what is "G601: Implicit memory aliasing in for loop" and if we should disable that particular rule globally.

artemgavrilov: That's because I pass pointer `results` into `createAlert` function, but that function is synchronous and it makes a copy of required `results` fields.

PR #392, opened 2020-05-15
SAAS-67 API for manual checks execution

artemgavrilov:
- https://jira.percona.com/browse/SAAS-67

Requires:

- [x] https://github.com/percona/pmm/pull/529

FB: https://github.com/Percona-Lab/pmm-submodules/pull/857

` curl -k -X POST https://admin:admin@127.0.0.1:443/v1/management/SecurityChecks/Start`


services/errors.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package services
+
+import "errors"
+
+// STTDisabledError means that STT checks are disabled and can't be executed.
+var STTDisabledError = errors.New("STT disabled")
```
artemgavrilov, line 22: Well, I don't like it, but don't see a good alternative.

services/server/server.go
```diff
@@ -331,6 +335,21 @@ func (s *Server) UpdateStatus(ctx context.Context, req *serverpb.UpdateStatusReq
 	}, nil
 }
 
+// StartSecurityChecks starts STT checks execution.
+func (s *Server) StartSecurityChecks(ctx context.Context, request *serverpb.StartSecurityChecksRequest) (*serverpb.StartSecurityChecksResponse, error) {
+
```
BupycHuk, line 340: we usually don't leave blank lines as first line in method.

artemgavrilov: Fixed

services/checks/checks.go
```diff
@@ -165,6 +156,29 @@ func (s *Service) Run(ctx context.Context) {
 	}
 }
 
+// StartChecks triggers STT checks downloading and execution. It returns services.STTDisabledError if STT is disabled.
+func (s *Service) StartChecks(ctx context.Context) error {
```
BupycHuk, line 160: Looks like we can write test for it.

services/checks/checks.go
```diff
@@ -165,6 +156,29 @@ func (s *Service) Run(ctx context.Context) {
...
+	var sttEnabled bool
+	settings, err := models.GetSettings(s.db)
+	if err != nil {
+		return err
+	}
+	if settings != nil && settings.SaaS.STTEnabled {
+		sttEnabled = true
+	}
+
+	if !sttEnabled {
```
BupycHuk, line 170: I'm pretty sure that settings can't be nil.
```suggestion
	if !settings.SaaS.STTEnabled {
```

services/checks/checks.go
```diff
@@ -165,6 +156,29 @@ func (s *Service) Run(ctx context.Context) {
 	}
 }
 
+// StartChecks triggers STT checks downloading and execution. It returns services.STTDisabledError if STT is disabled.
+func (s *Service) StartChecks(ctx context.Context) error {
+	var sttEnabled bool
```
BupycHuk, line 161: We don't need this variable

services/checks/checks.go
artemgavrilov: Added few tests

services/checks/checks.go

services/checks/checks.go

BupycHuk:
please resolve conflicts

services/checks/checks_test.go
```diff
@@ -94,3 +101,41 @@ uEF33ScMPYpvHvBKv8+yBkJ9k4+DCfV4nDs6kKYwGhalvkkqwWkyfJffO+KW7a1m3y42WHpOnzBxLJ+I
...
+
+	t.Run("stt enabled", func(t *testing.T) {
+		sqlDB := testdb.Open(t, models.SkipFixtures, nil)
+		db := reform.NewDB(sqlDB, postgresql.Dialect, nil)
+
+		defer func() {
+			require.NoError(t, sqlDB.Close())
+		}()
+
+		ar := &mockAlertRegistry{}
```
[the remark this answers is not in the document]
artemgavrilov: Used third variant `var ar mockAlertRegistry`, what do you think?

services/errors.go
```diff
@@ -0,0 +1,22 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package services
+
+import "errors"
```
BupycHuk, line 19: **[golangci-lint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
`errors` is in the blacklist (depguard)

services/errors.go
BupycHuk, line 22: **[golangci-lint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
error var STTDisabledError should have name of the form ErrFoo (golint)

services/checks/checks.go
```diff
@@ -138,22 +139,12 @@ func (s *Service) Run(ctx context.Context) {
...
-
-		if sttEnabled {
-			nCtx, cancel := context.WithTimeout(ctx, checksTimeout)
-			s.collectChecks(nCtx)
-			s.executeChecks(nCtx)
-			cancel()
-		} else {
-			s.l.Info("STT is not enabled, doing nothing.")
+		if err := s.StartChecks(ctx); err != nil {
+			if err == services.STTDisabledError {
```
BupycHuk, line 143: **[golangci-lint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
err113: do not compare errors directly, use errors.Is() instead: "err == services.STTDisabledError" (goerr113)

services/server/server.go
```diff
@@ -331,6 +335,20 @@ func (s *Server) UpdateStatus(ctx context.Context, req *serverpb.UpdateStatusReq
 	}, nil
 }
 
+// StartSecurityChecks starts STT checks execution.
+func (s *Server) StartSecurityChecks(ctx context.Context, request *serverpb.StartSecurityChecksRequest) (*serverpb.StartSecurityChecksResponse, error) {
+	err := s.checksService.StartChecks(ctx)
+	if err != nil {
+		if err == services.STTDisabledError {
```
BupycHuk, line 342: **[golangci-lint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub>
err113: do not compare errors directly, use errors.Is() instead: "err == services.STTDisabledError" (goerr113)

services/checks/checks.go
```diff
@@ -138,22 +139,12 @@ func (s *Service) Run(ctx context.Context) {
...
-
-		if sttEnabled {
-			nCtx, cancel := context.WithTimeout(ctx, checksTimeout)
-			s.collectChecks(nCtx)
-			s.executeChecks(nCtx)
-			cancel()
-		} else {
-			s.l.Info("STT is not enabled, doing nothing.")
+		if err := s.StartChecks(ctx); err != nil {
+			if errors.Is(err, services.ErrSTTDisabled) {
```
AlekSi, line 143: Sorry, we don't use errors.Is/As in PMM yet: there are differences between it and `errors.Cause`. Use the later function or just compare with `==`. And let's blacklist a linter/rule that suggests otherwise.

Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "SAAS-67-stt-api"
```
AlekSi, line 19: Please don't forget to update that

services/checks/checks.go
```diff
@@ -139,22 +140,12 @@ func (s *Service) Run(ctx context.Context) {
...
-		}
-
-		if sttEnabled {
-			nCtx, cancel := context.WithTimeout(ctx, checksTimeout)
-			s.collectChecks(nCtx)
-			s.executeChecks(nCtx)
-			cancel()
-		} else {
-			s.l.Info("STT is not enabled, doing nothing.")
+		if err := s.StartChecks(ctx); err != nil {
```
AlekSi, line 143: Something like that to drop nested `if` and `else`:
```go
err := s.StartChecks(ctx)
switch err {
case nil:
  // nothing, continue
case services.ErrSTTDisabled:
  s.l.Info("STT is not enabled, doing nothing.")
default:
  s.l.Error(err)
}
```

services/checks/checks_test.go
```diff
@@ -156,6 +163,44 @@ uEF33ScMPYpvHvBKv8+yBkJ9k4+DCfV4nDs6kKYwGhalvkkqwWkyfJffO+KW7a1m3y42WHpOnzBxLJ+I
...
+
+	t.Run("stt enabled", func(t *testing.T) {
+		sqlDB := testdb.Open(t, models.SkipFixtures, nil)
+		db := reform.NewDB(sqlDB, postgresql.Dialect, nil)
+
+		defer func() {
+			require.NoError(t, sqlDB.Close())
+		}()
+
+		var ar mockAlertRegistry
```
AlekSi, line 188: Why we mock it?

services/errors.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package services
+
+import "github.com/pkg/errors"
+
+// ErrSTTDisabled means that STT checks are disabled and can't be executed.
+var ErrSTTDisabled = errors.New("STT disabled")
```
AlekSi, line 22: Do we ignore the value (`"STT disabled"`) everywhere? It seems that we use custom log and error messages. I think we could actually use a value in one of those places.

services/management/checks.go
```diff
@@ -0,0 +1,51 @@
...
+
+// NewChecksAPIService creates new Checks API Service.
+func NewChecksAPIService(checksService checksService) *ChecksAPIService {
+	return &ChecksAPIService{checksService: checksService}
+}
+
+// StartSecurityChecks starts STT checks execution.
+func (s *ChecksAPIService) StartSecurityChecks(ctx context.Context, request *managementpb.StartSecurityChecksRequest) (*managementpb.StartSecurityChecksResponse, error) {
+	err := s.checksService.StartChecks(ctx)
+	if err != nil {
```
AlekSi, line 42: `switch err` to avoid nested `if`? (but at least there is no `else`, so might be ok)

services/management/deps.go
```diff
@@ -41,3 +41,7 @@ type agentsRegistry interface {
 type prometheusService interface {
 	RequestConfigurationUpdate()
 }
+
+type checksService interface {
```
AlekSi, line 45: This interface is subtly different from others – we don't generate a mock for it. Should we?

/cc @BupycHuk @askomorokhov I'm having a déjà vu there – did we discuss it just recently?

services/checks/checks_test.go
```diff
@@ -156,6 +163,44 @@ uEF33ScMPYpvHvBKv8+yBkJ9k4+DCfV4nDs6kKYwGhalvkkqwWkyfJffO+KW7a1m3y42WHpOnzBxLJ+I
...
 }
 
+func TestStartChecks(t *testing.T) {
+	t.Run("stt disabled", func(t *testing.T) {
+		sqlDB := testdb.Open(t, models.SkipFixtures, nil)
+		db := reform.NewDB(sqlDB, postgresql.Dialect, nil)
+
+		defer func() {
+			require.NoError(t, sqlDB.Close())
+		}()
```
askomorokhov, line 173: I guess function wrapper can be omitted?

services/checks/checks.go
```diff
@@ -166,6 +157,25 @@ func (s *Service) Run(ctx context.Context) {
...
+	settings, err := models.GetSettings(s.db)
+	if err != nil {
+		return err
+	}
+
+	if !settings.SaaS.STTEnabled {
+		return services.ErrSTTDisabled
+	}
+
+	nCtx, cancel := context.WithTimeout(ctx, checksTimeout)
```
askomorokhov, line 171: According to the ticket topic "API for manual checks execution" - an hour timeout looks too long to wait on error.
Shouldn't we change it here or make configurable?

services/management/deps.go
askomorokhov: Carlos has reminded me it was discussed here https://github.com/percona/pmm-managed/pull/360

services/checks/checks_test.go
artemgavrilov: No, it can't because `A deferred function's arguments are evaluated when the defer statement is evaluated.` https://blog.golang.org/defer-panic-and-recover

Here's an example: https://play.golang.org/p/5uZnIPOt7VJ

services/checks/checks.go
artemgavrilov: One hour is the default timeout for checks execution, however, you can reduce that time by passing `ctx` with another timeout setting(or with cancel/deadline). So the user of that function has an option how to change awaiting time.

services/management/deps.go
artemgavrilov: So, I think there is no need to generate mocks until they used somewhere, but I can add them if it's common rule for the project. 
That interface named `checksService` because it hides `checks.Service`, so it's a direct inheritance of naming. However, maybe we should rename the `checks` package to `stt` or something else?

services/checks/checks_test.go
artemgavrilov: To not import one service package into another.

services/management/checks.go
```diff
@@ -0,0 +1,51 @@
...
+	checksService checksService
+}
+
+// NewChecksAPIService creates new Checks API Service.
+func NewChecksAPIService(checksService checksService) *ChecksAPIService {
+	return &ChecksAPIService{checksService: checksService}
+}
+
+// StartSecurityChecks starts STT checks execution.
+func (s *ChecksAPIService) StartSecurityChecks(ctx context.Context, request *managementpb.StartSecurityChecksRequest) (*managementpb.StartSecurityChecksResponse, error) {
```
BupycHuk, line 40: What do you think if we write a few tests for this method? at least that we handle errors correctly

services/management/deps.go
BupycHuk: I think we shouldn't generate until we require it. Another question: Why don't we use mocks, because we don't have tests :)
About renaming, I agree that it should be renamed to something like `stt`

services/management/checks.go
artemgavrilov: Here tests for that method: https://github.com/Percona-Lab/pmm-api-tests/pull/99/files
I don't think there is much sense to duplicate them here.

services/checks/checks.go

services/errors.go
artemgavrilov: Done

services/management/checks.go
artemgavrilov: For me, it looks ok. Don't think that `switch` will make it simpler.

services/management/checks.go
BupycHuk: Looks good, bu what about a case when we are unable to download checks or some internal error?

artemgavrilov: Added some tests

services/management/deps.go
AlekSi: "STT" is a product / feature name. We have other features in our roadmap with different names that will use this code.

services/management/checks_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestStartSecurityChecks(t *testing.T) {
+	t.Run("internal error", func(t *testing.T) {
+		var checksService mockChecksService
+		checksService.On("StartChecks", mock.Anything).Return(errors.New("random error"))
+
+		s := NewChecksAPIService(&checksService)
+
+		resp, err := s.StartSecurityChecks(context.Background(), &managementpb.StartSecurityChecksRequest{})
+		assert.EqualError(t, err, "rpc error: code = Internal desc = Failed to start security checks.")
```
AlekSi, line 39: Please use tests.AssertGRPCError

PR #393, opened 2020-05-18
PMM-5723 Reviewdog checks.

BupycHuk:


AlekSi:
What did not work inside devcontainer?

BupycHuk:
some ENV variables wasn't passed there and instead of finding all of them and passing them I decided to make check outside of devcontainer

.devcontainer/setup.py
```diff
@@ -73,6 +73,7 @@ def install_tools():
     run_commands([
         "curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh",
         "curl https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /root/go/bin",
+        "curl -sfL https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh | sh -s -- -b /root/go/bin",
```
AlekSi, line 76: Why `-sfL`? Is it worth to break consistency there?

.travis.yml
```diff
@@ -1,5 +1,5 @@
 dist: bionic
-language: minimal
+language: go
```
AlekSi, line 2: I guess that can be reverted too – minimal image should start faster.

.travis.yml
```diff
@@ -59,9 +66,12 @@ before_install:
   - docker exec pmm-managed-server ls /root/.cache/go-build
 
 install:
+  # install devcontainer
   - make env-devcontainer
 
 before_script:
+  # static analyze
+  - make env TARGET=ci-reviewdog
```
AlekSi, line 74: Not sure if we want to run linters before tests. Why not move it into `travis` target?

Makefile.devcontainer
```diff
@@ -88,3 +88,6 @@ travis: test-race test-cover test-crosscover
 		PMM_TEST_RUN_UPDATE=1 \
 		make test-race-crosscover
 	make check
+
+ci-reviewdog:
+	golangci-lint run -c=.golangci.yml --out-format=line-number | reviewdog -f=golangci-lint -level=error -reporter=github-pr-check
```
AlekSi, line 93: I think `error` will produce that many false positives that we will learn to ignore output completely. The trick we use in SaaS with two files with different sets of linters and levels is useful.

artemgavrilov: Also there is the second type of reporter `github-pr-review`, it adds comments to code.

on SAAS we use `github-pr-check` for critical linters to broke CI pipeline and `github-pr-review` for non-critical.

.travis.yml
BupycHuk: It's better to get linters feedback earlier to fix ASAP

Makefile.devcontainer
BupycHuk: fixed

.devcontainer/setup.py
AlekSi: @BupycHuk ^

.travis.yml
```diff
@@ -20,6 +20,9 @@ before_cache:
   - rm -fr /home/travis/go-build
   - docker cp pmm-managed-server:/root/.cache/go-build /home/travis/go-build
 
+go:
+  - 1.13.x
```
AlekSi: Please remove that one too.

.golangci-required.yml
```diff
@@ -0,0 +1,34 @@
...
+    - depguard
+    - goimports
+    - ineffassign
+    - govet
+    - staticcheck
+
+run:
+  skip-dirs:
+    # skip files copied from Prometheus
+    - services/prometheus/internal
```
AlekSi, line 31: We don't have that directory anymore.

.travis.yml
AlekSi: Well, yeah, sure. Just run them locally instead of using CI for that. We may add a separate target for running `golangci-lint --new-from-rev=PMM-2.0` if that would help. But I don't think we should encourage "push and wait" development.

.travis.yml
```diff
@@ -50,6 +55,8 @@ env:
 matrix:
   fast_finish: true
 
+go_import_path: github.com/percona/pmm-managed
```
AlekSi, line 55: Please remove that one too.

.travis.yml
BupycHuk: idk why but it doesn't work in devcontainer, it returns all errors, not only diff.

AlekSi:
```
[root@pmm-managed-server pmm-managed]# golangci-lint run | wc
   1571    7334   84166
[root@pmm-managed-server pmm-managed]# golangci-lint run --new-from-rev=PMM-2.0 | wc
     12      68     627
```

BupycHuk: Any reason to put lint after tests? it's often case when developers push changes before checking in linter and I think developer should see issues earlier then later.

AlekSi: > it's often case when developers push changes before checking in linter

Please don't do that. Please do not encourage others to do that. Please do run tests and linters locally.

I'm really worried about the decline of our engineering practices. Somehow we started to think that pushing code to CI and waiting is okay. That pushing code without tests is okay. That tests somehow are not in the estimates. Why?

BupycHuk: I don't think that it's okay, but we have this problem and it's better to catch the problem earlier. And it's clear that it doesn't work this way just telling people it again and again.

About test related to Alerting rules, I didn't know how to test it as part of API tests and I'm still not sure that it can be tested by checking AlertManager API. I'll check Prometheus API to check if it's updated or not.

artemgavrilov: >  it returns all errors, not only diff.

Mmm, I don't know, but how about fixing all linter warnings? There a lot of redundant `nolint` annotations and different small warnings that easy to fix. We can do it not at once, but each time that we create new PR, to include also a few old warnings fixes.

BupycHuk: @artemgavrilov it's already fixed.

askomorokhov: > I'm really worried about the decline of our engineering practices. Somehow we started to think that pushing code to CI and waiting is okay. That pushing code without tests is okay. That tests somehow are not in the estimates. Why?

It sounds like two different problem al-a "client-side vs server-side validation". 
And could be solved as a separate task with for example https://pre-commit.com.

AlekSi: > About test related to Alerting rules, I didn't know how to test it as part of API tests and I'm still not sure that it can be tested by checking AlertManager API.

https://github.com/Percona-Lab/pmm-api-tests/pull/94 – that PR, reviewed by you, uses Alertmanager API for almost the same problem. Generate alerting rule, save it, wait for it to appear on Alertmanager.

BupycHuk: You mentioned two or more times that these are not alermanager rules. So it should be checked as prometheus alerting rules.

AlekSi: Please check Slack

PR #397, opened 2020-05-26
SAAS-134 Skip PMM Server's PostgreSQL in checks

artemgavrilov:
- https://jira.percona.com/browse/SAAS-134

FB: https://github.com/Percona-Lab/pmm-submodules/pull/868


models/database.go
```diff
@@ -32,6 +32,9 @@ import (
 	"gopkg.in/reform.v1/dialects/postgresql"
 )
 
+// PMMPostgreSQLServiceName is a special Service Name representing PMM Server's PostgreSQL Service.
+const PMMPostgreSQLServiceName = "pmm-server-postgresql"
```
AlekSi, line 36: I would call it `PMMServer…` to match `PMMServerAgentID` and `PMMServerNodeID`

services/checks/checks.go
```diff
@@ -521,6 +527,17 @@ func (s *Service) findTargets(serviceType models.ServiceType) ([]target, error)
 	return targets, nil
 }
 
+// isSystem returns true if provided service is system
+func isSystem(service *models.Service) bool {
```
AlekSi, line 531: We skip them because we control everything inside PMM Server container/VM, and don't want users to try to update them.
That reason might affect the name of the function ("isSystem" seems to be too generic? and we don't call them "system services" anywhere) and implementation (we really care about "service.NodeID == models.PMMServerNodeID" only)

@BupycHuk thoughts?

models/database.go
artemgavrilov: Renamed to `PMMServerPostgreSQLServiceName`, it's long but precise.

services/checks/checks.go
AlekSi: @BupycHuk ^

BupycHuk: Why don't we just check if current service has name equal to `PMMPostgreSQLServiceName`? In that case we don't even need new method for that

artemgavrilov: If such comparison is enough then ok. I tried to achieve a maximum match to eliminate bugs.

artemgavrilov: @AlekSi @BupycHuk 
So we have three options:
1)Current function, that compares NodeID, ServiceType and ServiceName
2)Compare only NodeID
3)Compare only ServiceName

Service name seems to me like a weak condition. If we haven't any plans to run STT checks against any PMM server component(probably not), then comparison by NodeID is enough IMHO. Do you agree?

artemgavrilov: Ok, let's try to check only NodeID

artemgavrilov:
@AlekSi ah, It almost ready. Ok, I will make separate PR.

AlekSi:
Sorry :)

artemgavrilov:
PR with the test: https://github.com/percona/pmm-managed/pull/401

PR #398, opened 2020-05-27
Fix some linter warnings

artemgavrilov:


services/agents/channel/channel_test.go
```diff
@@ -104,7 +104,7 @@ func TestAgentRequest(t *testing.T) {
 	require.True(t, count > agentRequestsCap)
 
 	var channel *Channel
-	connect := func(ch *Channel) error { //nolint:unparam
+	connect := func(ch *Channel) error {
```
AlekSi, line 107: Oh, that false positive was very annoying, that's good it is fixed now

PR #399, opened 2020-05-29
PMM-5968 Avoid removing PMM-Server.

BupycHuk:
https://github.com/Percona-Lab/pmm-submodules/pull/874


models/agent_helpers.go
```diff
@@ -447,6 +447,10 @@ func RemoveAgent(q *reform.Querier, id string, mode RemoveMode) (*Agent, error)
 		return nil, err
 	}
 
+	if id == PMMServerAgentID {
+		return nil, status.Error(codes.PermissionDenied, "Removing pmm-agent on PMM Server prevents remote monitoring.")
```
askomorokhov, line 451: But does not this agent monitor **local** PostgreSQL by default as well?
If yes, maybe we need a more general error message here.

models/node_helpers.go
```diff
@@ -247,6 +247,10 @@ func RemoveNode(q *reform.Querier, id string, mode RemoveMode) error {
 		return err
 	}
 
+	if id == PMMServerNodeID {
+		return status.Error(codes.PermissionDenied, "Removing pmm-server node prevents remote monitoring.")
```
askomorokhov, line 251: same question as above.

models/agent_helpers.go
BupycHuk: Yes, it is, but it's not critical.

BupycHuk:
@AlekSi ok, but what is wrong with this PR? Could you point what should be fixed?

AlekSi:
I could, but I would prefer to address the root cause and discuss solutions, not "just fix" that PR.
The fix should not take too much time and does not block QA, etc.

BupycHuk:
@AlekSi https://github.com/Percona-Lab/pmm-submodules/pull/874

AlekSi:
/s/have/update. API tests fail there because it wasn't updated properly

PR #400, opened 2020-05-29
SAAS-126 Add mongoDB getCmdLineOpts action support

artemgavrilov:
- https://jira.percona.com/browse/SAAS-118

- [x] https://github.com/percona/pmm/pull/535
- [x] https://github.com/percona-platform/platform/pull/66

FB: https://github.com/Percona-Lab/pmm-submodules/pull/877


Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "SAAS-126-mongodb-action"
```
AlekSi, line 19: ^

services/checks/checks.go
```diff
@@ -413,7 +418,7 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
 	r, err := s.waitForResult(nCtx, resID)
 	cancel()
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to get check result")
+		return nil, errors.Wrap(err, "failed to get action result")
```
AlekSi, line 421: 👍

services/checks/checks.go
```diff
@@ -531,7 +554,7 @@ func (s *Service) findTargets(serviceType models.ServiceType) ([]target, error)
 			return nil
 		})
 		if e != nil {
-			s.l.Errorf("Failed to find agents for service %s, reason: %s", service.ServiceID, err)
+			s.l.Errorf("Failed to find agents for service %s, reason: %s", service.ServiceID, e)
```
artemgavrilov, line 557: That was bug

services/checks/checks.go
```diff
@@ -493,16 +505,32 @@ func (s *Service) findTargets(serviceType models.ServiceType) ([]target, error)
...
 
-			dsn, err := models.FindDSNByServiceIDandPMMAgentID(s.db.Querier, service.ServiceID, a[0].AgentID, "")
+			var agent *models.Agent
+			for _, a := range agents {
+				agentVersion, err := version.Parse(*a.Version)
+				if err != nil {
+					return errors.Wrap(err, "failed to checks agent version")
+				}
+
+				if !s.pmmVersion.Less(agentVersion) {
```
AlekSi, line 523: We should check for features (Action types) available in specific pmm-agent versions: MongoDBGetCmdLineOpts for 2.7.0+, other for 2.6.0+. We should not unconditionally skip all agents if they are older than PMM Server.

BupycHuk, line 523: I think we should have own version for each action type and looks like it's better to make this check here https://github.com/percona/pmm-managed/pull/400/files#diff-e0ff53b9498062b1544f8a3a40c5eb6bR384.

BupycHuk: `executeMongoDBChecks`

artemgavrilov: So, I left version check here because we iterate each agent for each service to find proper version

services/checks/checks.go
```diff
@@ -274,16 +301,23 @@ func (s *Service) executeChecks(ctx context.Context) {
...
-	for _, target := range targets {
-		for _, c := range checks {
+	for _, c := range checks {
+		minAgentVersion, err := getMinAgentVersionForCheckType(c.Type)
+		if err != nil {
+			s.l.Errorf("failed to get min agents version for type %s, reason: %s", c.Type, err)
+			continue
+		}
+
+		targets, err := s.findTargets(models.MySQLServiceType, minAgentVersion)
```
BupycHuk, line 314: looks like we can just pass c.Type into findTargets and call getMinAgentVersionForCheckType there to not copy-paste it everywhere.

artemgavrilov: Technically yes, but I think it will be monster-complex function. Now it knows only about services\agents, if we move `getMinAgentVersionForCheckType` then it also should know about checks. For me better to have this logic separately 🧐

services/checks/checks.go

services/checks/checks.go
artemgavrilov: @BupycHuk ^ what do you think?

services/checks/checks.go
```diff
@@ -244,6 +246,31 @@ func (s *Service) waitForResult(ctx context.Context, resultID string) ([]map[str
...
+		fallthrough
+	case check.PostgreSQLShow:
+		fallthrough
+	case check.MongoDBBuildInfo:
+		fallthrough
+	case check.MongoDBGetParameter:
+		return version.Parse("2.6.0")
+
+	case check.MongoDBGetCmdLineOpts:
+		return version.Parse("2.6.0")
```
AlekSi, line 267: That should be 2.7.0.

Those also could be global variables so we don't have to re-parse them every time.

services/checks/checks.go
BupycHuk: I'm ok with that, but we duplicated this code for each service type and we won't use them independently.

services/checks/checks.go
```diff
@@ -274,16 +301,23 @@ func (s *Service) executeChecks(ctx context.Context) {
...
-
 	checks := s.getMySQLChecks()
 
 	var res []string
-	for _, target := range targets {
-		for _, c := range checks {
+	for _, c := range checks {
+		minAgentVersion, err := getMinAgentVersionForCheckType(c.Type)
+		if err != nil {
+			s.l.Errorf("failed to get min agents version for type %s, reason: %s", c.Type, err)
```
AlekSi, line 310: In general, we use warnings for handled errors (that are not visible to users). Please do that everywhere.

AlekSi: Also, the error message would be really bad:

> failed to get min agents version for type MONGODB_GETCMDLINEOPTS, reason: missing min agent version for check type: MONGODB_GETCMDLINEOPTS

It should not repeat itself and be properly formatted full sentence. Please do that everywhere.

services/checks/checks.go
AlekSi: I think the best solution could be to store checks with additional data:
```
type checkInfo struct {
  check.Check
  minAgentVersion *version.Parsed
}
…
mySQLChecks      []checkInfo
```

Then findTargets could use that info.

But let's do that during our next refactoring. I think there are other bigger pieces that could be rearranged.

services/checks/checks.go
artemgavrilov: Damn, left after testing 🤦‍♂️

services/checks/checks.go
```diff
@@ -70,6 +71,11 @@ const (
 	maxSupportedVersion = 1
 )
 
+var (
+	pmmAgentVersionTwoDotSix   *version.Parsed
+	pmmAgentVersionTwoDotSeven *version.Parsed
+)
```
artemgavrilov, line 77: Yes, it's ugly, but the other option was `pmmAgentVersion_2_6_7`

AlekSi: I'm pretty sure there are more options :)

services/checks/checks.go
```diff
@@ -96,6 +102,10 @@ type Service struct {
 // New returns Service with given PMM version.
 func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform.DB, pmmVersion string) *Service {
 	l := logrus.WithField("component", "checks")
+
+	pmmAgentVersionTwoDotSix = mustParseVersion("2.6.0")
+	pmmAgentVersionTwoDotSeven = mustParseVersion("2.7.0")
```
AlekSi, line 107: We shouldn't modify a global state in a function. Either use a Service field, or `init()`.

artemgavrilov: `init` then. To be honest, it's exactly how I want to initialize those variables at first, but we have strange linter that asks to not use `init` functions.

services/checks/checks.go
artemgavrilov: My imagination has run out for today 😔. May be tomorrow I will find better names.

services/checks/checks_test.go
```diff
@@ -264,4 +265,51 @@ func TestFindTargets(t *testing.T) {
 	targets, err := s.findTargets(models.PostgreSQLServiceType, nil)
 	require.NoError(t, err)
 	assert.Len(t, targets, 0)
+
+	t.Skip("FIXME")
+
+	mysql, err := models.AddNewService(db.Querier, models.MySQLServiceType, &models.AddDBMSServiceParams{
+		ServiceName: "mysql",
+		NodeID:      models.PMMServerNodeID,
```
artemgavrilov: That test failed because we ignore services on pmm server node

services/checks/checks.go
```diff
@@ -226,24 +234,48 @@ func (s *Service) waitForResult(ctx context.Context, resultID string) ([]map[str
...
 		}
 
-		// FIXME we still need to delete old result - they may never be done: https://jira.percona.com/browse/PMM-5840
 		if err = s.db.Delete(res); err != nil {
 			s.l.Warnf("Failed to delete action result %s: %s.", resultID, err)
 		}
 
 		if res.Error != "" {
-			return nil, errors.Errorf("Action %s failed: %s.", resultID, res.Error)
+			return nil, errors.Errorf("action %s failed: %s.", resultID, res.Error)
```
AlekSi, line 242: A dot at the end of the `error` message should be added only to errors that are visible to users as-is – in our case, that's only (or mostly) gRPC statuses. In that case, we intentionally break the best practices because of the way we return gRPC statuses. (And even that decision is somewhat unfortunate; see https://jira.percona.com/browse/PMM-4695)

In other cases, `error` messages should be sentence fragments that are wrapped by the caller. That is the best practice in Go.

TL;DR: there is an extra dot in the end.

services/checks/checks.go
```diff
@@ -226,24 +234,48 @@ func (s *Service) waitForResult(ctx context.Context, resultID string) ([]map[str
...
 		}
 
 		out, err := agentpb.UnmarshalActionQueryResult([]byte(res.Output))
 		if err != nil {
-			return nil, errors.Errorf("Failed to parse action result : %s.", err)
+			return nil, errors.Errorf("failed to parse action result : %s.", err)
```
AlekSi, line 247: Extra dot, extra space before `:`

services/checks/checks_test.go
```diff
@@ -261,10 +264,140 @@ func TestFindTargets(t *testing.T) {
...
+				targets, err := s.findTargets(models.MySQLServiceType, test.minRequiredVersion)
+				require.NoError(t, err)
+				assert.Len(t, targets, test.count)
+			})
+		}
+	})
+}
 
-	targets, err := s.findTargets(models.PostgreSQLServiceType)
+func setup(t *testing.T, db *reform.DB, serviceName, nodeID, pmmAgentVersion string) {
```
AlekSi, line 312: Somewhat strange to see `setup` function being defined after the test that uses it.

PR #404, opened 2020-06-08
PMM-5563 Annotation service, node.

JiriCtvrtka:
- [ ] https://github.com/percona/pmm/pull/539
- [ ] https://github.com/percona/pmm-managed/pull/404
- [ ] https://github.com/percona/pmm-admin/pull/113
- [ ] https://github.com/Percona-Lab/pmm-api-tests/pull/111
- [ ] https://github.com/Percona-Lab/pmm-submodules/pull/912


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/pmm-managed?pullRequest=404) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/pmm-managed?pullRequest=404) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/pmm-managed?pullRequest=404) it.</sub>

services/inventory/nodes.go
```diff
@@ -233,3 +235,26 @@ func (s *NodesService) AddRemoteRDSNode(ctx context.Context, req *inventorypb.Ad
 
 	return invNode.(*inventorypb.RemoteRDSNode), nil
 }
+
+// Check if node exists.
+//nolint:unparam
+func (s *NodesService) Check(ctx context.Context, req *inventorypb.CheckNodeRequest) (*inventorypb.CheckNodeResponse, error) {
```
BupycHuk, line 241: Any reason to have extra check?

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+	var service, node bool
+	var serviceErr, nodeErr error
+	if len(req.ServiceName) > 0 {
+		service = true
+	}
+	if req.NodeName != "" {
+		node = true
+	}
+
+	if service {
```
BupycHuk, line 70: all queries to db should be in one transaction.

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+		serviceErr = findService(as.db, req.ServiceName)
+	}
+
+	if node {
+		nodeErr = findNode(as.db, req.NodeName)
+	}
+
+	if serviceErr != nil && nodeErr != nil || serviceErr != nil {
```
BupycHuk, line 78: this looks weird. Why don't it placed right after `findService()`

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+	if !node && !service {
+		_, err := as.grafanaClient.CreateAnnotation(ctx, req.Tags, time.Now(), req.Text, authorizationHeaders[0])
+		if err != nil {
+			return nil, err
+		}
 	}
 	return &managementpb.AddAnnotationResponse{}, nil
 }
+
+func findService(db *reform.DB, serviceName []string) error {
```
BupycHuk, line 113: `findServices`

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+	}
+
+	if nodeErr != nil {
+		return nil, nodeErr
+	}
+
+	if service {
+		for _, sn := range req.ServiceName {
+			text := fmt.Sprintf("%s (Service Name: %s)", req.Text, sn)
+			_, err := as.grafanaClient.CreateAnnotation(ctx, req.Tags, time.Now(), text, authorizationHeaders[0])
```
BupycHuk, line 89: As I said before we need only one annotation with all tags

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	if node {
+		text := fmt.Sprintf("%s (Node Name: %s)", req.Text, req.NodeName)
+		_, err := as.grafanaClient.CreateAnnotation(ctx, req.Tags, time.Now(), text, authorizationHeaders[0])
```
BupycHuk, line 98: this annotation will have `pmm_annotations` tag

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+		return nil, serviceErr
+	}
+
+	if nodeErr != nil {
+		return nil, nodeErr
+	}
+
+	if service {
+		for _, sn := range req.ServiceName {
+			text := fmt.Sprintf("%s (Service Name: %s)", req.Text, sn)
```
BupycHuk, line 88: Please clarify this with @rnovikovP in cases when we have more than one service

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
 		return nil, status.Error(codes.Unauthenticated, "Authorization error.")
 	}
 
-	_, err := as.grafanaClient.CreateAnnotation(ctx, req.Tags, time.Now(), req.Text, authorizationHeaders[0])
-	if err != nil {
-		return nil, err
+	var service, node bool
```
BupycHuk, line 61: Why do we need separate variable for this? Why don't just include everything belongs to `service = true` into this `if`.

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+	var serviceErr, nodeErr error
+	if len(req.ServiceName) > 0 {
+		service = true
+	}
+	if req.NodeName != "" {
```
BupycHuk, line 66: the same with `req.NodeName != ""`

services/management/grpc/annotation_server.go
BupycHuk, line 78: `serviceErr != nil` should be enough

services/management/grpc/annotation_server.go
askomorokhov: or `isAllServicesExist` - as services are not returned here.

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func findNode(db *reform.DB, nodeName string) error {
```
askomorokhov, line 131: Maybe `isNodeExists` and return bool?

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+	var serviceErr, nodeErr error
+	if len(req.ServiceName) > 0 {
+		service = true
+	}
+	if req.NodeName != "" {
+		node = true
+	}
+
+	if service {
+		serviceErr = findService(as.db, req.ServiceName)
```
askomorokhov, line 71: I guess it will be more simple: 

for serviceName in range req.ServiceNames {
    if  _, err = models.FindServiceByName(tx.Querier, sn); err != nil {
 	return  fmt.Errof("%s human readable error service with such name is not registered.", serviceName)
 }
}

and same for nodes.

services/management/grpc/annotation_server.go
askomorokhov: @BupycHuk, why do we need to do a select query in the transaction?
If a user removes service later - annotation will be left in grafana.

services/management/grpc/annotation_server.go
askomorokhov: or this function even redundant (with similar `findNode` below).

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +58,88 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+
+	if nodeErr != nil {
+		return nil, nodeErr
+	}
+
+	if service {
+		for _, sn := range req.ServiceName {
+			text := fmt.Sprintf("%s (Service Name: %s)", req.Text, sn)
+			_, err := as.grafanaClient.CreateAnnotation(ctx, req.Tags, time.Now(), text, authorizationHeaders[0])
+			if err != nil {
```
askomorokhov, line 90: You need to append to user-defined tags service names and node names

curl 'http://admin:admin@localhost/graph/api/annotations' -H 'Content-Type: application/json;charset=UTF-8' --data-binary '{"time":1585053244718,"tags":["user specified tag1", "user specified tag2", "ps1", "ps2", "ps3", "node1"],"text":"Alarm!!! (Service Name: ps1,ps2,ps3, Node Name:node1)"}'

services/management/grpc/annotation_server.go
```diff
@@ -54,9 +59,41 @@ func (as *AnnotationServer) AddAnnotation(ctx context.Context, req *managementpb
...
+		if err != nil {
+			return nil, err
+		}
+
+		tags = append(tags, req.NodeName)
+		postfix = append(postfix, "Node Name: "+req.NodeName)
+	}
+
+	if len(postfix) > 0 {
+		req.Text += "(" + strings.Join(postfix, ",") + ")"
```
askomorokhov, line 90:
```suggestion
		req.Text += " (" + strings.Join(postfix, ", ") + ")"
```

services/management/annotation_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/percona/pmm-managed/utils/logger"
+	"github.com/percona/pmm-managed/utils/testdb"
+	"github.com/percona/pmm-managed/utils/tests"
+
+	"github.com/percona/pmm/api/managementpb"
+)
+
+type grafanaClientMock struct {
+}
```
BupycHuk, line 40: please check other mocks and generate it like them

services/management/annotation.go
```diff
@@ -0,0 +1,85 @@
...
+	"strings"
+	"time"
+
+	"github.com/percona/pmm/api/managementpb"
+	"gopkg.in/reform.v1"
+
+	"github.com/percona/pmm-managed/models"
+)
+
+type grafanaClient interface {
```
askomorokhov, line 30: @JiriCtvrtka, as Nurlan mentioned it is more convenient to move this interface to https://github.com/percona/pmm-managed/blob/PMM-2.0/services/management/deps.go

and add a similar line as https://github.com/percona/pmm-managed/blob/8a6d53d26120ccdb1314bf234405092530afa4e1/services/management/deps.go#L29

then `make gen` will generate mock as https://github.com/percona/pmm-managed/blob/PMM-2.0/services/management/mock_checks_service_test.go
to include in `annotation_test.go`

services/management/deps.go
```diff
@@ -48,3 +50,9 @@ type prometheusService interface {
 type checksService interface {
 	StartChecks(ctx context.Context) error
 }
+
+// grafanaClient is a subset of methods of annotation.Service used by this package.
```
askomorokhov, line 54:
```suggestion
// grafanaClient is a subset of methods of grafana.Client used by this package.
```

services/management/annotation_test.go
```diff
@@ -0,0 +1,182 @@
...
+		_, err = models.AddNewService(s.db.Querier, models.MySQLServiceType, &models.AddDBMSServiceParams{
+			ServiceName: "service-test2",
+			NodeID:      models.PMMServerNodeID,
+			Address:     pointer.ToString("127.0.0.1"),
+			Port:        pointer.ToUint16(3306),
+		})
+		require.NoError(t, err)
+
+		expectedText := "Some text (Service Name: service-test1,service-test2)"
+		grafanaClient.On("CreateAnnotation", ctx, mock.Anything, mock.Anything, expectedText, mock.Anything).Return("", nil)
```
BupycHuk, line 119: let's add `expectedTags` too

services/management/annotation_test.go
```diff
@@ -0,0 +1,182 @@
...
+			ServiceNames: []string{"service-test1", "service-test2"},
+		})
+		require.NoError(t, err)
+	})
+
+	t.Run("Non-existing node", func(t *testing.T) {
+		ctx, db, teardown := setup(t)
+		defer teardown(t)
+		var grafanaClient mockGrafanaClient
+		grafanaClient.On(mock.Anything).Return()
```
BupycHuk, line 131: Do we really need this line?

services/management/annotation_test.go
```diff
@@ -0,0 +1,182 @@
...
+		s := NewAnnotationService(db, &grafanaClient)
+
+		_, err := s.AddAnnotation(ctx, autorization, &managementpb.AddAnnotationRequest{
+			Text:         "Some text",
+			ServiceNames: []string{"no-node"},
+		})
+		require.Error(t, err)
```
BupycHuk, line 138: let's use `require.EqualError`

services/management/annotation_test.go
BupycHuk: and `assertExpectations` at the end of the test.

BupycHuk: The same in other places where we expect calling `CreateAnnotation`

services/management/annotation_test.go
```diff
@@ -0,0 +1,182 @@
...
+			Text:         "Some text",
+			ServiceNames: []string{"service-test1", "service-test2"},
+		})
+		require.NoError(t, err)
+	})
+
+	t.Run("Non-existing node", func(t *testing.T) {
+		ctx, db, teardown := setup(t)
+		defer teardown(t)
+		var grafanaClient mockGrafanaClient
```
BupycHuk, line 130: var grafanaClient = new(mockGrafanaClient)

services/management/grpc/annotation_server.go
```diff
@@ -20,25 +20,27 @@ package grpc
...
-	"time"
 
 	"github.com/percona/pmm/api/managementpb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
+	"gopkg.in/reform.v1"
 
 	"github.com/percona/pmm-managed/services/grafana"
+
```
BupycHuk, line 31: unnecessary line

services/management/annotation_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/stretchr/testify/require"
+	"gopkg.in/reform.v1"
+	"gopkg.in/reform.v1/dialects/postgresql"
+
+	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/utils/logger"
+	"github.com/percona/pmm-managed/utils/testdb"
+	"github.com/percona/pmm-managed/utils/tests"
+
+	"github.com/percona/pmm/api/managementpb"
```
BupycHuk, line 36: it should be in the second block.

services/management/annotation_test.go
```diff
@@ -0,0 +1,195 @@
...
+		ctx, db, teardown := setup(t)
+		defer teardown(t)
+		var grafanaClient = new(mockGrafanaClient)
+		s := NewAnnotationService(db, grafanaClient)
+
+		_, err := s.AddAnnotation(ctx, autorization, &managementpb.AddAnnotationRequest{
+			Text:         "Some text",
+			ServiceNames: []string{"no-service"},
+		})
+		require.EqualError(t, errors.New("rpc error: code = NotFound desc = Service with name \"no-service\" not found."), err.Error())
```
AlekSi, line 69: `tests.AssertGRPCError`

services/management/annotation_test.go
```diff
@@ -0,0 +1,195 @@
...
+			ServiceName: "service-test",
+			NodeID:      models.PMMServerNodeID,
+			Address:     pointer.ToString("127.0.0.1"),
+			Port:        pointer.ToUint16(3306),
+		})
+		require.NoError(t, err)
+
+		expectedTags := []string{"service-test"}
+		expectedText := "Some text (Service Name: service-test)"
+		grafanaClient.On("CreateAnnotation", ctx, expectedTags, mock.Anything, expectedText, autorization[0]).Return("", nil)
```
AlekSi, line 90: Why do we mock it? Essentially, we are testing a mock there.

services/management/annotation.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+}
+
+// AddAnnotation create annotation in grafana
+func (as *AnnotationService) AddAnnotation(ctx context.Context, authorizationHeaders []string, req *managementpb.AddAnnotationRequest) (*managementpb.AddAnnotationResponse, error) {
+	tags := req.Tags
+	if len(req.ServiceNames) == 0 && req.NodeName == "" {
+		tags = append([]string{"pmm_annotation"}, tags...)
+	}
+	postfix := []string{}
```
AlekSi, line 50:
```go
var postfix []string
```

https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_use_a_consistent_declaration_style (but please read the whole piece)

services/management/annotation_test.go
BupycHuk: No, we are testing logic in AnnotationService. It's unit tests, not integration.

BupycHuk: We have separate tests for grafana client to check integration with grafana.

services/management/annotation.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(req.ServiceNames) > 0 {
+		for _, sn := range req.ServiceNames {
+			_, err := models.FindServiceByName(as.db.Querier, sn)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		tags = append(tags, req.ServiceNames...)
+		postfix = append(postfix, "Service Name: "+strings.Join(req.ServiceNames, ","))
```
AlekSi, line 60: `", "`?

services/management/annotation.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		tags = append(tags, req.NodeName)
+		postfix = append(postfix, "Node Name: "+req.NodeName)
+	}
+
+	if len(postfix) > 0 {
+		req.Text += " (" + strings.Join(postfix, ", ") + ")"
```
AlekSi, line 74: I think that should be `.` or `;`. Do we have a test for that? I think we might want to use @PaulJacobs-percona help with messages.

services/management/annotation_test.go
AlekSi: I understand your point, but: Does it really have to be a unit test with mocks instead of a real thing? The whole point of introducing a devcontainer was to be able to write tests that use real things, not mocks. We use mocks only we really have too. That philosophy runs in the whole PMM 2.x codebase, so I don't think we suddenly should change that now in a single place.

BupycHuk: 1. What will be difference between API tests and tests in pmm-managed? When should we write API tests and when tests in pmm-managed?
2. We already have places where we use mocks instead of real component(QAN, Prometheus), but we can use real component.
3. We have to export FindAnnotations method in grafana.Client only for tests, which I don't think is a good idea.
According to the description of https://jira.percona.com/browse/PMM-4301, devcontainer was introduced to make it easy to use for everybody.

services/management/annotation.go
```diff
@@ -0,0 +1,88 @@
...
+// NewAnnotationService create new Annotation Service.
+func NewAnnotationService(db *reform.DB, grafanaClient grafanaClient) *AnnotationService {
+	return &AnnotationService{
+		db:            db,
+		grafanaClient: grafanaClient,
+	}
+}
+
+// AddAnnotation create annotation in grafana.
+// nolint: unparam
```
AlekSi, line 45: `//nolint`, no space for "directives"

services/management/annotation.go
```diff
@@ -0,0 +1,88 @@
...
+func (as *AnnotationService) AddAnnotation(
+	ctx context.Context,
+	authorizationHeaders []string,
+	req *managementpb.AddAnnotationRequest,
```
AlekSi, line 49: If `lll` linter complains in cases like that, feel free to ignore it, or add to `nolint`.

@BupycHuk thoughts about disabling `lll` everywhere? It is not like we use very long lines in other cases

services/management/annotation_test.go
AlekSi: 1. We should write API tests if our change affects API, and we should always write tests in pmm-managed repo if that's possible. Tests in pmm-managed repo should not use mocks if possible. In particular, I don't think we should call them "unit tests" as this is a loaded term, and we don't test only "that" package in our tests.
2. Please share examples. To the best of my knowledge, we use mocks in two cases:
* where the usage of a real object would cause import cycles that would require us to shift our code around and make it less organized just for the sake of the tests;
* where we can't trigger some edge case, but we still want to test it (in most cases, we can trigger it and do it);
Both points do not apply there.
3. I semi-agree with that point but don't think it overweight the usage of mocks.

Most importantly, our discussion about changing the usage of mocks should be global, and not a part of a single PR review (but that PR may trigger that discussion).

BupycHuk:
I'm also surprised why reviewdog doesn't work, I definitely checked it. Will fix it.

services/management/annotation.go
JiriCtvrtka: Ok, next time I will know. Thank you.

services/management/annotation_test.go
BupycHuk: https://github.com/percona/pmm-managed/blob/8e97c4ebaae1388610b5a4af81596bcd467e78f3/services/qan/client_test.go#L128
https://github.com/percona/pmm-managed/blob/8272f31746c5e1d654f895702b97e12ba57ea2e1/services/inventory/agents_test.go#L56
https://github.com/percona/pmm-managed/blob/4b73e3215b39a42c1babc97ecc09b1425fe4a805/services/management/service_test.go#L53

What's the difference between prometheusClient and grafanaClient?

JiriCtvrtka: Need to check it, but if we need real test, then I will remove that mock as well.

JiriCtvrtka:
Build fails because of:
Package "github.com/percona/pmm-managed/services/management" should not import package "github.com/percona/pmm-managed/services/grafana" (blacklisted by "github.com/percona/pmm-managed/services/").

Also here was some lint warnings when I made findAnnotations public. Due to I made public also annotation struct

services/management/annotation_test.go
BupycHuk: @AlekSi ^

services/management/annotation.go
```diff
@@ -0,0 +1,90 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package management
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/percona/pmm-managed/services/grafana"
```
BupycHuk, line 24: it should be in last block

services/management/annotation.go
BupycHuk: +1

services/management/annotation_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/AlekSi/pointer"
+	"github.com/percona/pmm/api/inventorypb"
+	"github.com/percona/pmm/api/managementpb"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
BupycHuk, line 30: unnecessary blank line

services/management/annotation_test.go
```diff
@@ -0,0 +1,230 @@
...
+		NodeName: "test-node",
+		Address:  "some.address.org",
+		Region:   "region",
+	})
+	require.NoError(t, err)
+
+	p := new(mockPrometheusService)
+	p.Test(t)
+	s := NewServiceService(db, r, p)
+	_, err = models.AddNewService(s.db.Querier, models.MySQLServiceType, &models.AddDBMSServiceParams{
```
BupycHuk, line 80: let's just use db.Querier? Any reason to use it from ServiceService?

services/management/annotation_test.go
```diff
@@ -0,0 +1,230 @@
...
+		NodeType: inventorypb.NodeType_GENERIC_NODE,
+		NodeName: "test-node",
+		Address:  "some.address.org",
+		Region:   "region",
+	})
+	require.NoError(t, err)
+
+	p := new(mockPrometheusService)
+	p.Test(t)
+	s := NewServiceService(db, r, p)
```
BupycHuk, line 79: I don't think it's required

services/management/annotation_test.go
```diff
@@ -0,0 +1,230 @@
...
+		return
+	}
+
+	ctx, db, teardown := setup(t)
+	defer teardown(t)
+
+	r := new(mockAgentsRegistry)
+	r.Test(t)
+	n := NewNodeService(db, r)
+	_, err := n.Register(ctx, &managementpb.RegisterNodeRequest{
```
BupycHuk, line 69: let's use `models.CreateNode`

BupycHuk: Could you check do we remove created node after test?

services/management/annotation_test.go
```diff
@@ -0,0 +1,230 @@
...
+		Address:     pointer.ToString("127.0.0.1"),
+		Port:        pointer.ToUint16(3306),
+	})
+	require.NoError(t, err)
+
+	c := grafana.NewClient("127.0.0.1:3000")
+	req, err := http.NewRequest("GET", "/dummy", nil)
+	require.NoError(t, err)
+	req.SetBasicAuth("admin", "admin")
+	authorization := []string{req.Header.Get("Authorization")}
```
BupycHuk, line 99: `headers`?

BupycHuk, line 99: let's use separate variable for auth header to reuse it in tests.

AlekSi:
> Build fails because of:
> Package "github.com/percona/pmm-managed/services/management" should not import package "github.com/percona/pmm-managed/services/grafana" (blacklisted by "github.com/percona/pmm-managed/services/").

This test exists exactly for enforcing code boundaries to keep code organized. Above I said, "Both points do not apply there." The first point _do_ apply here; I stand corrected. Let's use a mock in that case.

services/management/annotation.go
PaulJacobs-percona: A comma separating items in this context is fine. A full-stop (period) at the end if the text reads as a sentence but generally console-style messages don't. Depends how much coding is needed to make it look pretty.

JiriCtvrtka: After discuss with Paul:
(Service Name: x,y,z, Node Name: n) => (Service Name: x, y, z. Node Name: n)

services/grafana/client_test.go
```diff
@@ -163,7 +163,7 @@ func TestClient(t *testing.T) {
 			require.NoError(t, err)
 			for _, a := range annotations {
 				if a.Text == "No tags" {
-					assert.Equal(t, []string{"pmm_annotation"}, a.Tags)
+					assert.Equal(t, []string{}, a.Tags)
```
BupycHuk, line 166: maybe check for zero length?

services/management/annotation_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+			require.NoError(t, sqlDB.Close())
+		}
+
+		return
+	}
+	authorization := "admin:admin"
+	authorizationHeaders := []string{authorization}
+
+	t.Run("Non-existing service", func(t *testing.T) {
```
BupycHuk, line 60: Let's add one more test with no services and no node in the request. To check that we add "pmm_annotations" tag.

services/management/annotation_test.go
```diff
@@ -0,0 +1,267 @@
...
+)
+
+func TestAnnotations(t *testing.T) {
+	setup := func(t *testing.T) (ctx context.Context, db *reform.DB, teardown func(t *testing.T)) {
+		t.Helper()
+
+		ctx = logger.Set(context.Background(), t.Name())
+		uuid.SetRand(new(tests.IDReader))
+
+		sqlDB := testdb.Open(t, models.SetupFixtures, nil)
```
BupycHuk, line 46: Let's skip fixtures since we don't use them.

services/management/annotation_test.go
JiriCtvrtka: Here is test for it. Line 139. "Non-existing service and non-existing node".

services/management/annotation_test.go
JiriCtvrtka: OK.

services/grafana/client_test.go
JiriCtvrtka: Sure I check check len for Tags. Will change it.

services/management/annotation_test.go
JiriCtvrtka: @BupycHuk With models.SkipFixtures I getting error: rpc error: code = NotFound desc = Node with ID "pmm-server" not found. I using this node for tests where is only service to save some lines of creating another.

BupycHuk: oh, yeah, missed that.

services/management/annotation_test.go
BupycHuk: it's when we send request with non-existing service and node, but there might be request with empty services and node.

JiriCtvrtka: Ok I will add it.

JiriCtvrtka:
FB: https://github.com/Percona-Lab/pmm-submodules/pull/895

services/grafana/client_test.go
BupycHuk: check should be changed in assert, not in if. :)

JiriCtvrtka: Sorry, I read it wrong. Fixed.

AlekSi:
Sorry, but FB is not correct: https://github.com/Percona-Lab/pmm-submodules/pull/895 – there are not changes in branches

JiriCtvrtka:
Link on new FB:
https://github.com/Percona-Lab/pmm-submodules/pull/912

PR #405, opened 2020-06-09
SAAS-125 Send EndsAt with alerts to auto-resolve them

palash25:
[Jira link](https://jira.percona.com/browse/SAAS-125)

- [API Tests](https://github.com/Percona-Lab/pmm-api-tests/pull/114)
- [FB](https://github.com/Percona-Lab/pmm-submodules/pull/925)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/pmm-managed?pullRequest=405) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/pmm-managed?pullRequest=405) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/pmm-managed?pullRequest=405) it.</sub>

services/alertmanager/registry_test.go
```diff
@@ -53,6 +55,7 @@ func TestRegistry(t *testing.T) {
 
 		expected := &ammodels.PostableAlert{
 			Annotations: annotations,
+			EndsAt:      strfmt.DateTime(nowValueCopy.Add(resolveTimeoutFactor * resendInterval)),
```
AlekSi, line 58: Add method does not modify the receiver, so there is no need to use `nowValueCopy`

palash25: I tried `nowValue` but the test failed because the `ext` field in `time.Time` didn't match so I made a copy

palash25:
@AlekSi I have made `resendInterval` configurable, PTAL.

services/alertmanager/registry_test.go
palash25:
```
--- FAIL: TestRegistry (0.00s)
    --- FAIL: TestRegistry/DelayFor (0.00s)
        registry_test.go:68: 
            	Error Trace:	registry_test.go:68
            	Error:      	Not equal: 
            	            	expected: &ammodels.PostableAlert{Annotations:ammodels.LabelSet{"annotation":"test"}, EndsAt:strfmt.DateTime{wall:0xbfb03dcac15cd572, ext:151570362490, loc:(*time.Location)(0x11c76e0)}, StartsAt:strfmt.DateTime{wall:0x0, ext:0, loc:(*time.Location)(nil)}, Alert:ammodels.Alert{GeneratorURL:"", Labels:ammodels.LabelSet{"label":"demo"}}}
            	            	actual  : &ammodels.PostableAlert{Annotations:ammodels.LabelSet{"annotation":"test"}, EndsAt:strfmt.DateTime{wall:0xbfb03dbbc15cd572, ext:91570362490, loc:(*time.Location)(0x11c76e0)}, StartsAt:strfmt.DateTime{wall:0x0, ext:0, loc:(*time.Location)(nil)}, Alert:ammodels.Alert{GeneratorURL:"", Labels:ammodels.LabelSet{"label":"demo"}}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -4,3 +4,3 @@
            	            	  },
            	            	- EndsAt: (strfmt.DateTime) 2020-06-10T06:37:31.022Z,
            	            	+ EndsAt: (strfmt.DateTime) 2020-06-10T06:36:31.022Z,
            	            	  StartsAt: (strfmt.DateTime) 0001-01-01T00:00:00.000Z,
            	Test:       	TestRegistry/DelayFor
FAIL

```

services/alertmanager/alertmanager.go
```diff
@@ -35,12 +35,17 @@ import (
 )
 
 const (
-	resendInterval      = 30 * time.Second
-	alertmanagerDataDir = "/srv/alertmanager/data"
-	prometheusDir       = "/srv/prometheus"
-	path                = "/srv/alertmanager/alertmanager.base.yml"
+	// Environment variable to overwrite resendInterval during testing
+	envInterval = "PERCONA_TEST_ALERTMANAGER_INTERVAL"
```
AlekSi, line 39: Let's call it `envResendInterval` and `PERCONA_TEST_ALERTMANAGER_RESEND_INTERVAL` to match `resendInterval` below.

services/alertmanager/alertmanager.go
```diff
@@ -35,12 +35,17 @@ import (
...
+
+	resolveTimeoutFactor = 3
+	alertmanagerDataDir  = "/srv/alertmanager/data"
+	prometheusDir        = "/srv/prometheus"
+	path                 = "/srv/alertmanager/alertmanager.base.yml"
 )
 
+var resendInterval = 30 * time.Second
```
AlekSi, line 47: Let's make it a `Service` field, not a package-level variable.

services/alertmanager/alertmanager.go
```diff
@@ -50,6 +55,13 @@ type Service struct {
 
 // New creates new service.
 func New(db *reform.DB, alertsRegistry *Registry) *Service {
+	l := logrus.WithField("component", "alertmanager")
+
+	if d, err := time.ParseDuration(os.Getenv(envInterval)); err == nil && d > 0 {
+		l.Warnf("Interval changed to %s.", d)
+		resendInterval = d
```
AlekSi, line 62: `New` "constructor" should not modify non-local state. Make resendInterval a field.

services/alertmanager/registry_test.go
AlekSi: Well, that's because we modify nowValue just below :) I think we should change that

services/alertmanager/alertmanager.go
palash25: I had thought of that first but since the EndsAt is added in a `Registry` method, I couldn't think of how will `CreateAlert` acquire the value of the `resendInterval` if it were a field of `Service` cc @AlekSi

AlekSi: Then it should be in a registry.go and set during process initialisation.
The problem with the current code that `New` "constructor" also changes global state. That's unexpected.

services/alertmanager/registry.go
```diff
@@ -55,6 +56,7 @@ func (r *Registry) CreateAlert(id string, labels, annotations map[string]string,
 			Labels: labels,
 		},
 
+		EndsAt: strfmt.DateTime(now().Add(resolveTimeoutFactor * resendInterval)),
 		// StartsAt and EndAt can't be added there without changes in Registry
```
AlekSi, line 60: Actually, that comment is correct – `delayFor` will not work correctly with `EndsAt` without modifications.

services/alertmanager/registry_test.go
```diff
@@ -20,13 +20,15 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestRegistry(t *testing.T) {
 	nowValue, origNow := now(), now
+	nowValueCopy := nowValue
 	now = func() time.Time {
 		return nowValue
 	}
```
artemgavrilov, line 34: I'm not sure, but maybe It's better to pass something that implements `Now() time.Time` as a parameter to `NewRegistry()`

That code is not obvious and takes me some time to figure out what is happening here.

services/alertmanager/registry_test.go
palash25: but if we don't modify the `now` value below then we won't get an alert from the `r.colect()` call

BupycHuk: so maybe we should extract value of `endsAt` to a variable before we change `now`?

services/alertmanager/registry.go
```diff
@@ -17,11 +17,21 @@
...
+	"github.com/go-openapi/strfmt"
 	"github.com/percona/pmm/api/alertmanager/ammodels"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// Environment variable to overwrite resendInterval during testing
+	envResendInterval     = "PERCONA_TEST_ALERTMANAGER_RESEND_INTERVAL"
+	resolveTimeoutFactor  = 3
+	defaultResendInterval = 30 * time.Second
```
artemgavrilov, line 34: I think `envResendInterval` and `defaultResendInterval`  variables should be declared in `alertmanager.go` as resend logic declared there.

services/alertmanager/registry.go
```diff
@@ -93,7 +115,9 @@ func (r *Registry) collect() ammodels.PostableAlerts {
 	now := now()
 	for id, t := range r.times {
 		if t.Before(now) {
-			res = append(res, r.alerts[id])
+			alert := r.alerts[id]
+			alert.EndsAt = strfmt.DateTime(now.Add(resolveTimeoutFactor * r.resendInterval))
```
artemgavrilov, line 119: `resolveTimeoutFactor * r.resendInterval` better precalculate that value in the constructor as it never changed after service start. You can name it `alertTTL` for example.

palash25:
@AlekSi @BupycHuk Artem suggested some fixes which made the tests pass without keeping a copy of the `nowValue` PTAL.

AlekSi:
@palash25 
image

services/alertmanager/registry.go
palash25: but they are used in `registry.go` so it didn't make sense for me to move them out to `alertmanager.go`, let me know if you still insist on doing this and I will make the change

services/alertmanager/registry.go
palash25: done

services/alertmanager/registry.go
palash25: if i recall correctly in your draft PR you were using resend interval and env var in your `alertmanager.go` code, that is why it made sense to put them there, but here I have moved these to registry because Alexy wanted to make `resendInterval` a `registry` field  https://github.com/percona/pmm-managed/pull/405#discussion_r438031028

artemgavrilov: My point is that `resendInterval` is not `registry` value, as the registry doesn't resend anything. The registry requires another value, `alert TTL` that depends on `resendInterval` value.  In my draft PR both variables are fields of corresponding services, but also there is a constant `defaultResendInterval`. So both serviced just check env variable to figure out was `defaultResendInterval` overwritten or not.  If it's not, then alertmanager uses `defaultResendInterval` as `resendInterval` and `registry` uses `defaultResendInterval` to calculate `alertTTL` (`resolveTimeoutFactor * defaultResendInterval`). Otherwise both use value of env variable for that.

For me, such approach looks more natural, as all variables declared in expected files. 

@AlekSi what do you think?  

p.s. here is the draft we discussing: https://github.com/percona/pmm-managed/pull/412/files

AlekSi: While I agree with most of it, it sounds very concerning that we use _environment variable_ to pass information in code. Not function parameters, method parameters, global constants, etc – envrionment variables. Something that compiler and linters can't check.

I don't know what is the best way to solve this problem, but using "global" environenment variable is not it.

I wonder if changing the scope of "registry" from "it stores all generic alerts" to "it stores current STT alerts" and moving it to checks package, and then passing alerts from check's registry to alertmanager problem will solve the problem. Alert type definition (without Alertmanager and STT details, generic enough) can be in shared `services` package.

palash25: > I wonder if changing the scope of "registry" from "it stores all generic alerts" to "it stores current STT alerts" and moving it to checks package, and then passing alerts from check's registry to alertmanager problem will solve the problem. Alert type definition (without Alertmanager and STT details, generic enough) can be in shared services package.

so should we try doing this as a part of this ticket or a new ticket?

AlekSi: Let's quickly try it in this ticket

palash25: > I wonder if changing the scope of "registry" from "it stores all generic alerts" to "it stores current STT alerts" and moving it to checks package, and then passing alerts from check's registry to alertmanager problem will solve the problem. Alert type definition (without Alertmanager and STT details, generic enough) can be in shared services package.

@aleksi I am not sure how this would help us eliminate the use of `resendInterval` as an env var (can you elaborate on this?), even if I move the registry to `checks` package, I will need a `resendInterval` value either as a global constant or env var to set the `EndsAt` field. Why not just keep the code where it is and pass `resendInterval` as a parameter to `NewRegistry`

AlekSi: That should allow us to un-export `registry` type and make it an implementation detail. In turn, that should allow us to make `checks` and `alertmanager` services less coupled

palash25: > That should allow us to un-export registry type

Currently we are exporting `Registry` `alertmanager` to make it visible in the `checks` package but if we move `Registry` to `checks` we will have to export it anyway to make it visible in `alertmanager` @AlekSi

AlekSi: Let’s continue with the fastest way for now. It looks like we will have to refactor this part anyway. I have an idea how to do that, but it will take more time

AlekSi:
@palash25 @artemgavrilov please check my recent commits there

palash25:
LGTM @AlekSi

PR #407, opened 2020-06-12
SAAS-104 Basic Prometheus Metrics for STT

palash25:
[SAAS-104](https://jira.percona.com/browse/SAAS-104)

 - [ ] [A link to the feature build](https://github.com/Percona-Lab/pmm-submodules/pull/904)
 - [x] API tests: https://github.com/Percona-Lab/pmm-api-tests/pull/108
---
 - [x] Tests passed.
 - [ ] Feature build pass
 - [x] (Re)requested review.
 - [ ] Update API dependency.
---

Adds prometheus metrics for the checks service to record the number of check scripts executed and the number of alerts generated.

### DIFF
--- a/services/checks/checks.go
+++ b/services/checks/checks.go
@@ -64,9 +64,8 @@ const (
 	resultTimeout       = 15 * time.Second
 	resultCheckInterval = time.Second
 
-	// TODO https://jira.percona.com/browse/SAAS-104
-	// prometheusNamespace = "pmm_managed"
-	// prometheusSubsystem = "checks"
+	prometheusNamespace = "pmm_managed"
+	prometheusSubsystem = "checks"
 
 	alertsPrefix        = "/stt/"
 	maxSupportedVersion = 1
@@ -99,6 +98,9 @@ type Service struct {
 	mySQLChecks      []check.Check
 	postgreSQLChecks []check.Check
 	mongoDBChecks    []check.Check
+
+	mScriptsExecuted *prom.CounterVec
+	mAlertsGenerated *prom.CounterVec
 }
 
 // New returns Service with given PMM version.
@@ -115,6 +117,20 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
 		publicKeys: defaultPublicKeys,
 		interval:   defaultInterval,
 		startDelay: defaultStartDelay,
+
+		mScriptsExecuted: prom.NewCounterVec(prom.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "scripts_executed_total",
+			Help:      "Counter of check scripts executed per service type",
+		}, []string{"service_type"}),
+
+		mAlertsGenerated: prom.NewCounterVec(prom.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "alerts_generated_total",
+			Help:      "Counter of alerts generated per service type per check type",
+		}, []string{"service_type", "check_type"}),
 	}
 
 	if h := os.Getenv(envHost); h != "" {
@@ -130,6 +146,17 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
 		s.interval = d
 		s.startDelay = 0
 	}
+
+	s.mScriptsExecuted.WithLabelValues(string(models.MySQLServiceType))
+	s.mScriptsExecuted.WithLabelValues(string(models.PostgreSQLServiceType))
+	s.mScriptsExecuted.WithLabelValues(string(models.MongoDBServiceType))
+	s.mAlertsGenerated.WithLabelValues(string(models.MySQLServiceType), string(check.MySQLShow))
+	s.mAlertsGenerated.WithLabelValues(string(models.MySQLServiceType), string(check.MySQLSelect))
+	s.mAlertsGenerated.WithLabelValues(string(models.PostgreSQLServiceType), string(check.PostgreSQLShow))
+	s.mAlertsGenerated.WithLabelValues(string(models.PostgreSQLServiceType), string(check.PostgreSQLSelect))
+	s.mAlertsGenerated.WithLabelValues(string(models.MongoDBServiceType), string(check.MongoDBBuildInfo))
+	s.mAlertsGenerated.WithLabelValues(string(models.MongoDBServiceType), string(check.MongoDBGetCmdLineOpts))
+	s.mAlertsGenerated.WithLabelValues(string(models.MongoDBServiceType), string(check.MongoDBGetParameter))
 
 	return s
 }
@@ -346,6 +373,9 @@ func (s *Service) executeMySQLChecks(ctx context.Context) ([]string, error) {
 				s.l.Warnf("Failed to process action result: %s.", err)
 				continue
 			}
+
+			s.mScriptsExecuted.WithLabelValues(string(models.MySQLServiceType)).Inc()
+			s.mAlertsGenerated.WithLabelValues(string(models.MySQLServiceType), string(c.Type)).Add(float64(len(alerts)))
 			res = append(res, alerts...)
 		}
 	}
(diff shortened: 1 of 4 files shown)


services/checks/checks.go
```diff
@@ -346,6 +363,9 @@ func (s *Service) executeMySQLChecks(ctx context.Context) ([]string, error) {
 				s.l.Warnf("Failed to process action result: %s", err)
 				continue
 			}
+			// record metrics
+			s.mExecutedScripts.WithLabelValues(string(models.MySQLServiceType)).Inc()
+			s.mGeneratedAlerts.WithLabelValues(string(models.MySQLServiceType), string(c.Type)).Inc()
```
artemgavrilov, line 368: One check can produce multiple alerts, moreover `alerts` can be empty at all. So I think that counter should be incremented on `len(alerts)`.

services/checks/checks.go
```diff
@@ -395,6 +415,9 @@ func (s *Service) executePostgreSQLChecks(ctx context.Context) ([]string, error)
 				s.l.Warnf("Failed to process action result: %s", err)
 				continue
 			}
+			// record metrics
+			s.mExecutedScripts.WithLabelValues(string(models.PostgreSQLServiceType)).Inc()
+			s.mGeneratedAlerts.WithLabelValues(string(models.PostgreSQLServiceType), string(c.Type)).Inc()
```
artemgavrilov, line 420: Same as above.

services/checks/checks.go
```diff
@@ -450,6 +473,9 @@ func (s *Service) executeMongoDBChecks(ctx context.Context) ([]string, error) {
 				s.l.Warnf("Failed to process action result: %s", err)
 				continue
 			}
+			// record metrics
+			s.mExecutedScripts.WithLabelValues(string(models.MongoDBServiceType)).Inc()
+			s.mGeneratedAlerts.WithLabelValues(string(models.MongoDBServiceType), string(c.Type)).Inc()
```
artemgavrilov, line 478: Same as above.

services/checks/checks.go
```diff
@@ -115,6 +118,20 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
...
+			Subsystem: prometheusSubsystem,
+			Name:      "check_script_executed_total",
+			Help:      "Counter of check scripts executed per service type",
+		}, []string{"service_type"}),
+
+		mGeneratedAlerts: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "generated_alerts_total",
+			Help:      "Counter of request type rewrites",
```
artemgavrilov, line 133: Seems like wrong help message

artemgavrilov:
@palash25 also please fill the PR description. You can take text from https://github.com/percona/pmm-managed/pull/408/files

palash25:
done @artemgavrilov made the changes PTAL

palash25:
@artemgavrilov done, PTAL

services/checks/checks.go
```diff
@@ -115,6 +118,20 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
 		publicKeys: defaultPublicKeys,
 		interval:   defaultInterval,
 		startDelay: defaultStartDelay,
+
+		mExecutedScripts: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "check_script_executed_total",
```
AlekSi, line 125: This way the full metric name will be `pmm_managed_checks_check_script_executed_total`. I think that `checks_check` part is too mouthful. I think `"script_executed_total"` would be enough there.

services/checks/checks.go
```diff
@@ -115,6 +118,20 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
...
+			Help:      "Counter of check scripts executed per service type",
+		}, []string{"service_type"}),
+
+		mGeneratedAlerts: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: prometheusNamespace,
+			Subsystem: prometheusSubsystem,
+			Name:      "generated_alerts_total",
```
AlekSi, line 132: `alerts_generated_total` to match the order of words of the metric above.

Please also consider renaming field names to match metric names in word order.

services/checks/checks.go
```diff
@@ -346,6 +363,9 @@ func (s *Service) executeMySQLChecks(ctx context.Context) ([]string, error) {
 				s.l.Warnf("Failed to process action result: %s.", err)
 				continue
 			}
+			// record metrics
```
AlekSi, line 366:
```
// cat
🐱
```

I think we don't need a comment there :)

But an empty line can improve readability.

services/checks/checks.go
```diff
@@ -346,6 +363,9 @@ func (s *Service) executeMySQLChecks(ctx context.Context) ([]string, error) {
...
+			s.mExecutedScripts.WithLabelValues(string(models.MySQLServiceType)).Inc()
```
AlekSi, line 367: Do we need an explicit `string` conversion there? (I did not check, but probably it will work without explicit conversion)

palash25: I tried that first but they editor showed me red squiggly lines

```
cannot use c.Type (variable of type check.Type) as string value in argument to s.mGeneratedAlerts.WithLabelValues
```

I was surprised that it can't infer that `Type` is actually a string

services/checks/checks.go
```diff
@@ -115,6 +118,20 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
 		publicKeys: defaultPublicKeys,
 		interval:   defaultInterval,
 		startDelay: defaultStartDelay,
+
+		mExecutedScripts: prometheus.NewCounterVec(prometheus.CounterOpts{
```
BupycHuk, line 122: It would be good to have the same order of words that we have in `Name`

palash25:
@BupycHuk @AlekSi please take another look at this

services/checks/checks_test.go
```diff
@@ -117,6 +118,33 @@ func TestCollectChecks(t *testing.T) {
...
+// A proper unit test could not be written due
+// to problems with the code responsible for locating agents
+// Once it is fixed rewrite this test to actually run `executeChecks`
+// method and test for recorded metrics.
+func TestSTTMetrics(t *testing.T) {
+	t.Run("check for recorded metrics", func(t *testing.T) {
+		s := New(nil, nil, nil)
+		expected := strings.NewReader(`
+			# HELP pmm_managed_checks_alerts_generated_total Counter of alerts generated per service type per check type
+        	# TYPE pmm_managed_checks_alerts_generated_total counter
```
palash25, line 130: indentation seems to be messed up on Github, it looks fine on my editor though

AlekSi: I guess you mixed tabs and spaces there. Please fix that

palash25: done @AlekSi

services/checks/checks.go
```diff
@@ -131,6 +147,17 @@ func New(agentsRegistry agentsRegistry, alertsRegistry alertRegistry, db *reform
 		s.startDelay = 0
 	}
 
+	s.mScriptsExecuted.WithLabelValues(string(models.MySQLServiceType))
+	s.mScriptsExecuted.WithLabelValues(string(models.PostgreSQLServiceType))
+	s.mScriptsExecuted.WithLabelValues(string(models.MongoDBServiceType))
+	s.mAlertsGenerated.WithLabelValues(string(models.MySQLServiceType), string(check.MySQLShow))
```
BupycHuk, line 153: shouldn't we increment values for each check type when we run them?

services/checks/checks_test.go
palash25: done

AlekSi:
> palash25 force-pushed the SAAS-104-Prometheus-Metrics-STT  branch from 9bdfb7d to 903add2 yesterday

@palash25 there is not really need to squash all commits and force-push them – we squash/merge them all with "Squash and merge" button on GitHub

services/checks/checks.go
AlekSi, line 153: Let's insert a blank line there when we change metric

AlekSi:
@BupycHuk PTAL

palash25:
> @palash25 there is not really need to squash all commits and force-push them – we squash/merge them all with "Squash and merge" button on GitHub

I didn't know that I usually amend my commits, will keep that in mind from the next time, thanks.

services/checks/checks.go
palash25: done in this commit https://github.com/percona/pmm-managed/pull/407/commits/487a1ec497c921d94a561a2383252470bf3cb251  cc @AlekSi

PR #412, opened 2020-06-25
PMM-125 add ends at support to regisrty

artemgavrilov:
Requires tests fixing

### DIFF
--- a/services/alertmanager/alertmanager.go
+++ b/services/alertmanager/alertmanager.go
@@ -40,22 +40,37 @@ const (
 	path                = "/srv/alertmanager/alertmanager.base.yml"
 )
 
-var resendInterval = 30 * time.Second
+// Environment variable to overwrite resend interval during testing
+const envResendInterval = "PERCONA_TEST_ALERTMANAGER_RESEND_INTERVAL"
+
+const defaultResendInterval = 30 * time.Second
 
 // Service is responsible for interactions with Alertmanager.
 type Service struct {
 	db *reform.DB
 	r  *Registry
 	l  *logrus.Entry
+
+	resendInterval time.Duration
 }
 
 // New creates new service.
 func New(db *reform.DB, alertsRegistry *Registry) *Service {
-	return &Service{
-		db: db,
-		r:  alertsRegistry,
-		l:  logrus.WithField("component", "alertmanager"),
+	l := logrus.WithField("component", "alertmanager")
+
+	s := &Service{
+		db:             db,
+		r:              alertsRegistry,
+		l:              l,
+		resendInterval: defaultResendInterval,
 	}
+
+	if d, err := time.ParseDuration(os.Getenv(envResendInterval)); err == nil && d > 0 {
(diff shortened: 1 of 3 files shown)


services/alertmanager/alertmanager.go
```diff
@@ -40,22 +40,37 @@ const (
...
+		db:             db,
+		r:              alertsRegistry,
+		l:              l,
+		resendInterval: defaultResendInterval,
 	}
+
+	if d, err := time.ParseDuration(os.Getenv(envResendInterval)); err == nil && d > 0 {
+		s.resendInterval = d
+		l.Warnf("Resend interval changed to %s.", d)
+	}
```
artemgavrilov, line 71: I think we want a resend interval update to affect both, alertmanager and registry.

services/alertmanager/registry_test.go
```diff
@@ -55,7 +55,7 @@ func TestRegistry(t *testing.T) {
 
 		expected := &ammodels.PostableAlert{
 			Annotations: annotations,
-			EndsAt:      strfmt.DateTime(nowValueCopy.Add(resolveTimeoutFactor * resendInterval)),
+			EndsAt:      strfmt.DateTime(nowValueCopy.Add(resolveTimeoutFactor * defaultResendInterval)),
```
artemgavrilov, line 58: Seems like that right place to set `EndsAt`. We want alerts to live that time after we published them.

services/alertmanager/registry.go
```diff
@@ -115,6 +109,8 @@ func (r *Registry) collect() ammodels.PostableAlerts {
 	now := now()
 	for id, t := range r.times {
 		if t.Before(now) {
+			alert := r.alerts[id]
+			alert.EndsAt = strfmt.DateTime(now.Add(r.alertTTL))
```
artemgavrilov, line 113: Seems like it is the right place for `EndsAt`. We want alert to live that time after we publish it.

services/alertmanager/alertmanager.go
palash25: yup, that is why I moved resendInterval to registry and that way it can be accessed by both

palash25: its in the latest commit that I pushed yesterday

services/alertmanager/registry.go
palash25: i still don't see how this will make the test pass, it will still fail because one of the fields in `time.Time` will differ, I think the problem is with how we use `now`, maybe we should focus on that and somehow try to provide in the same `now` value in the registry code and the test?

services/alertmanager/alertmanager.go
artemgavrilov: Great then)

services/alertmanager/registry.go
artemgavrilov: I don't fully understand the problem. I tests we control value returned by `now()` so we control time used by tests and registry. Take a look at the last commit.

PR #417, opened 2020-07-07
PMM-5812 Fix relabeling

AlekSi:


BupycHuk:
Can we make a test for this fix?

AlekSi:
It was done at https://github.com/percona/promconfig/pull/3: the fix itself is just two brackets; and then there are 200 lines of tests.

PR #419, opened 2020-07-14
SAAS-236 Move Registry from Alertmanager to Checks service

palash25:
https://jira.percona.com/browse/SAAS-236

[FB](https://github.com/Percona-Lab/pmm-submodules/pull/988)


services/checks/checks_test.go
```diff
@@ -215,10 +214,7 @@ func TestStartChecks(t *testing.T) {
 			require.NoError(t, sqlDB.Close())
 		}()
 
-		var ar mockAlertRegistry
-		ar.On("RemovePrefix", mock.Anything, mock.Anything).Return()
-
-		s := New(nil, &ar, db)
+		s := New(nil, nil, db)
```
AlekSi, line 217: @palash25 here you pass `nil`

palash25: thanks

services/checks/checks.go
```diff
@@ -211,6 +213,23 @@ func (s *Service) StartChecks(ctx context.Context) error {
 
 	s.collectChecks(nCtx)
 	s.executeChecks(nCtx)
+
+	go func() {
+		t := time.NewTicker(s.alertsRegistry.resendInterval)
```
BupycHuk, line 218: Shouldnt it be somewhere where we send alerts? Here or in alertmanagerService?

services/checks/checks.go
```diff
@@ -308,26 +327,32 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
 
 	postgreSQLAlertsIDs := s.executePostgreSQLChecks(ctx)
-	alertsIDs = append(alertsIDs, postgreSQLAlertsIDs...)
+	checkResults = append(checkResults, postgreSQLAlertsIDs...)
 
 	mongoDBAlertsIDs := s.executeMongoDBChecks(ctx)
-	alertsIDs = append(alertsIDs, mongoDBAlertsIDs...)
+	checkResults = append(checkResults, mongoDBAlertsIDs...)
+
+	var alertsIDs []string
```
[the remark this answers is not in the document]
BupycHuk: Looks like we can preallocate memory here

services/checks/checks_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/percona/pmm/version"
 	promtest "github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
+	mock "github.com/stretchr/testify/mock"
```
BupycHuk, line 31: Any reason to make this?

palash25: I actually don't remember making that alias, maybe vscode did it :thinking:

services/checks/checks.go
palash25: actually we wanted to move the responsibility of calling `alertmanager.SendAlerts` to its caller (see [this](https://github.com/percona/pmm-managed/pull/416/files#diff-a8171d88a4e24ab2acd4b471b52230faR133-R135)) and since we want to keep resending alerts at an interval I moved the ticker here

The last 2 points in the ticket's [description](https://jira.percona.com/browse/SAAS-236) also mention that SendAlerts call to be moved to StartChecks method and called in a separate goroutine

services/checks/registry.go
```diff
@@ -91,8 +91,8 @@ func (r *Registry) CreateAlert(id string, labels, annotations map[string]string,
 	}
 }
 
-// RemovePrefix removes all alerts with given ID prefix except a given list of IDs.
-func (r *Registry) RemovePrefix(prefix string, keepIDs map[string]struct{}) {
+// removePrefix removes all alerts with given ID prefix except a given list of IDs.
+func (r *registry) removePrefix(prefix string, keepIDs map[string]struct{}) {
```
artemgavrilov, line 95: Looks like that function can be refactored. It will process only STT alerts now, so `prefix` always the same. And probably we can wipe all alerts from registry each iteration and add new as there are no alerts form other components in that registry.

services/checks/checks.go
```diff
@@ -308,26 +327,32 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
 func (s *Service) executeChecks(ctx context.Context) {
 	s.l.Info("Executing checks...")
 
-	var alertsIDs []string
+	var checkResults []sttCheckResult
 
 	mySQLAlertsIDs := s.executeMySQLChecks(ctx)
```
artemgavrilov, line 332: `mySQLAlertsIDs` not IDs anymore.

services/checks/checks.go
```diff
@@ -308,26 +327,32 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
-	alertsIDs = append(alertsIDs, mySQLAlertsIDs...)
+	checkResults = append(checkResults, mySQLAlertsIDs...)
 
 	postgreSQLAlertsIDs := s.executePostgreSQLChecks(ctx)
```
artemgavrilov, line 335: Same as above

services/checks/checks.go
```diff
@@ -308,26 +327,32 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
-	alertsIDs = append(alertsIDs, postgreSQLAlertsIDs...)
+	checkResults = append(checkResults, postgreSQLAlertsIDs...)
 
 	mongoDBAlertsIDs := s.executeMongoDBChecks(ctx)
```
artemgavrilov, line 338: Same as above

services/checks/checks_test.go
palash25: removed it

services/checks/registry.go
palash25: so I guess we should call this something like `flushAlerts` now? @artemgavrilov

artemgavrilov: I'm not sure, `flush` typically means "send everything from buffer" or something like that. Here we clean the registry. 

I think we can call that method `clean()` and invoke it before https://github.com/percona/pmm-managed/blob/cd1aa44a579ed32a25c136701db08384eb94fc39/services/checks/checks.go#L342-L346

So `clean()` will remove all existing alerts from the registry, and then we will fill the registry with newly received alerts from STT.

palash25: done @artemgavrilov

services/checks/registry.go
```diff
@@ -91,24 +90,19 @@ func (r *Registry) CreateAlert(id string, labels, annotations map[string]string,
...
-		if _, ok := keepIDs[id]; ok {
-			continue
-		}
-		if strings.HasPrefix(id, prefix) {
-			delete(r.alerts, id)
-			delete(r.times, id)
-		}
+		delete(r.alerts, id)
+		delete(r.times, id)
 	}
```
artemgavrilov, line 101: It can be simply
```
r.alerts = make(map[string]ammodels.PostableAlert)
r.times = make(map[string]time.Time)
```

services/checks/registry.go
```diff
@@ -91,24 +90,19 @@ func (r *Registry) CreateAlert(id string, labels, annotations map[string]string,
...
 }
 
-// RemovePrefix removes all alerts with given ID prefix except a given list of IDs.
-func (r *Registry) RemovePrefix(prefix string, keepIDs map[string]struct{}) {
+// removePrefix removes all alerts with given ID prefix except a given list of IDs.
```
artemgavrilov, line 93: Wrong description

services/checks/registry.go
```diff
@@ -67,18 +66,18 @@ func NewRegistry() *Registry {
 	}
 }
 
-// CreateAlert creates alert from given AlertParams and adds or replaces alert with given ID in registry.
+// createAlert creates alert from given AlertParams and adds or replaces alert with given ID in registry.
 // If that ID wasn't present before, alert is added in the pending state. It we be transitioned to the firing
 // state after delayFor interval. This is similar to `for` field of Prometheus alerting rule:
 // https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/
-func (r *Registry) CreateAlert(id string, labels, annotations map[string]string, delayFor time.Duration) {
+func (r *registry) createAlert(id string, labels, annotations map[string]string, delayFor time.Duration) {
```
artemgavrilov, line 73: I wonder, do we need delay for STT alerts at all? 🤔

services/checks/registry.go
palash25: forgot to edit that thanks :+1:

services/checks/registry.go
AlekSi: Funny enough, for/range/delete loop is optimised by the compiler: https://golang.org/doc/go1.11#performance-compiler
(I do not imply that we need that optimisation there, that's JFYI)

services/alertmanager/alertmanager.go
```diff
@@ -37,21 +37,21 @@ import (
 const (
 	alertmanagerDataDir = "/srv/alertmanager/data"
 	prometheusDir       = "/srv/prometheus"
-	path                = "/srv/alertmanager/alertmanager.base.yml"
+	dirPerm             = os.FileMode(0775)
+
+	path = "/srv/alertmanager/alertmanager.base.yml"
```
AlekSi, line 42: That short name was originally a local constant. I think we should find a better (slightly) longer name for package-level constant.

services/alertmanager/alertmanager.go
```diff
@@ -144,9 +132,8 @@ receivers:
 	}
 }
 
-// sendAlerts sends alerts collected in the Registry.
-func (svc *Service) sendAlerts(ctx context.Context) {
-	alerts := svc.r.collect()
+// SendAlerts sends given alerts
```
AlekSi, line 135: I think we should add something like "it is caller's responsibility to call this method every now and then". We change a logic there, and we need slightly more comments to change our mental model.

services/checks/checks.go
AlekSi: That code is not correct. `StartChecks` is being called from the API. Two calls means that we create two goroutines that never finish. 1000 calls – 1000 goroutines. But we should have one goroutine for that.

Last two points in description mean that this goroutine should be in a separate method, and there we should be able trigger another send loop manually, without waiting 24 hours.

services/checks/checks.go
```diff
@@ -470,22 +491,29 @@ func (s *Service) executeMongoDBChecks(ctx context.Context) []string {
...
 
 	return res
 }
 
-func (s *Service) processResults(ctx context.Context, check check.Check, target target, resID string) ([]string, error) {
+type sttCheckResult struct {
+	checkName string
+	target    target
+	result    check.Result
+	alertID   string
```
AlekSi, line 513: It should not be there

services/checks/registry.go
AlekSi: Yes, let's get rid of delayFor completely – it was over-engineered, and it is clear now that our future "dynamic alerting" feature will work differently

services/alertmanager/alertmanager.go
palash25: I left that part out because we have exported the method so I thought that it will be understood that this method will have to be called by a caller outside of this package, seemed a bit redundant to me

will add the comment in the next commit

services/checks/checks.go
palash25: but I need alertID later on to pass in the `createAlert` call

AlekSi: Yes. What I'm saying is that one point of our refactoring is to decouple check results from alerts. Therefore, `alertID` does not belong to `sttCheckResult`. I think it can be computed later?

services/checks/checks.go
palash25: > Last two points in description mean that this goroutine should be in a separate method, and there we should be able trigger another send loop manually, without waiting 24 hours.

I have moved the goroutine to a separate method, but one thing I can't understand is "trigger another send loop manually" which method will call the separate method?

AlekSi: First of all, method `resendAlerts` currently does nothing :)

services/checks/checks.go
palash25: got it

services/checks/registry.go
```diff
@@ -87,28 +83,21 @@ func (r *Registry) CreateAlert(id string, labels, annotations map[string]string,
...
 }
 
-// RemovePrefix removes all alerts with given ID prefix except a given list of IDs.
-func (r *Registry) RemovePrefix(prefix string, keepIDs map[string]struct{}) {
+// clean clears all alerts from the registry
+func (r *registry) clean() {
 	r.rw.Lock()
 	defer r.rw.Unlock()
+	r.alerts = make(map[string]ammodels.PostableAlert)
+	r.times = make(map[string]time.Time)
```
artemgavrilov, line 95: You don't need that `times` map without delays at all. In `collect` method you can return all alerts that you have.

services/checks/checks.go
palash25: have updated the code, `resendAlerts` is now called from `Run` after starting the checks so one resend loop will always be running now what I don't understand is what you mean by manually triggering the loop, we don't want the loop in StartChecks so I am assuming that we don't want the user to be manually triggering the loop so which part of our code will trigger it and why?

AlekSi: > First of all, method resendAlerts currently does nothing :)

That's literally it:

```go
// resendAlerts can be used for manually triggerring a SendAlerts loop
func (s *Service) resendAlerts(ctx context.Context) {
	go func() {
		t := time.NewTicker(s.alertsRegistry.resendInterval)
		defer t.Stop()

		for {
			select {
			case <-ctx.Done():
				return
			case <-t.C:
				// nothing, continue for loop
			}
		}
	}()
}
```

All that code does is creates a goroutine that does nothing.

Once it starts doing _something_ (like actually sending alerts), I think other things will fall into places.

palash25: not sure how that happened, I remember having a call to SendAlerts there and pushing the code  :confused:

anyway I have added it and pushed it now, sorry for bothering you without properly updating the code @AlekSi

services/checks/registry.go
palash25: done

services/checks/checks.go
```diff
@@ -308,26 +351,28 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
-	alertsIDs = append(alertsIDs, mongoDBAlertsIDs...)
+	mongoDBCheckResults := s.executeMongoDBChecks(ctx)
+	checkResults = append(checkResults, mongoDBCheckResults...)
 
-	// removing old STT alerts except created during current run
-	s.alertsRegistry.RemovePrefix(alertsPrefix, sliceToSet(alertsIDs))
+	s.alertsRegistry.clean()
+	for _, result := range checkResults {
+		s.createAlert(result.checkName, result.target, &result.result)
+	}
```
AlekSi, line 368: Ticket specifically says "Replace checks stored in the registry in executeChecks method **atomically**." StartChecks API can be called concurrently, and we will have a race here.

I think that complexity should be in registry: remove `clear()` method, add `set(alerts ammodels.PostableAlerts)` method that replaces alerts

services/checks/checks.go
```diff
@@ -172,8 +177,44 @@ func (s *Service) Run(ctx context.Context) {
...
+		defer wg.Done()
+		s.restartChecks(ctx)
+	}()
+
+	wg.Wait()
+}
+
+// resendAlerts resends collected alerts until ctx is canceled.
+func (s *Service) resendAlerts(ctx context.Context) {
+	t := time.NewTicker(s.alertsRegistry.resendInterval)
```
AlekSi, line 199: That (and TTL) should be a property of the service, not registry

services/checks/checks.go
palash25: > Ticket specifically says "Replace checks stored in the registry in executeChecks method atomically." StartChecks API can be called concurrently, and we will have a race here.

registry uses a mutex before mutating the alert map won't that be enough to avoid the race?

AlekSi: Registry uses mutex for each individual operation. That's not enough. Consider goroutine 1 being on that line, and goroutine 2 calling `clean()` on line 365. There no data race that Go race detector can detect, but the registry is empty and data is lost.

artemgavrilov: But that situation covered by alert TTL that three times longer than resend interval. Technically on UI such situation will be unnoticed, isn't it?

AlekSi: Yeah, until we add API to get current check results (not alerts) :)

artemgavrilov: But then "it is a completely different story" )

Sorry @palash25 for that solution

AlekSi: Yes, we have that different story. I'm not saying that we should implement that API now. But we also should not make something that will make it harder in the future. The biggest reason for that ticket is to have a registry of check results, not alerts – exactly because we will add APIs later.

I think this part can be reworked something like that:
1. Make `createAlert` a function returning alert object.
2. Replace both `clean` and `createAlert` methods with `set(alerts []portableAlert)` method that sets the whole slice.
3. `collect()` will still return a copy of that slice.

services/checks/checks.go
palash25: I did that in the latest push but for registry unit tests I can no longer use `resendInterval` and had to use the constant `defaultResendInterval`

palash25:
New FB PR  https://github.com/Percona-Lab/pmm-submodules/pull/983

services/checks/checks.go
```diff
@@ -532,7 +604,12 @@ func hashID(s string) string {
 	return hex.EncodeToString(data[:])
 }
 
-func (s *Service) createAlert(id, name string, target target, result *check.Result) {
+type alertWithID struct {
+	alert ammodels.PostableAlert
+	id    string
+}
+
+func (s *Service) createAlert(name string, target target, result *check.Result) alertWithID {
```
BupycHuk, line 612: maybe it's better to rename this method to `prepareAlert` or something like that?

services/checks/registry.go
```diff
@@ -0,0 +1,76 @@
...
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/percona/pmm/api/alertmanager/ammodels"
+)
+
+// registry stores alerts and delay information by IDs.
+type registry struct {
+	rw     sync.RWMutex
+	alerts map[string]ammodels.PostableAlert
```
BupycHuk, line 30: why do we still use map here? Looks like it's unnecessary anymore

services/checks/registry.go
```diff
@@ -0,0 +1,76 @@
...
+			// GeneratorURL: "TODO",
+			Labels: labels,
+		},
+		EndsAt:      strfmt.DateTime(r.nowF().Add(alertTTL)),
+		Annotations: annotations,
+	}
+}
+
+// set clears the previous alerts and sets a new slice of alerts in the registry
+func (r *registry) set(alerts []alertWithID) {
```
BupycHuk, line 55: Is there any possibility to have two or more alerts with the same alert id in this slice?

services/checks/registry.go
palash25: right, will change it to a slice in the next push

services/checks/checks.go
palash25: hmm.. I'm okay with either of those names and personally I don't find `createAlert` to be ambiguous enough to rename it, but if you feel otherwise I am okay with renaming the method

services/checks/registry.go
palash25: the id is calculated like this `id := alertsPrefix + hashID(target.serviceID+"/"+result.Summary)` so if the service.ID and result.Summary  are the same then yes.

the summary can be same in case the user writes a check that sends lets say 2 alerts with the same summary and if these alerts come from any one target then yes I think they will have the same id

```
script: |
          def check(rows):
              results = []
    
              results.append({
                  "summary": "Fake check",
                  "description": "failed check .....",
                  "severity": "warning",
              })
              results.append({
                  "summary": "Fake check",
                  "description": "failed check .....",
                  "severity": "warning",
              })
    
              return results
```

i dont think anyone would write such checks but I guess it can be possible

AlekSi: 7fd812edb035b08d0d7518ebced375f1000f021b

services/checks/registry.go
```diff
@@ -0,0 +1,76 @@
...
+// newRegistry creates a new registry.
+func newRegistry() *registry {
+	return &registry{
+		alerts: make(map[string]ammodels.PostableAlert),
+		nowF:   time.Now,
+	}
+}
+
+// createAlert creates alert from given AlertParams
+func (r *registry) createAlert(labels, annotations map[string]string, alertTTL time.Duration) ammodels.PostableAlert {
```
AlekSi, line 43: That should be a function, not a method. Just remove receiver

services/checks/registry.go
AlekSi: I guess we can rename `alertWithID` to just `alert` (we don't use that name yet, right?) and store `alerts []alert`.

services/checks/registry.go
```diff
@@ -0,0 +1,76 @@
...
+
+// newRegistry creates a new registry.
+func newRegistry() *registry {
+	return &registry{
+		alerts: make(map[string]ammodels.PostableAlert),
+		nowF:   time.Now,
+	}
+}
+
+// createAlert creates alert from given AlertParams
```
artemgavrilov, line 42: Missing dot

services/checks/registry.go
```diff
@@ -0,0 +1,76 @@
...
+		Alert: ammodels.Alert{
+			// GeneratorURL: "TODO",
+			Labels: labels,
+		},
+		EndsAt:      strfmt.DateTime(r.nowF().Add(alertTTL)),
+		Annotations: annotations,
+	}
+}
+
+// set clears the previous alerts and sets a new slice of alerts in the registry
```
artemgavrilov, line 54: Missing dot

artemgavrilov:
@AlekSi I wonder, do we actually need registry, in fact, it's just a map with mutex. Maybe we can move that map to the checks.go, or better decouple that stuff?

AlekSi:
> Maybe we can move that map to the checks.go, or better decouple that stuff?

Maybe, but how? :)

In SAAS-237 we will add API for that collected data, and ignore the rest of checks.Service logic. Not sure if we want keep registry separate or not…

I think we should wrap this task with registry as a separate type, and maybe re-think that later.

services/checks/registry.go
palash25: if you look further down the method I need access to `nowF` which is a field of registry that is why it is a method. Do you still want me to remove the receiver and add the endsAt field in the collect method? @AlekSi

AlekSi: Nope, I stand corrected

services/checks/registry.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// set clears the previous alerts and sets a new slice of alerts in the registry.
+func (r *registry) set(alerts []alert) {
+	r.rw.Lock()
+	defer r.rw.Unlock()
+
+	alertSet := make([]ammodels.PostableAlert, len(alerts))
+	for _, alert := range alerts {
+		alertSet = append(r.alerts, alert.alert)
```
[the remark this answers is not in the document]
artemgavrilov: @palash25 yep, looks like a bug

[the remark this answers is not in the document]
palash25: fixed it

services/checks/registry.go
```diff
@@ -0,0 +1,78 @@
...
+		Annotations: annotations,
+	}
+}
+
+// set clears the previous alerts and sets a new slice of alerts in the registry.
+func (r *registry) set(alerts []alert) {
+	r.rw.Lock()
+	defer r.rw.Unlock()
+
+	var alertSet []ammodels.PostableAlert
```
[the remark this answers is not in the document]
palash25: prealloc with length `len(alerts)` breaks the unit test

```
time="2020-07-28T13:47:46Z" level=warning msg="Failed to parse pmm-agent version: failed to parse \"invalid\"." component=checks
--- FAIL: TestRegistry (0.00s)
    --- FAIL: TestRegistry/Collect_Alerts (0.00s)
        registry_test.go:56: 
                Error Trace:    registry_test.go:56
                Error:          "[%!s(*ammodels.PostableAlert=&{map[] {0 0 <nil>} {0 0 <nil>} { map[]}}) %!s(*ammodels.PostableAlert=&{map[annotation:test] {13817088745824072076 15862433041 0x17657c0} {0 0 <nil>} { map[label:demo]}})]" should have 1 item(s), but has 2
                Test:           TestRegistry/Collect_Alerts
FAIL
FAIL    github.com/percona/pmm-managed/services/checks  9.865s
```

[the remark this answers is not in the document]
artemgavrilov: Preallocate that slice with zero length, and capacity = len(alerts)

```
alertSet = make([]ammodels.PostableAlert, 0, len(alerts))
```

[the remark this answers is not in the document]
palash25: done thanks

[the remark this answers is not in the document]
artemgavrilov: Or another option is allocate slice with length = len(aletrs), but then you should put elements by index, not to append them:
```
	alertSet := make([]ammodels.PostableAlert, len(alerts))
	for i :=0; i<len(alerts); i++ {
		alertSet[i] = alerts[i].alert
	}
```

palash25:
FB https://github.com/Percona-Lab/pmm-submodules/pull/988

services/checks/checks.go
```diff
@@ -516,23 +587,34 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
...
+	fmt.Fprintf(s, "%s\n", result.Summary)
+	fmt.Fprintf(s, "%s\n", result.Description)
+	fmt.Fprintf(s, "%v\n", result.Severity)
+	return alertsPrefix + hex.EncodeToString(s.Sum(nil))
 }
 
-func (s *Service) createAlert(id, name string, target target, result *check.Result) {
+type alert struct {
+	alert ammodels.PostableAlert
+	id    string
```
AlekSi, line 614: We don't actually use that field anywhere anymore (we calculate it and add to labels just below), so that type can be replaced by `ammodels.PostableAlert` directly

services/checks/registry.go
```diff
@@ -0,0 +1,78 @@
...
+
+// set clears the previous alerts and sets a new slice of alerts in the registry.
+func (r *registry) set(alerts []alert) {
+	r.rw.Lock()
+	defer r.rw.Unlock()
+
+	alertSet := make([]ammodels.PostableAlert, 0, len(alerts))
+	for _, alert := range alerts {
+		alertSet = append(alertSet, alert.alert)
+	}
```
AlekSi, line 62: https://golang.org/pkg/builtin/#copy

services/checks/checks.go
```diff
@@ -308,26 +370,29 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
-	mongoDBAlertsIDs := s.executeMongoDBChecks(ctx)
-	alertsIDs = append(alertsIDs, mongoDBAlertsIDs...)
+	mongoDBCheckResults := s.executeMongoDBChecks(ctx)
+	checkResults = append(checkResults, mongoDBCheckResults...)
 
-	// removing old STT alerts except created during current run
-	s.alertsRegistry.RemovePrefix(alertsPrefix, sliceToSet(alertsIDs))
+	alerts := make([]alert, len(checkResults))
+	for _, result := range checkResults {
+		alerts = append(alerts, s.createAlert(result.checkName, &result.target, &result.result))
```
AlekSi, line 386: Here we append to a slice of non-zero length, so the end length will be `len(checkResults) * 2` and first `len(checkResults)` elements will be empty

AlekSi:
@palash25 `set` should accept a slice of check results; `collect` should return alerts with updated EndsAt

services/checks/checks.go
```diff
@@ -19,9 +19,7 @@ package checks
 
 import (
 	"bytes"
-	"context"
-	"crypto/sha1" //nolint:gosec
-	"encoding/hex"
+	"context" //nolint:gosec
```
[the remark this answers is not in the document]
artemgavrilov: Yep, it should be on "crypto/sha1"

services/checks/checks.go
```diff
@@ -104,19 +109,30 @@ type Service struct {
...
+		db:                  db,
+		alertsRegistry:      newRegistry(),
+
+		l:              l,
+		host:           defaultHost,
+		publicKeys:     defaultPublicKeys,
+		interval:       defaultInterval,
+		startDelay:     defaultStartDelay,
+		resendInterval: resendInterval,
+		alertTTL:       resolveTimeoutFactor * resendInterval,
```
artemgavrilov, line 135: I think better to pass that value to registry constructor on line 127.

PR #420, opened 2020-07-14
SAAS-175 Auth API

artemgavrilov:
- https://jira.percona.com/browse/SAAS-175

- [ ] https://github.com/percona/pmm/pull/553

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/967

Integration with Percona Auth Service, added three server API methods SignUp and SingIn accessible only for grafana admin.


Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "SAAS-175-auth-api"
```
artemgavrilov, line 19: TODO

artemgavrilov:
Feature build in progress

services/auth/auth_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	defer func() {
+		require.NoError(t, sqlDB.Close())
+	}()
+
+	s := New(db)
+	s.host = devAuthHost
+
+	login := gofakeit.Email()
+	password := "Password12345"
```
palash25, line 47: can't we use gofakeit to generate the passwords too? just for consistency

services/auth/auth.go
```diff
@@ -0,0 +1,161 @@
...
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to save session id")
+	}
+
+	return nil
+}
+
+// RefreshSession resets session timeout.
+func (s *Service) RefreshSession(ctx context.Context) error {
```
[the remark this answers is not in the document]
AlekSi: Yeah, we should add method `Run(ctx)` that will call that method every day

services/auth/auth.go
```diff
@@ -0,0 +1,161 @@
...
+		return errors.New("no active sessions")
+	}
+
+	cc, err := s.getConnection(ctx)
+	if err != nil {
+		return errors.Wrap(err, "failed establish connection with Percona")
+	}
+	defer cc.Close() //nolint:errcheck
+
+	md := metadata.New(map[string]string{"authorization": "PP-v1beta1 " + settings.SessionID})
```
AlekSi, line 127: Let's move "PP-v1beta1" to the named constant

services/auth/auth_test.go
```diff
@@ -0,0 +1,72 @@
...
+	db := reform.NewDB(sqlDB, postgresql.Dialect, nil)
+
+	defer func() {
+		require.NoError(t, sqlDB.Close())
+	}()
+
+	s := New(db)
+	s.host = devAuthHost
+
+	login := gofakeit.Email()
```
AlekSi, line 46: I think we already use some other fake data package in pmm-managed

models/settings.go
```diff
@@ -54,6 +54,8 @@ type Settings struct {
 		// Security Threat Tool enabled
 		STTEnabled bool `json:"stt_enabled"`
 	} `json:"sass"`
```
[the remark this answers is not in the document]
artemgavrilov: Done

models/settings.go
```diff
@@ -54,6 +54,8 @@ type Settings struct {
...
+
+	SessionID string `json:"session_id"`
```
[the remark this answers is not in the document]
artemgavrilov: Done

services/auth/auth.go
```diff
@@ -0,0 +1,161 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// Package auth provides authentication/authorization functionality.
+package auth
```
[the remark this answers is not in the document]
artemgavrilov: Renamed

services/auth/auth.go
artemgavrilov: Done

services/auth/auth_test.go
artemgavrilov: Done

services/auth/auth_test.go
artemgavrilov: I didn't find anything like that in dependencies 🤷‍♂️

services/auth/auth.go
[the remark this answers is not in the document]
artemgavrilov: Done

models/settings.go
```diff
@@ -51,6 +51,10 @@ type Settings struct {
 
 	// Saas config options
 	SaaS struct {
+		// Platform user email
+		Email string `json:"email"`
+		// Percona platform session Id
```
AlekSi, line 56: Let's stick to "Percona Platform" in comments/docs, not "Percona platform" or just "Platform" as above

models/settings_helpers.go
```diff
@@ -129,6 +134,14 @@ func UpdateSettings(q reform.DBTX, params *ChangeSettingsParams) (*Settings, err
 		settings.SaaS.STTEnabled = true
 	}
 
+	if params.SessionID != "" {
+		settings.SaaS.SessionID = params.SessionID
+	}
+
+	if params.Email != "" {
+		settings.SaaS.Email = params.Email
+	}
```
AlekSi, line 143: How that code will handle sign-out? I guess we will have to add boolean flag like `RemoveAlertManagerURL`, but can we do something better?

services/platform/platform.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/percona/pmm-managed/models"
+)
+
+const (
+	defaultHost            = "check.percona.com:443"
+	defaultRefreshInterval = 24 * time.Hour
+	timeout                = 5 * time.Second
+
+	envHost            = "PERCONA_TEST_AUTH_HOST"
+	envRefreshInterval = "PERCONA_TEST_REFRESH_INTERVAL"
```
AlekSi, line 45: `PERCONA_TEST_AUTH_REFRESH_INTERVAL`. The current name is too generic

services/platform/platform.go
```diff
@@ -0,0 +1,200 @@
...
+
+	authType = "PP-v1beta1" // TODO Change to PP-1 after auth API release
+
+)
+
+var ErrNoActiveSessions = errors.New("no active sessions") //nolint:golint
```
AlekSi, line 51: golint hints that this variable should not be exported

services/platform/platform.go
```diff
@@ -0,0 +1,200 @@
...
+
+// Run refreshes Percona Platform session every interval until context is canceled.
+func (s *Service) Run(ctx context.Context) {
+	ticker := time.NewTicker(s.refreshInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+		// continue with next loop iteration
```
AlekSi, line 93:
```suggestion
			// continue with next loop iteration
```

services/platform/platform.go
```diff
@@ -0,0 +1,200 @@
...
+		case <-ctx.Done():
+			return
+		}
+
+		err := s.RefreshSession(ctx)
+		if err != nil {
+			s.l.Warn(err)
+		}
+		if err != nil && err != ErrNoActiveSessions {
```
AlekSi, line 102: Two checks for `err != nil`, so we can log the same error twice?

services/server/deps.go
```diff
@@ -86,3 +86,8 @@ type supervisordService interface {
 type telemetryService interface {
 	DistributionMethod() serverpb.DistributionMethod
 }
+
+type platformService interface {
```
AlekSi, line 90: Please add a comment

services/platform/platform.go
artemgavrilov: Golint asks to add comments to exported variables or make them private. IMHO it's nice to have exported errors, however, comment will repeat the variable name, it's descriptive enough I think.

models/settings_helpers.go
artemgavrilov: We haven't signout mechanics for now. It requires also session termination before deleting sessionID in managed.
There should be another ticket created for that, I think.

Boolean flag looks ok, at least I don't see something more simple for now. It could be "LogOut bool", and when it's true we remove `Email` and `SessionID`.

services/platform/platform.go
artemgavrilov: Fixed

services/platform/platform.go
artemgavrilov: Changed to `PERCONA_TEST_SESSION_REFRESH_INTERVAL`, long but precise.

services/platform/platform.go
artemgavrilov: Yep, it's a mistake. Removed.

services/server/deps.go
artemgavrilov: Done.

services/platform/platform.go
AlekSi: > IMHO it's nice to have exported errors

Why? We are not using it in other packages

models/settings.go
artemgavrilov: Fixed.

services/platform/platform.go
artemgavrilov: Maybe I think too generic, but someday someone will see that error as package contract... nevermind. Will make it private)

artemgavrilov: Done

models/settings_helpers.go
artemgavrilov: Add such filed now, or leave it to another ticket?

AlekSi: The intention was not do anything that will make future sign-out functionality worse / harder.
If boolean flag is the best we can think of, so be it

artemgavrilov: Done

services/platform/platform.go
```diff
@@ -0,0 +1,197 @@
...
+
+	for {
+		select {
+		case <-ticker.C:
+			// continue with next loop iteration
+		case <-ctx.Done():
+			return
+		}
+
+		err := s.RefreshSession(ctx)
```
BupycHuk, line 98: What's a ttl of session and what will happen if pmm-server doesn't have internet connection when it tries to refresh session?

artemgavrilov: Session TTL configured on SaaS side, now it's 7 days. 

Added context with a timeout to prevent problems with hanged requests 👍

services/platform/platform.go
```diff
@@ -0,0 +1,199 @@
...
+	opts := []grpc.DialOption{
+		// replacement is marked as experimental
+		grpc.WithBackoffMaxDelay(timeout), //nolint:staticcheck
+
+		grpc.WithBlock(),
+		grpc.WithUserAgent("pmm-managed/" + version.Version),
+		grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)),
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
```
AlekSi, line 191: @artemgavrilov We can remove that one then

AlekSi: Actually, no.
We should have two different timeout: dial timeout (5 seconds) and total timeout (like 10 seconds). We should pass the second timeout to all Auth Service calls, not only to refresh, but also to SignIn/SignUp

artemgavrilov: Hm, ok. But shouldn't that timeout be controlled by managed API client? I mean if UI creates request with timeout 30s why we limit it 10s?

AlekSi: UI doesn't actually control timeout – we don't document `Grpc-Timeout` header in our exposed JSON API (and I don't think we ever tested it). Timeout in Go code will allow UI to get "good" error code instead of generic 504 from nginx.

AlekSi: And we _almost_ can set timeout in interceptor – except that we have several special cases where we need higher timeout: logs.zip, pmm-agent protocol, some QAN APIs, etc. So for now we stuck with per-method or per-service timeouts.

artemgavrilov: Ok, thanks for explanation. Will add timeouts 👌

artemgavrilov: @AlekSi take a look at the last commit, please. I think timeouts should be defined on the caller side, not inside methods, so I defined them in `server.go`

services/platform/platform.go
```diff
@@ -0,0 +1,199 @@
...
+	"google.golang.org/grpc/metadata"
+	"gopkg.in/reform.v1"
+
+	"github.com/percona/pmm-managed/models"
+)
+
+const (
+	defaultHost                   = "check.percona.com:443"
+	defaultSessionRefreshInterval = 24 * time.Hour
+	timeout                       = 5 * time.Second
```
AlekSi, line 42: Let's rename it to `dialTimeout`

services/platform/platform.go
```diff
@@ -0,0 +1,199 @@
...
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to save session id")
+	}
+
+	return nil
+}
+
+// RefreshSession resets session timeout.
+func (s *Service) RefreshSession(ctx context.Context) error {
```
AlekSi, line 149: btw, let's un-export this method

services/server/server.go
```diff
@@ -50,6 +50,8 @@ import (
 	"github.com/percona/pmm-managed/utils/envvars"
 )
 
+const timeout = 10 * time.Second
```
AlekSi, line 53: Let's use a better name. `platformAPITimeout`?

services/platform/platform.go
artemgavrilov: Done

services/platform/platform.go
artemgavrilov: It's also used as `BackoffMaxDelay`, but it's related to the connection. Renamed.

AlekSi: Yeah. May be we can also rename getConnection to `dial` or something to have the same name/prefix

services/server/server.go
artemgavrilov: Seems that `platformAPITimeout` is good enough. Renamed.

services/platform/platform.go
artemgavrilov: `getConnection` is a more common name than `dial`, but the second is specific for gRPC. Oh, I don't know what is better. Let's try: last commit.

artemgavrilov:
Something strange happening with CI.

PR #447, opened 2020-08-28
PMM-6389 Add VictoriaMetrics

ctrlok:
- [ ] https://github.com/Percona-Lab/percona-images/pull/86
- [ ] https://github.com/percona/pmm-update/pull/96
- [ ] https://github.com/percona/pmm-managed/pull/447

https://github.com/Percona-Lab/pmm-submodules/pull/1020


services/agents/registry.go
```diff
@@ -62,9 +62,9 @@ type pmmAgentInfo struct {
 //
 // TODO Split into several types https://jira.percona.com/browse/PMM-4932
 type Registry struct {
-	db         *reform.DB
-	prometheus prometheusService
-	qanClient  qanClient
+	db             *reform.DB
+	scrapeServices []prometheusService
```
BupycHuk, line 66: isn't it better to create some proxy service which will pass request to prometheus service or victoria metrics service or both based on environment variables?

services/agents/registry.go
```diff
@@ -392,8 +392,9 @@ func (r *Registry) stateChanged(ctx context.Context, req *agentpb.StateChangedRe
 	if e != nil {
 		return e
 	}
-
-	r.prometheus.RequestConfigurationUpdate()
+	for _, svc := range r.scrapeServices {
```
BupycHuk, line 395: So we will not have to write these loops everywhere.

services/agents/registry.go
BupycHuk: And then remove prometheus will be easier.

services/prometheus/scrape_configs.go
```diff
@@ -474,3 +477,168 @@ func scrapeConfigsForExternalExporter(s *models.MetricsResolutions, params *scra
 
 	return []*config.ScrapeConfig{cfg}, nil
 }
+
+// PopulateScrapeConfigs populates scrape configs for basic running service and agents.
+func PopulateScrapeConfigs(cfg *config.Config, l *logrus.Entry, q *reform.Querier, s *models.MetricsResolutions) error {
```
BupycHuk, line 482: Is it a good idea to make it public? Maybe it's better to create separate structure which will handle all scraper_configs and pass it, so victoriaMetrics packaged will not use prometheus package. or move this file somewhere else.
It should help us with removing prometheus once victoria will be ready.

BupycHuk (changes requested):
CI Failed

services/supervisord/supervisord.go
```diff
@@ -558,3 +570,28 @@ stdout_logfile_backups = 3
...
+priority = 7
+command =
+	/usr/sbin/victoria-metrics
+		--promscrape.config=/etc/victoriametrics-promscrape.yml
+		--retentionPeriod={{ .DataRetentionMonth }}
+		--storageDataPath=/srv/victoriametrics/data
+		--httpListenAddr=127.0.0.1:8428
+user = pmm
+autorestart = true
+autostart = true
```
AlekSi, line 587: We should render that file in all cases, but that value, and one above, should be `false` if `isVMEnabled` is false

AlekSi (changes requested):

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,333 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/services/prometheus"
+)
+
+const updateBatchDelay = 3 * time.Second
+
+var (
+	enabled       = os.Getenv("PERCONA_TEST_VM") == "1"
```
AlekSi, line 48: And what if it is `"true"`?

services/agents/registry.go
tenmozes: we don't need to do orchestration for removing of prometheus, i think this is not a hard work. 
Please, also take into account that `vmalert` will be added and it has similar interface and should be reload by when config changes, as right now, this is one call (change settings) we need to ask every service about config verification/updating

services/prometheus/scrape_configs.go
f41gh7: I think, that removing prometheus will require some work anyway and copy paste code isn't good. For now we can reuse it.

services/victoriametrics/victoriametrics.go
f41gh7: Added true case.

services/supervisord/supervisord.go
f41gh7: Done, config rendered in all cases.

services/prometheus/scrape_configs.go
```diff
@@ -474,3 +477,168 @@ func scrapeConfigsForExternalExporter(s *models.MetricsResolutions, params *scra
...
+			}
+			paramsHost = externalExporterNode.Address
+		default:
+			l.Warnf("It's not possible to get host, skipping scrape config for %s.", agent)
+
+			continue
+		}
+
+		var scfgs []*config.ScrapeConfig
+		switch agent.AgentType {
```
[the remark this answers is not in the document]
f41gh7: checked above if agent.AgentType == models.PMMAgentType {

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,336 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/services/prometheus"
+)
+
+const updateBatchDelay = 3 * time.Second
+
+var (
+	enabled       = os.Getenv("PERCONA_TEST_VM") == "1" || os.Getenv("PERCONA_TEST_VM") == "true"
```
AlekSi, line 48: And what if it is "TRUE"? Please use strconv.ParseBool as in other pmm-managed code

f41gh7: change dto strconv.Bool

PR #451, opened 2020-09-02
SAAS-63 Move starlark env.Run to binary

palash25:
[Ticket](https://jira.percona.com/browse/SAAS-63)
[SAAS PR updating Result type](https://github.com/percona-platform/saas/pull/13)
[FB](https://github.com/Percona-Lab/pmm-submodules/pull/1090)


services/checks/checks.go
```diff
@@ -578,23 +590,63 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
...
+
+	input := &StarlarkScriptData{
+		CheckName:   check.Name,
+		Script:      check.Script,
+		Funcs:       funcs,
+		PrintFn:     l.Debugln,
+		ScriptInput: scriptInput,
+	}
+
+	process := exec.Command("./bin/pmm-managed-starlark")
```
palash25, line 612: keep getting an error saying `fork/exec ./bin/pmm-managed-starlark: no such file or directory`

AlekSi: Just "pmm-managed-starlark" – it is in the $PATH

services/checks/checks.go
```diff
@@ -578,23 +590,63 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
...
+	process.SysProcAttr = &syscall.SysProcAttr{
+		Pdeathsig: syscall.SIGTERM,
```
AlekSi, line 614: Use `pdeathsig.Set` from https://github.com/percona/pmm/tree/PMM-2.0/utils/pdeathsig

services/checks/checks.go
```diff
@@ -578,23 +590,63 @@ func (s *Service) processResults(ctx context.Context, check check.Check, target
...
+	_ = unix.Setrlimit(unix.RLIMIT_CPU, &unix.Rlimit{Cur: 4, Max: 4})
+	_ = unix.Setrlimit(unix.RLIMIT_DATA, &unix.Rlimit{Cur: 100000000, Max: 100000000})
+
+	pipe, err := process.StdinPipe()
+	defer pipe.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	encoder := gob.NewEncoder(pipe)
```
AlekSi, line 625: pmm-agent sends protobuf-encoded data to pmm-managed; just pass the same data there. No need to use gob or anything else

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +52,25 @@ func main() {
 	if on, _ := strconv.ParseBool(os.Getenv("PMM_TRACE")); on {
 		logrus.SetLevel(logrus.TraceLevel)
 	}
+
+	gobDecoder := gob.NewDecoder(os.Stdin)
```
AlekSi, line 56: JSON, not Gob. I think result type has `json` tags already. If not, please add them

AlekSi: https://github.com/percona-platform/platform/blob/032c4cfdf444fac4fddaa95a1e4ca806ce496ea1/pkg/check/result.go#L51-L57 – please add `json` struct tags there

services/checks/checks.go
palash25: pmm-agent only gives us the input for the scripts but I need other data too like the check name, script body, funcs, etc. 

This is what I pass to the binary 
```go
type StarlarkScriptData struct {
	CheckName   string            
	Script      string                    `
	Funcs       map[string]starlark.GoFunc 
	PrintFn     starlark.PrintFunc         
	ScriptInput []map[string]interface{}   
}
```

and for that I need `gob` because `json` can't encode types like `starlark.GoFunc `

AlekSi: gob can't actually encode Go code either. And why you want to pass it via stdin instead of initialising it in the binary itself?

palash25: by initializing in the binary I think you mean `starlark.NewEnv` which I already do, but the arguments to this function and `env.Run` are in `checks.go` which I need to supply to the binary

AlekSi: Yeah, so want prevents you from importing that package in pmm-managed-starlark and using it?

palash25: moved some function calls to the binary but I still need to pass some extra data like the check name, version and the script which cannot be available in the binary by importing a package and calling a function since these are arguments to `processResults` for that I have to encode them in a json struct and send it over using STDIN

services/checks/checks.go
```diff
@@ -565,46 +574,83 @@ type sttCheckResult struct {
...
+	cmdCtx, cancel := context.WithTimeout(ctx, scriptTimeout)
+	defer cancel()
+
+	process := exec.CommandContext(cmdCtx, "pmm-managed-starlark")
+	pdeathsig.Set(process, syscall.SIGTERM)
+	err = unix.Setrlimit(unix.RLIMIT_CPU, &unix.Rlimit{Cur: cpuLimit, Max: cpuLimit})
+	if err != nil {
+		l.Warn("Failed to limit CPU usage: ", err)
+	}
+	//err = unix.Setrlimit(unix.RLIMIT_DATA, &unix.Rlimit{Cur: memoryLimit, Max: memoryLimit})
```
palash25, line 615: @AlekSi if I set a memory limit then I get an `out of memory` error, I am guessing the reason for that is that the memory is being set for the parent process too rather than just the child process. Maybe I am doing it the wrong way, can you take a look at this and tell me what I am doing wrong?

cmd/pmm-managed-starlark/main.go
palash25: done

services/checks/checks.go
AlekSi: Please uncomment this code to log warning, I will take a look

AlekSi:
Where is PR for updating github.com/percona-platform/saas/pkg/check/result.go ?

palash25:
>  Where is PR for updating github.com/percona-platform/saas/pkg/check/result.go ?

oh fuck I made the changes to the vendor file thinking that struct was a part of this repo, will fix that and add a new PR to the saas repo

palash25:
https://github.com/percona-platform/saas/pull/13

Gopkg.toml
```diff
@@ -20,7 +20,7 @@ required = [
 
 [[constraint]]
   name = "github.com/percona-platform/saas"
-  branch = "main"
+  branch = "SAAS-63-Better-Starlark-Sandbox"
```
palash25, line 23: revert after https://github.com/percona-platform/saas/pull/13 is merged

cmd/pmm-managed-starlark/main.go
AlekSi: @palash25 Not a typo – it is in the platform repo

services/checks/checks.go
```diff
@@ -71,6 +75,11 @@ const (
 
 	alertsPrefix        = "/stt/"
 	maxSupportedVersion = 1
+
+	// limits for running pmm-managed-starlark
+	cpuLimit      = 4         // 4 seconds of CPU time
+	memoryLimit   = 100000000 // 100MB of memory in bytes
```
artemgavrilov, line 81: Probably it will be nice to have units in a variable name, `sec` and `bytes`

Also 100MB it is 104857600 Bytes, you can write it as:
`memoryLimitBytes   = 100 * 1024 * 1024`

services/checks/checks.go
artemgavrilov: As I understand yes, `Setrlimit` adds resource limitations to the current process and then it is inherited by child processes. 
https://man7.org/linux/man-pages/man2/setrlimit.2.html
```
       A child process created via fork(2) inherits its parent's resource
       limits.  Resource limits are preserved across execve(2).
```

artemgavrilov: > an unprivileged process may set only **its** soft limit to a value in the range from 0 up to the hard limit, and (irreversibly) lower its hard limit.

Seems that solution is move limitations setting into the child process.

cmd/pmm-managed-starlark/main.go
palash25: https://github.com/percona-platform/platform/pull/127

services/checks/checks.go
palash25: > Seems that solution is move limitations setting into the child process.

yeah but I can't find a good way to do that, there is one way, that would be to get the child's pid and set the limits against the pid, but by the the pid the process would have already started.

cmd/pmm-managed-starlark/main.go
AlekSi: Merged, saas repo update, please use JSON there

Gopkg.toml
AlekSi: saas repo updated

services/checks/checks.go
palash25: I could run a shell instance and run ulimit along with the binary something like this `process := exec.CommandContext(cmdCtx, "/bin/sh", "-c", "ulimit -m 100000000; pmm-managed-starlark")`

This will return `exit status 137` on running scripts like these 
```python
script: |
          def check(rows):
              return [1] * (1 << 30-1)
```

cc @AlekSi

palash25: I could run a shell instance and run ulimit along with the binary something like this process := exec.CommandContext(cmdCtx, "/bin/sh", "-c", "ulimit -m 100000000; pmm-managed-starlark")

This will return exit status 137 on running scripts like these

script: |
          def check(rows):
              return [1] * (1 << 30-1)

cc @AlekSi

AlekSi: I think we should just set limits in `pmm-managed-starlark` binary itself

services/checks/checks.go
```diff
@@ -565,46 +574,83 @@ type sttCheckResult struct {
...
+		CheckVersion: sttCheck.Version,
+		ScriptInput:  r,
+	}
+
+	cmdCtx, cancel := context.WithTimeout(ctx, scriptTimeout)
+	defer cancel()
+
+	process := exec.CommandContext(cmdCtx, "pmm-managed-starlark")
+	pdeathsig.Set(process, syscall.SIGTERM)
+	err = unix.Setrlimit(unix.RLIMIT_CPU, &unix.Rlimit{Cur: cpuLimit, Max: cpuLimit})
```
AlekSi, line 611: You do not use `process` there, so this line sets limit to pmm-managed binary, not pmm-managed-starlark

AlekSi (changes requested):

cmd/pmm-managed-starlark/main.go
```diff
@@ -17,15 +17,26 @@
...
 
 import (
+	"encoding/json"
+	"fmt"
 	"log"
 	"os"
 	"strconv"
 
+	"github.com/percona-platform/saas/pkg/starlark"
+	"github.com/percona/pmm-managed/services/checks"
```
AlekSi, line 27: Imports from this repository should be in the third group

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
 	result    check.Result
 }
 
-func (s *Service) processResults(ctx context.Context, check check.Check, target target, resID string) ([]sttCheckResult, error) {
+// StarlarkScriptData represents the data we need to pass to the binary to run starlark scripts.
+type StarlarkScriptData struct {
+	CheckName    string                   `json:"name"`
+	Script       string                   `json:"script"`
+	CheckVersion uint32                   `json:"version"`
+	ScriptInput  []map[string]interface{} `json:"input"`
```
AlekSi, line 578: ⚠️ 

> pass input as binary protobuf we receive from pmm-agent via stdin

The reason is that JSON supports only a small subset of types we use. For example, information about all specific integer types is lost (all JSON numbers are float64), time.Time type information is lost, etc. So yeah, we really have to pass the binary data there. That would require you to move `agentpb.UnmarshalActionQueryResult` call out of `waitForResult` to pmm-managed-starlark

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
+		Script:       sttCheck.Script,
+		CheckVersion: sttCheck.Version,
+		ScriptInput:  r,
+	}
+
+	cmdCtx, cancel := context.WithTimeout(ctx, scriptTimeout)
+	defer cancel()
+
+	process := exec.CommandContext(cmdCtx, "pmm-managed-starlark")
+	pdeathsig.Set(process, syscall.SIGTERM)
```
AlekSi, line 606: It should be SIGKILL – if our program hang during clean shutdown, for example, it may ignore SIGTERM, but SIGKILL will always be handled by a kernel

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +57,48 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	results, err := env.Run(data.CheckName, data.ScriptInput, l.Debugln)
+	if err != nil {
+		l.Error("Error running starlark env: ", err)
+		os.Exit(1)
+	}
+
+	jsonResults, err := json.Marshal(results)
```
AlekSi, line 98: `json.NewEncoder` for symmetry with `json.NewDecoder` above?

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
+
+	pipe, err := process.StdinPipe()
```
AlekSi, line 608: Can we avoid pipes and set `process.Stdin` directly? Pipes closing is surprisingly complicated and error-prone process.

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
+	defer pipe.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	encoder := json.NewEncoder(pipe)
+	err = encoder.Encode(input)
 	if err != nil {
+		l.Error("Error encoding data to STDIN, ", err)
```
AlekSi, line 617: Strange comma in the error message

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
 		return nil, err
 	}
 
-	env, err := starlark.NewEnv(check.Name, check.Script, funcs)
+	procOut, err := process.CombinedOutput()
```
AlekSi, line 621: It should not be combined output – we need only stdout. Stderr should be captured as usual.

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to prepare starlark environment")
+		l.Error("Error processing script output, ", err)
```
AlekSi, line 623: Let's also log `os/exec.ExitError.Stderr` if available

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
 
-	l := s.l.WithFields(logrus.Fields{
-		"name":       check.Name,
-		"id":         resID,
-		"service_id": target.serviceID,
-	})
-	l.Debugf("Running check script with: %+v.", r)
-	results, err := env.Run(check.Name, r, l.Debugln)
+	var results []check.Result
+	err = json.Unmarshal(procOut, &results)
```
AlekSi, line 628: json.NewDecoder

services/checks/checks.go
```diff
@@ -565,46 +570,75 @@ type sttCheckResult struct {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to execute script")
+		l.Error("Error processing json output, ", err)
+		return nil, err
```
AlekSi, line 631: In general, we should either log error or return it (so it will be logged by the caller), not both

AlekSi (changes requested):

services/checks/checks.go
palash25: > os/exec.ExitError.Stderr

I tried this, it was empty

`process.stderr` however was not but it also prints a huge stacktrace along with the error

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,85 @@
...
+				CheckName:         "invalid query action result",
+				Script:            "def check(): return []",
+				CheckVersion:      1,
+				QueryActionResult: "some result",
+			},
+			err:    true,
+			panics: false,
+		},
+		{
+			data: checks.StarlarkScriptData{
```
palash25, line 50: for some reason this test case fails, even though it is supposed to panic, it fails the test, I can't figure out why. PTAL @AlekSi @artemgavrilov

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,85 @@
...
+				CheckName:         "bad starlark script",
+				Script:            "def check(rows): return [1] * (1 << 30-1)",
+				CheckVersion:      1,
+				QueryActionResult: "\n\rVariable_name\n\x05Value\x12\x1c\n\t2\aversion\n\x0f2\r5.7.30-33-log\x12I\n\x112\x0fversion_comment\n422Percona Server (GPL), Release 33, Revision 6517692\x12%\n\x192\x17version_compile_machine\n\b2\x06x86_64\x12\x1f\n\x142\x12version_compile_os\n\a2\x05Linux\x12\x1a\n\x102\x0eversion_suffix\n\x062\x04-log",
```
AlekSi, line 54: C'mon! 😆 We can do better there

palash25: I just logged the result of the script that I was using locally and put it there, what changes do you want me to make here?

AlekSi: To make it human-readable and -editable, of course. What prevents us from accepting data, encoding it, and passing to pmm-managed-starlark script where it will be decoded?

cmd/pmm-managed-starlark/main_test.go
AlekSi: Ok, but we don't have a code that panics, so why it should panic?

palash25: this case will run this script `"def check(rows): return [1] * (1 << 30-1)"` which will fail with `runtime error: out of memory` and a traceback, isn't that a panic?

AlekSi: Nope, that unrecoverable runtime error, not a panic

palash25: i don't think I can test that then, will remove this case

cmd/pmm-managed-starlark/main_test.go
palash25: so you want the string encoded separately and then sent with the json to the script where the json will be decoded and then the string can be further decoded before passing it to the starlark env?

I am not sure which encoding scheme to use to dealing with a string with ascii and non-printable characters mixed like this one. Do you have any suggestions for that?

AlekSi: First of all, I want to be able to change test data. The current way of doing that does not allow it.

That binary data is a result of some function call, so why we not use the parameter of that function call as a test data, and the call that function ourselves in test?

palash25: > That binary data is a result of some function call, so why we not use the parameter of that function call as a test data, and the call that function ourselves in test?

so the function call that gets me that data is `models.FindActionResultByID(s.db.Querier, resultID)` inside the `waitForResult` method, if I use `models.FindActionResultByID` arguments as the test data that means moving it to `pmm-managed-starlark` and supplying the arguments, the problem with that approach is that since `FindActionResultByID` gets called in a infinite for loop I will essentially have to rewrite the `waitForResults` method in the binary and for that I will need to send more data through STDIN like the context, pointer to `checks.Service` to be able to call `s.db.Delete(res)` afterwards, `resultCheckInterval` to run the ticker.

I just don't see how I can separate just the call to `models.FindActionResultByID` from the rest of the logic and move it to the binary 

One more thing after this review https://github.com/percona/pmm-managed/pull/451#discussion_r484532663  I moved the call to `agentpb.UnmarshalActionQueryResult` to the binary, but it broke the checks during QA, all of them failed while unmarshalling with `unexepected EOF` and when I moved the call to `agentpb.UnmarshalActionQueryResult` back to `waitForResults` method, the checks started working again.

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,85 @@
...
+			if tc.panics {
+				require.Panics(t, func() { runChecks(l, tc.data) })
+			} else {
+				err := runChecks(l, tc.data)
+				if tc.err {
+					require.Error(t, err)
+				} else {
+					require.NoError(t, err)
+				}
+			}
```
artemgavrilov, line 82:
```suggestion
			if tc.panics {
				require.Panics(t, func() { runChecks(l, tc.data) })
				return
			}

			err := runChecks(l, tc.data)
			if tc.err {
				require.Error(t, err)
				return
			}

			require.NoError(t, err)
```

palash25: i was wrong about the panics so I rewrote this, marking this as resolved.

palash25:
@AlekSi is there a way to fix this CI error? `git diff --exit-code" failed`

AlekSi:
@palash25 `make format`

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{"Value": "-log", "Variable_name": "version_suffix"},
+	}
+
+	invalidQueryActionResult = []map[string]interface{}{
+		{"key-1": "val-1", "key-2": "val-2"},
+	}
+)
+
+func TestRunChecks(t *testing.T) {
+	testCases := []struct {
```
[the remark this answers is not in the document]
palash25: fixing this in the next push

palash25:
tests are failing after doing a `dep ensure`, looking into it

cmd/pmm-managed-starlark/main.go
```diff
@@ -17,13 +17,25 @@
...
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 	"gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/percona/pmm-managed/services/checks"
+)
+
+const (
+	cpuLimit    = 4                 // 4 seconds of CPU time
+	memoryLimit = 100 * 1024 * 1024 // 100MB of memory in bytes
```
AlekSi, line 38: That's one of the cases where I think better type and better variable name would be better than a comment: use time.Duration for cpuLimit and `memoryLimitBytes`

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +58,59 @@ func main() {
...
+
+	decoder := json.NewDecoder(os.Stdin)
+	var data checks.StarlarkScriptData
+	err := decoder.Decode(&data)
+	if err != nil {
+		l.Errorf("Error decoding json data: %s", err)
+		os.Exit(1)
+	}
+
+	err = runChecks(l, data)
```
AlekSi, line 72: Let's make `runChecks` return results so both stdin and stdout are in the same function

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +58,59 @@ func main() {
...
+	if err != nil {
+		l.Errorf("Error running starlark script: %s", err)
+		os.Exit(1)
+	}
+}
+
+func runChecks(l *logrus.Entry, data checks.StarlarkScriptData) error {
```
AlekSi, line 79: Pass structs via pointers

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +58,59 @@ func main() {
...
+	err := unix.Setrlimit(unix.RLIMIT_CPU, &unix.Rlimit{Cur: cpuLimit, Max: cpuLimit})
+	if err != nil {
+		l.Warnf("Failed to limit CPU usage: %s", err)
+	}
+	err = unix.Setrlimit(unix.RLIMIT_DATA, &unix.Rlimit{Cur: memoryLimit, Max: memoryLimit})
```
AlekSi, line 84: Are you sure it should be RLIMIT_DATA?

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+		err     bool
+		version uint32
+		name    string
+		script  string
+		result  []map[string]interface{}
+	}{
+		{
+			err:     true,
+			version: 5,
+			name:    "invalid version",
```
AlekSi, line 56: I think we have tests for that elsewhere, let's not overtest

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+		{
+			err:     false,
+			version: 1,
+			name:    "valid starlark script",
+			script:  "def check(rows): return []",
+			result:  validQueryActionResult,
+		},
+	}
+
+	err := exec.Command("/bin/sh", "-c", "cd ../../; make release").Run()
```
AlekSi, line 90: `("make", "-C", "../..", "release")`

And let's add a comment explaining why we do that

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+		{
+			err:     true,
+			version: 1,
+			name:    "invalid query action result",
+			script:  "def check(): return []",
+			result:  invalidQueryActionResult,
+		},
+		{
+			err:     true,
```
AlekSi, line 75: I think we should check stderr content via ExitError.Stderr

AlekSi (commented):
Almost there!

cmd/pmm-managed-starlark/main.go
palash25: `unix.Rlimit` expects the limit to be in `uint64` so I change `cpuLimit` to  `time.Duration` will have to convert it back to `uint64` to be able to use it, so I would prefer to let it stay the way it is.

:+1: for `memoryLimitBytes`

AlekSi: `uint64(cpuLimit.Seconds())`

cmd/pmm-managed-starlark/main.go
palash25: i was trying to decide between `RLIMIT_DATA` and `RLIMIT_MEMLOCK`

From the manpage

```
RLIMIT_DATA
    The maximum size of the process's data segment (initialized data, uninitialized data, and heap).
```

> (initialized data, uninitialized data, and heap)

this :point_up: seemed to me like what you were referring to when talking about the process's memory

and 

I couldn't fully understand what `RLIMIT_MEMLOCK` was for from the manual
```
RLIMIT_MEMLOCK
    The maximum number of bytes of memory that may be locked into RAM. In effect this limit is rounded down to the nearest multiple of the system page size. This limit affects mlock(2) and mlockall(2) and the mmap(2) MAP_LOCKED operation. Since Linux 2.6.9 it also affects the shmctl(2) SHM_LOCK operation, where it sets a maximum on the total bytes in shared memory segments (see shmget(2)) that may be locked by the real user ID of the calling process. The shmctl(2) SHM_LOCK locks are accounted for separately from the per-process memory locks established by mlock(2), mlockall(2), and mmap(2) MAP_LOCKED; a process can lock bytes up to this limit in each of these two categories. In Linux kernels before 2.6.9, this limit controlled the amount of memory that could be locked by a privileged process. Since Linux 2.6.9, no limits are placed on the amount of memory that a privileged process may lock, and this limit instead governs the amount of memory that an unprivileged process may lock. 
```

do you want to set some other kind of limit for the memory? if so please point it out to me.

AlekSi: Add me to the "Linux is a mess" camp – why on Earth data segment includes heap?! Everywhere else data, bss, and heap are in separate sections…

Ok, let's use RLIMIT_DATA

cmd/pmm-managed-starlark/main_test.go
palash25: i logged `string(ExitError.Stderr)` in the tests but it turned out to be empty

palash25: whereas if I log it from `checks.go` I can see the error and the stack trace

palash25: do you still want me to check it?

AlekSi: Yeah, we should ensure that this bad script is killed by our limit, not by generic Linux OOM killer, for example. There should be a way to do that

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,116 @@
...
+		name    string
+		script  string
+		result  []map[string]interface{}
+	}{
+		{
+			err:     true,
+			version: 1,
+			name:    "invalid starlark syntax",
+			script:  "@ + @",
+			result:  invalidQueryActionResult,
```
AlekSi, line 58: we test it elsewhere

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+		{
+			err:     true,
+			version: 1,
+			name:    "invalid query action result",
+			script:  "def check(): return []",
+			result:  invalidQueryActionResult,
```
AlekSi, line 65: we test it elsewhere

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+		{
+			err:     true,
+			version: 1,
+			name:    "bad starlark script",
+			script:  "def check(rows): return [1] * (1 << 30-1)",
+			result:  validQueryActionResult,
```
AlekSi, line 72: We should also add a test for CPU time limit (infinite loop, for example)

cmd/pmm-managed-starlark/main_test.go
palash25: ok I logged the `stderr` of the child process and got the whole stacktrace, and tried to match the first line for each test case, this is what I get from the memory consuming starlark script `fatal error: runtime: out of memory`

I also tried writing a test for CPU time I couldn't find a way to write an infinite loop as starlark uses `for-range` and doesn't allow `while`so I wrote a very long loop instead 

```
def check(rows):
						for x in range(10000000):
							pass
```

which gives me this error `fatal error: runtime: cannot allocate memory` and sometimes I get this `signal killed` which makes it unreliable

I think the reason for that is 

```
RLIMIT_CPU
    CPU time limit in seconds. When the process reaches the soft limit, it is sent a SIGXCPU signal. The default action for this signal is to terminate the process. However, the signal can be caught, and the handler can return control to the main program. If the process continues to consume CPU time, it will be sent SIGXCPU once per second until the hard limit is reached, at which time it is sent SIGKILL. (This latter point describes Linux behavior. Implementations vary in how they treat processes which continue to consume CPU time after reaching the soft limit. Portable applications that need to catch this signal should perform an orderly termination upon first receipt of SIGXCPU.) 
```

so I guess I will try to catch this signal now and kill the process forcefully from the code instead of relying on the OS

palash25: @AlekSi is there a way to write an infinite loop in starlark? I looked but couldn't find it

AlekSi: https://godoc.org/go.starlark.net/resolve, `resolve.AllowRecursion = true` if some special flag is passed to pmm-managed-starlark (that should be passed by default, but will be passed in test)

cmd/pmm-managed-starlark/main_test.go
AlekSi: Ok, I probably should clarify that.

We test various incorrect scripts in pkg/starlark package, so we should not test `env.Run` handling of invalid input, `check` function not accepting any arguments, etc.

But we definitely should test what we do there, in this package: that invalid scripts or data result in exit code 1 with an error message in stderr, and valid scripts with valid data actually return results in JSON in stdout (with exit code 0). Are both cases not covered now?

cmd/pmm-managed-starlark/main.go
AlekSi: The comment above has 👍, but is not addressed

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +63,63 @@ func main() {
...
+	}
+
+	results, err := runChecks(l, &data)
+	if err != nil {
+		l.Errorf("Error running starlark script: %s", err)
+		os.Exit(1)
+	}
+
+	encoder := json.NewEncoder(os.Stdout)
+	err = encoder.Encode(results)
```
AlekSi, line 87: Please add a test for that case: stdout (contains valid result JSON), stderr (empty), exit code (0)

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +63,63 @@ func main() {
...
+	if err != nil {
+		l.Errorf("error marshalling JSON results: %s", err)
+		os.Exit(1)
+	}
+}
+
+func runChecks(l *logrus.Entry, data *checks.StarlarkScriptData) ([]check.Result, error) {
+	err := unix.Setrlimit(unix.RLIMIT_CPU, &unix.Rlimit{Cur: cpuLimit, Max: cpuLimit})
+	if err != nil {
+		l.Warnf("Failed to limit CPU usage: %s", err)
```
AlekSi, line 97: We should test that this line is not present in stderr (and for that we probably should move that message to constant to avoid breaking our negative test with logging tweak)

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +63,63 @@ func main() {
...
+	}
+	err = unix.Setrlimit(unix.RLIMIT_DATA, &unix.Rlimit{Cur: memoryLimitBytes, Max: memoryLimitBytes})
+	if err != nil {
+		l.Warnf("Failed to limit memory usage: %s", err)
```
AlekSi, line 101: We should test that this line is not present in stderr (and for that we probably should move that message to constant to avoid breaking our negative test with logging tweak)

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+	{"Value": "5.7.30-33-log", "Variable_name": "version"},
+	{"Value": "Percona Server (GPL), Release 33, Revision 6517692", "Variable_name": "version_comment"},
+	{"Value": "x86_64", "Variable_name": "version_compile_machine"},
+	{"Value": "Linux", "Variable_name": "version_compile_os"},
+	{"Value": "-log", "Variable_name": "version_suffix"},
+}
+
+func TestRunChecks(t *testing.T) {
+	testCases := []struct {
+		err           bool
```
AlekSi, line 43: Why have this field if we have both `errorMessage` and `stdErrContent` below? (and then why we have both?)

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+		version       uint32
+		name          string
+		script        string
+		errorMessage  string
+		stdErrContent string // first line of the stack trace
+		result        []map[string]interface{}
```
AlekSi, line 49: That's script's incoming data, right? The name and the positioning (after script error results) is confusing

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func TestRunChecks(t *testing.T) {
+	testCases := []struct {
+		err           bool
+		version       uint32
+		name          string
+		script        string
+		errorMessage  string
+		stdErrContent string // first line of the stack trace
```
AlekSi, line 48: What we really should check is stdout (valid JSON), stderr (probably with regex) and exit code

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			var stdin, stderr bytes.Buffer
+			cmd.Stdin = &stdin
+			cmd.Stderr = &stderr
+
+			encoder := json.NewEncoder(&stdin)
+			err = encoder.Encode(data)
+			require.NoError(t, err)
+
+			err = cmd.Run()
```
AlekSi, line 111: We should also test stdout

AlekSi (changes requested):
Finally, don't forget to rewind FB – that's what our QA is testing

AlekSi:
Changed base to make PR much smaller

cmd/pmm-managed-starlark/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+			stdErrContent: "",
+			result:        validQueryActionResult,
+		},
+	}
+
+	// since we run the binary as a child process to test it we need to build it first.
+	err := exec.Command("make", "-C", "../..", "release").Run()
+	require.NoError(t, err)
+
+	err = os.Setenv("PERCONA_TEST_STARLARK_ALLOW_RECURSION", "true")
```
AlekSi, line 86: There is no need to set `PERCONA_TEST_STARLARK_ALLOW_RECURSION` for the parent process – use `cmd.Env` below

cmd/pmm-managed-starlark/main_test.go
palash25: added more cases for `exit status 1` errors now

cmd/pmm-managed-starlark/main_test.go
palash25: > (probably with regex) 

I am gonna use `strings.Contains` since we are looking for specific errors in each case rather than a pattern

cmd/pmm-managed-starlark/main.go
palash25: sorry forgot that, made the change will push

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +63,63 @@ func main() {
...
+	if err != nil {
+		l.Errorf("Error running starlark script: %s", err)
+		os.Exit(1)
+	}
+
+	encoder := json.NewEncoder(os.Stdout)
+	err = encoder.Encode(results)
+	if err != nil {
+		l.Errorf("error marshalling JSON results: %s", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
palash25: tried but couldnt find a case that would make it fail

cmd/pmm-managed-starlark/main.go
```diff
@@ -46,4 +63,63 @@ func main() {
...
+	}
+
+	l := logrus.WithField("component", "pmm-managed-starlark")
+
+	decoder := json.NewDecoder(os.Stdin)
+	var data checks.StarlarkScriptData
+	err := decoder.Decode(&data)
+	if err != nil {
+		l.Errorf("Error decoding json data: %s", err)
+		os.Exit(1)
```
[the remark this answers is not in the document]
palash25: tried but couldn't find an invalid query action result that would make it fail

cmd/pmm-managed-starlark/main_test.go
palash25: renamed it and changed a few positions

cmd/pmm-managed-starlark/main_test.go
palash25: removed

> (and then why we have both?

because I don't get the exit status from stderr

palash25:
@AlekSi for some reason the CI isn't running on every push

palash25:
closing in favour of https://github.com/percona/pmm-managed/pull/462

PR #455, opened 2020-09-09
PMM-6389 Add VictoriaMetrics

f41gh7:
Adds two new services - victoriametrics and vmalert
 victoriametrics replaces promtheus as metrics storage,
 scrapes targets.
 vmalert evaluates alerting and recording rules,
 sends notifications to alertmanager.
 Added tests and test data.
 Added wrappers for prometheus service functions.

[PMM-6389](https://jira.percona.com/browse/PMM-6389)

https://github.com/Percona-Lab/pmm-submodules/pull/1143

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


f41gh7:
moved to ready for review, coz it needs for triggering CI and MR was reviewed by @valyala

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,383 @@
...
+var (
+	enabled       bool
+	loadEnabled   = sync.Once{}
+	checkFailedRE = regexp.MustCompile(`FAILED: parsing YAML file \S+: (.+)\n`)
+)
+
+// Enabled indicates whether VictoriaMetrics enabled or not.
+func Enabled() bool {
+	loadEnabled.Do(func() {
+		if enabledVar := os.Getenv("PERCONA_TEST_VM"); enabledVar != "" {
```
tenmozes, line 62: return fast if empty

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,383 @@
...
+			parsedBool, err := strconv.ParseBool(enabledVar)
+			if err != nil {
+				panic(fmt.Sprintf("cannot parse PERCONA_TEST_VM, as bool, value: %s", enabledVar))
+			}
+			enabled = parsedBool
+		}
+		enabled = false
```
tenmozes, line 69: no need, it's already false

tenmozes (changes requested):

main.go
```diff
@@ -446,6 +448,15 @@ func main() {
 	prometheusConfigF := kingpin.Flag("prometheus-config", "Prometheus configuration file path").Default("/etc/prometheus.yml").String()
 	prometheusURLF := kingpin.Flag("prometheus-url", "Prometheus base URL").Default("http://127.0.0.1:9090/prometheus/").String()
 
+	victoriaMetricsURLF := kingpin.Flag("victoriametrics-url", "VictoriaMetrics base URL").
+		Default("http://127.0.0.1:8428/").String()
+	victoriaMetricsVMAlertURLF := kingpin.Flag("victoriametrics-vmalert-url", "VictoriaMetrics VMAlert base URL").
+		Default("http://127.0.0.1:8880/").String()
+	victoriaMetricsConfigF := kingpin.Flag("victoriametrics-config", "VictoriaMetrics scape configuration file path").
+		Default("/etc/victoriametrics-promscrape.yml").String()
+	victoriaMetricsBaseConfigF := kingpin.Flag("victoriametrics-baseconfig", "VictoriaMetrics basic configuration file, can be mounted to container").
```
tenmozes, line 457: could you please expand the description and define the purpose of the flags for base config and config

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	return vmp, nil
+}
+
+// UpdateParams - updates params for VictoriaMetrics services
+// reads configuration file and updates corresponding flags.
+func (vmp *VictoriaMetricsParams) UpdateParams() error {
+	if err := vmp.loadVMAlertParams(); err != nil {
+		return err
+	}
+	if err := vmp.loadVMDBParams(); err != nil {
```
tenmozes, line 74: just return
do we need to add context to errors?

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+
+		return nil
+	}
+	cfg := &config.Config{}
+	err = yaml.Unmarshal(buf, cfg)
+	if err != nil {
```
tenmozes, line 93: if err := err = yaml.Unmarshal(buf, cfg); err != nil .....

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+		return vmp, err
+	}
+
+	return vmp, nil
+}
+
+// UpdateParams - updates params for VictoriaMetrics services
+// reads configuration file and updates corresponding flags.
+func (vmp *VictoriaMetricsParams) UpdateParams() error {
```
tenmozes, line 70: do we need to have cache here? and don't update params if they are equal?

services/supervisord/supervisord.go
```diff
@@ -515,6 +558,64 @@ stdout_logfile_backups = 3
...
+priority = 7
+command =
+	/usr/sbin/vmalert
+        --notifier.url="http://127.0.0.1:9093/alertmanager{{- if .AlertmanagerURL }},{{ .AlertmanagerURL }}{{- end }}"
+{{- if and  .AlertManagerUser .AlertManagerPassword}}
+        --notifier.basicAuth.password=',{{.AlertManagerPassword }}'
+        --notifier.basicAuth.username=",{{.AlertManagerUser}}"
+{{- end }}
+        --external.url=http://localhost/graph
+        --external.alert.source='explore?orgId=1&left=["now-1h","now","Prometheus",{"expr": "{{"{{"}}$expr|quotesEscape|pathEscape{{"}}"}}"},{"mode":"Metrics"},{"ui":[true,true,true,"none"]}]'
```
tenmozes, line 597: as we discussed we don't this flag yet

utils/envvars/parser.go
```diff
@@ -34,6 +34,11 @@ type InvalidDurationError string
 
 func (e InvalidDurationError) Error() string { return string(e) }
 
+// InvalidIntegerError invalid integer error.
+type InvalidIntegerError string
```
tenmozes, line 38: is it used?

tenmozes (changes requested):

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+// VictoriaMetricsParams - defines flags and settings for victoriametrics.
+type VictoriaMetricsParams struct {
+	VMDB           []string
+	VMAlert        []string
+	Enabled        bool
+	baseConfigPath string
+}
+
+// NewVictoriaMetricsParams - returns configuration params for VictoriaMetrics
+// error can be suppressed, this func always return inited struct with default
```
valyala, line 44: The `error can be suppressed` comment isn't needed here. If the error isn't checked by the caller, then there is no need in returning it. Probably, it would be better adding logging for errors inside the function.

It would be great also explaining why there is no need in checking errors by the caller. As I understand, this is because it is expected that such errors may be fixed later by the user by reloading the config with `UpdateParams` call.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+func NewVictoriaMetricsParams(basePath string) (*VictoriaMetricsParams, error) {
+	vmp := &VictoriaMetricsParams{
+		VMDB:           []string{},
+		VMAlert:        []string{},
+		baseConfigPath: basePath,
+	}
+	if enabledVar := os.Getenv(vmTestEnableEnv); enabledVar != "" {
+		parsedBool, err := strconv.ParseBool(enabledVar)
+		if err != nil {
+			return vmp, fmt.Errorf("cannot parse PERCONA_TEST_VM, as bool, value: %s, err: %w", enabledVar, err)
```
valyala, line 55: Probably it would be better substituting this with `log.Errorf`?

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+const (
+	vmCacheDisableEnv = "VM_CACHE_DISABLE"
+	vmTestEnableEnv   = "PERCONA_TEST_VM"
+)
+
+// VictoriaMetricsParams - defines flags and settings for victoriametrics.
+type VictoriaMetricsParams struct {
+	VMDB           []string
```
valyala, line 37: It would be great adding explanatory comments for public members of `VictoriaMetricsParams` struct.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	Enabled        bool
+	baseConfigPath string
+}
+
+// NewVictoriaMetricsParams - returns configuration params for VictoriaMetrics
+// error can be suppressed, this func always return inited struct with default
+// params.
+func NewVictoriaMetricsParams(basePath string) (*VictoriaMetricsParams, error) {
+	vmp := &VictoriaMetricsParams{
+		VMDB:           []string{},
```
valyala, line 48: There is no need in using empty `[]string` initializers, since they work the same as default `nil` values for `[]string` types.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	err := vmp.UpdateParams()
+	if err != nil {
+		return vmp, err
+	}
+
+	return vmp, nil
+}
+
+// UpdateParams - updates params for VictoriaMetrics services
+// reads configuration file and updates corresponding flags.
```
valyala, line 69: It would be more clear if only the second line of the comment will be left with `UpdateParams ` prefix.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		return nil
+	}
+	cfg := &config.Config{}
+	err = yaml.Unmarshal(buf, cfg)
+	if err != nil {
+		return err
+	}
+	vmalertFlags := make([]string, 0, len(vmp.VMAlert))
```
valyala, line 96: Substitute this line with `var vmalertFlags []string`

models/victoriametrics.go
valyala: What is the purpose of such a cache? It looks like it doesn't prevent from reading all the configs and it doesn't prevent from doing less work. It only may complicate the code.

main.go
valyala: Probably this flag isn't needed? Probably it would be better just exporting `BaseConfigPath` const from `services/prometheus` package and passing it to VictoriaMetrics packages?

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, r := range cfg.RuleFiles {
+		vmalertFlags = append(vmalertFlags, "--rule="+r)
+	}
+	if cfg.GlobalConfig.EvaluationInterval != 0 {
+		vmalertFlags = append(vmalertFlags, "--evaluationInterval="+cfg.GlobalConfig.EvaluationInterval.String())
+	}
+
+	if !reflect.DeepEqual(vmalertFlags, vmp.VMAlert) {
```
valyala, line 104: I think this condition isn't needed, since it doesn't prevent from doing additional work and the end result without this line would be the same.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+		vmp.VMAlert = vmalertFlags
+	}
+
+	return nil
+}
+
+// loadVMDBParams - flags for VictoriaMetrics database can be added here.
+func (vmp *VictoriaMetricsParams) loadVMDBParams() error {
+	cacheDisabled := os.Getenv(vmCacheDisableEnv)
```
valyala, line 113: I think it would be better disabling response cache in VM by default and allowing users enabling it, i.e. to rename `VM_CACHE_DISABLE` env var to `VM_CACHE_ENABLE`.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	"os"
+	"reflect"
+	"strconv"
+
+	config "github.com/percona/promconfig"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	vmCacheDisableEnv = "VM_CACHE_DISABLE"
```
valyala, line 31: Could you add explanatory comments for these ENV vars? Also it would be great if all these env vars consistently start with `VM_*` prefix.

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	if cacheDisabled == "" {
+		return nil
+	}
+	parsedBool, err := strconv.ParseBool(cacheDisabled)
+	if err != nil {
+		return err
+	}
+	vmdbFlags := make([]string, 0, len(vmp.VMDB))
+	vmdbFlags = append(vmdbFlags, fmt.Sprintf("--search.disableCache=%t", parsedBool))
+	if !reflect.DeepEqual(vmdbFlags, vmp.VMDB) {
```
valyala, line 123: This line isn't needed, since it doesn't change the end result and doesn't prevent from doing additional work.

valyala (changes requested):

models/victoriametrics.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := vmp.loadVMDBParams(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// loadVMAlertParams - load params and converts it to vmalert flags.
+func (vmp *VictoriaMetricsParams) loadVMAlertParams() error {
+	buf, err := ioutil.ReadFile(vmp.baseConfigPath)
```
valyala, line 83: Should this function extract params from  https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config ?

main.go
f41gh7: Removed flag.

models/victoriametrics.go
f41gh7: done

services/supervisord/supervisord.go
f41gh7: done

utils/envvars/parser.go
f41gh7: Nope, removed it.

models/victoriametrics.go
f41gh7: Replaced with errors.Wrap

models/victoriametrics.go
f41gh7: Renamed and changed logic.

models/victoriametrics.go
f41gh7: Well, probably later, atm, `VMAlert` doesnt support this config params. And we cannot extract something useful.

valyala (commented):
LGTM

I'm going to update https://github.com/Percona-Lab/pmm-submodules/pull/1020 in order to build new Docker image from this PR, so it could be used in docker-compose.

AlekSi:
@valyala You said "LGTM", but did not approve PR: image

AlekSi:
There are conflicts now, please resolve

docker-compose.yml
```diff
@@ -2,7 +2,8 @@
 version: '3.7'
 services:
   pmm-managed-server:
-    image: ${PMM_SERVER_IMAGE:-perconalab/pmm-server:dev-latest}
+    # REPLACE IT BEFORE MERGE    image: ${PMM_SERVER_IMAGE:-perconalab/pmm-server:dev-latest}
+    image: perconalab/pmm-server-fb:PR-1020-1a8e2cd
```
AlekSi, line 6: Please use the one from https://github.com/Percona-Lab/pmm-submodules/pull/1111 when it will be there

f41gh7: sure

tenmozes (approved):
lgtm

docker-compose.yml
f41gh7: done

valyala (approved):

models/victoriametrics.go
```diff
@@ -0,0 +1,124 @@
...
+	if err := vmp.loadVMDBParams(); err != nil {
+		return errors.Wrap(err, "cannot update VMDBFlags config param")
+	}
+
+	return nil
+}
+
+// loadVMAlertParams - load params and converts it to vmalert flags.
+func (vmp *VictoriaMetricsParams) loadVMAlertParams() error {
+	buf, err := ioutil.ReadFile(vmp.BaseConfigPath)
```
BupycHuk, line 84: Why do we get flags from config file? Was this discussed?

models/victoriametrics.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+
+	// enables cache for vmdb responses.
+	vmCacheEnableEnv = "VM_CACHE_ENABLE"
+	// enables victoriametrics services at supervisor config.
+	vmTestEnableEnv = "VM_TEST_ENABLE"
```
BupycHuk, line 35: can we rename this env variable to `PERCONA_TEST_ENABLE_VM`?

models/victoriametrics.go
```diff
@@ -0,0 +1,124 @@
...
+	if cacheEnable := os.Getenv(vmCacheEnableEnv); cacheEnable != "" {
+		parsedBool, err := strconv.ParseBool(os.Getenv(vmCacheEnableEnv))
+		if err != nil {
+			return errors.Wrapf(err, "failed to parse %s as bool", vmCacheEnableEnv)
+		}
+		// we have to invert parsed variable
+		cacheDisabled = !parsedBool
+	}
+	vmdbFlags := make([]string, 0, len(vmp.VMDBFlags))
+	vmdbFlags = append(vmdbFlags, fmt.Sprintf("--search.disableCache=%t", cacheDisabled))
```
BupycHuk, line 120: the only flag we have in this array is `--search.disableCache`, can we move it to `settings` structure and `envvars` package? @AlekSi what do you think?

services/prometheus/prometheus.go
```diff
@@ -409,6 +414,17 @@ func (svc *Service) marshalConfig() ([]byte, error) {
 	return b, nil
 }
 
+// AddInternalServiceToScrape adds internal services metrics to scrape targets.
+func AddInternalServiceToScrape(cfg *config.Config, s models.MetricsResolutions) {
```
BupycHuk, line 418:
```suggestion
func AddInternalServicesToScrape(cfg *config.Config, s models.MetricsResolutions) {
```

services/supervisord/supervisord.go
```diff
@@ -406,6 +422,45 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
...
+	}
+
+	if u.Opaque != "" || u.Host == "" {
+		templateParams["AlertmanagerURL"] = ""
+
+		return errors.Errorf("AlertmanagerURL parsed incorrectly as %#v", u)
+	}
+	if username := u.User.Username(); username != "" {
+		if password, ok := u.User.Password(); ok {
+			n := url.URL{
```
BupycHuk, line 448: Any reason for this? What will change?

services/supervisord/supervisord.go
```diff
@@ -406,6 +422,45 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
...
+				Scheme:   u.Scheme,
+				Host:     u.Host,
+				Path:     u.Path,
+				RawQuery: u.RawQuery,
+				Fragment: u.Fragment,
+			}
+			templateParams["AlertManagerUser"] = username
```
BupycHuk, line 455: is it possible to have AlertManager user without password?

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+	if _, err = f.Write(cfg); err != nil {
+		return errors.WithStack(err)
+	}
+	defer func() {
+		_ = f.Close()
+		_ = os.Remove(f.Name())
+	}()
+	args := []string{"check", "config", f.Name()}
+	cmd := exec.CommandContext(ctx, "promtool", args...) //nolint:gosec
```
BupycHuk, line 303: don't you have separate tool to check config?

main.go
```diff
@@ -518,23 +527,34 @@ func main() {
 	cleaner := clean.New(db)
 	alertingRules := prometheus.NewAlertingRules()
 
+	vmParams, err := models.NewVictoriaMetricsParams(prometheus.BasePrometheusConfigPath)
```
BupycHuk, line 530: Based on the code looks like we don't need to initialize it here `VM_ENABLED` env var can be extracted in each place separately.

services/supervisord/supervisord.go
```diff
@@ -406,6 +422,45 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
 	return b, nil
 }
 
+// addAlertManagerParam parses alertManagerURL
+// and extracts username and password from it to templateParams.
+func addAlertManagerParam(alertManagerURL string, templateParams map[string]interface{}) error {
+	templateParams["AlertmanagerURL"] = alertManagerURL
```
BupycHuk, line 428: TBH this method looks like it can be simplified.

services/supervisord/supervisord.go
```diff
@@ -406,6 +422,45 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
...
+	templateParams["AlertManagerUser"] = ""
+	templateParams["AlertManagerPassword"] = ""
+	if alertManagerURL == "" {
+		return nil
+	}
+	u, err := url.Parse(alertManagerURL)
+	if err != nil {
+		templateParams["AlertmanagerURL"] = ""
```
BupycHuk, line 436: any reason to set empty value since we return an error.

BupycHuk: the same in cases below.

BupycHuk (changes requested):

models/victoriametrics.go
AlekSi: Agree

models/victoriametrics.go
f41gh7: Coz, user may define own rule files path with prometheus.yaml, mounted to container. And we have to pass this additional rules to `vmalert` supervisor config.

models/victoriametrics.go
f41gh7: Yep, sure.

models/victoriametrics.go
f41gh7: Btw, if need other configuration params in the future, its better to modify `settings` structure with envvars?

services/supervisord/supervisord.go
f41gh7: extract username and password, it will prevent `vmaler`t to print it out to log file and flag value as plaintext.

 This behavior was fixed at last release.

services/supervisord/supervisord.go
f41gh7: Nope, password is checked with `ok` value.

services/supervisord/supervisord.go
f41gh7: Yep, will do it.

services/victoriametrics/victoriametrics.go
f41gh7: We can check it with vmdb --dry-run, but current configuration is incomplete with our checker. So it will fail each time. Later i can replace it, but  in case of backward-capability, i decided to use promtool.

models/victoriametrics.go
BupycHuk: If user should be able to change this value via env vars, then yes.

main.go
BupycHuk: So, as far as I see, the only parameter is used in both places is `VM_ENABLED` and it's going to be removed oce we replace Prometheus with Victoria Metrics. At that moment it won't make sense to pass this structure to VM Alert.

BupycHuk: Let's keep it for now.

services/supervisord/supervisord.go
BupycHuk: So we can just set empty values to Username and Password of `u`, since we don't use `u` anymore, right? to make it clear for everybody

services/supervisord/supervisord.go
BupycHuk: No, I mean does vmalert support alert manager url's with username, but without password?

services/victoriametrics/victoriametrics.go
BupycHuk: ok

services/supervisord/supervisord.go
f41gh7: Yep, exactly. It's also possible to use username and password within url. But it's less secure and dont recommend.

services/supervisord/supervisord.go
f41gh7: Get it. Will fix.

models/victoriametrics.go
f41gh7: Added cacheEnable to settings.

services/supervisord/supervisord.go
f41gh7: Changed logic.

services/supervisord/supervisord.go
```diff
@@ -393,9 +398,20 @@ func (s *Service) reload(name string) error {
...
+	}
 	templateParams := map[string]interface{}{
-		"DataRetentionHours": int(settings.DataRetention.Hours()),
-		"DataRetentionDays":  int(settings.DataRetention.Hours() / 24),
+		"DataRetentionHours":  int(settings.DataRetention.Hours()),
+		"DataRetentionDays":   int(settings.DataRetention.Hours() / 24),
+		"DataRetentionMonths": retentionMonths,
+		"IsVMEnabled":         s.vmParams.Enabled,
+		"VMAlertParams":       s.vmParams.VMAlertFlags,
+		"VMDBCacheEnable":     !settings.VictoriaMetrics.CacheEnable,
```
BupycHuk, line 411: Why `!`?

services/supervisord/supervisord.go
```diff
@@ -406,6 +422,39 @@ func (s *Service) marshalConfig(tmpl *template.Template, settings *models.Settin
...
+	if username := u.User.Username(); username != "" {
+		password, _ := u.User.Password()
+		n := url.URL{
+			Scheme:   u.Scheme,
+			Host:     u.Host,
+			Path:     u.Path,
+			RawQuery: u.RawQuery,
+			Fragment: u.Fragment,
+		}
+		templateParams["AlertManagerUser"] = username
```
BupycHuk, line 450: looks like these two variables(`AlertManagerUser`, `AlertManagerPassword`)  can be moved out of this `if` and https://github.com/percona/pmm-managed/pull/455/files#diff-c2e3f6e9d52d388949d61ae9cfc0bd7bR429-R430 can be removed.

utils/envvars/parser.go
```diff
@@ -135,10 +145,25 @@ func parseStringDuration(value string) (time.Duration, error) {
 	return d, nil
 }
 
+// parseBool validate bool as string value.
+func parseBool(value string) (bool, error) {
+	if len(value) == 0 {
```
BupycHuk, line 150:
```suggestion
	if value == "" {
```
?

utils/envvars/parser.go
```diff
@@ -135,10 +145,25 @@ func parseStringDuration(value string) (time.Duration, error) {
 	return d, nil
 }
 
+// parseBool validate bool as string value.
+func parseBool(value string) (bool, error) {
```
BupycHuk, line 149: Do we really need this function?

BupycHuk (changes requested):

services/supervisord/supervisord.go
f41gh7: Vmdb has setting `-search.disableCache`, so boolean param CacheEnable must be inverted.

utils/envvars/parser.go
f41gh7: This function can be reused for other boolean params in the futute and there is `parseStringDuration` func. But other boolean values parsed in-place, it's a bit strange. I can do the same, should i?

services/supervisord/supervisord.go
f41gh7: If i remove  ```if alertManagerURL == "" {
		return nil
	}``` , parse will raise error.

 And is it correct to rewrite url each time, even if we don't need to do this?

utils/envvars/parser.go
f41gh7: Sry, will change it.

services/supervisord/supervisord.go
BupycHuk: I didn't ask to remove `https://github.com/percona/pmm-managed/pull/455/files#diff-c2e3f6e9d52d388949d61ae9cfc0bd7bR431`, I mentioned two lines above this line.

services/supervisord/supervisord.go
BupycHuk: but you pass it as `VMDBCacheEnable`, not as `VMDBCacheDisable`

services/supervisord/supervisord.go
f41gh7: Ah, get it, but in this case go-template will fail, because of `"missingkey=error"` is set for templates.

services/supervisord/supervisord.go
f41gh7: Yep, it's misleading, thx!

utils/envvars/parser.go
f41gh7: removed

utils/envvars/parser.go
f41gh7: removed

services/supervisord/supervisord.go
f41gh7: changed

services/supervisord/supervisord.go
f41gh7: changed

BupycHuk (approved):

services/supervisord/supervisord.go
```diff
@@ -393,9 +398,20 @@ func (s *Service) reload(name string) error {
...
+		retentionMonths = 1
+	}
 	templateParams := map[string]interface{}{
-		"DataRetentionHours": int(settings.DataRetention.Hours()),
-		"DataRetentionDays":  int(settings.DataRetention.Hours() / 24),
+		"DataRetentionHours":  int(settings.DataRetention.Hours()),
+		"DataRetentionDays":   int(settings.DataRetention.Hours() / 24),
+		"DataRetentionMonths": retentionMonths,
+		"IsVMEnabled":         s.vmParams.Enabled,
+		"VMAlertParams":       s.vmParams.VMAlertFlags,
```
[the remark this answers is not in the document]
f41gh7: renamed

services/server/server.go
```diff
@@ -102,6 +106,8 @@ func NewServer(params *Params) (*Server, error) {
 	s := &Server{
 		db:                      params.DB,
 		prometheus:              params.Prometheus,
+		vmdb:                    params.VictoriaMetrics,
```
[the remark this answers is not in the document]
f41gh7: Changed to VMDB.

AlekSi (changes requested):
Please resolve merge conflicts

f41gh7:
resolved conflicts.

docker-compose.yml
```diff
@@ -2,7 +2,8 @@
 version: '3.7'
 services:
   pmm-managed-server:
-    image: ${PMM_SERVER_IMAGE:-perconalab/pmm-server:dev-latest}
+    # REPLACE IT BEFORE MERGE    image: ${PMM_SERVER_IMAGE:-perconalab/pmm-server:dev-latest}
```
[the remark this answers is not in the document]
AlekSi: @f41gh7 Everything else should be merged now, please update it to re-run CI tests

services/prometheus/alerting_rules_test.go
```diff
@@ -53,14 +53,6 @@ groups:
...
 groups:
 - name: example
-rules:
-- alert: HighRequestLatency
-expr: job:request_latency_seconds:mean5m{job="myjob"} > 0.5
-for: 10m
-labels:
-severity: page
-annotations:
-summary: High request latency
```
AlekSi, line 63: No, that's an issue with updated Prometheus. Please update error message in that test

services/prometheus/prometheus.go
```diff
@@ -158,8 +163,13 @@ func (svc *Service) loadBaseConfig() *config.Config {
 	return &cfg
 }
 
+// AddScrapeConfigs wraps addScrapeConfigs for victoriametrics package.
```
AlekSi, line 166: Nice

services/supervisord/supervisord.go
AlekSi: Was it done?

services/supervisord/supervisord.go
```diff
@@ -533,6 +585,61 @@ stdout_logfile_backups = 3
...
+stdout_logfile_backups = 3
+redirect_stderr = true
+{{end}}
+
+{{define "vmalert"}}
+[program:vmalert]
+priority = 7
+command =
+	/usr/sbin/vmalert
+        --notifier.url="http://127.0.0.1:9093/alertmanager{{- if .AlertmanagerURL }},{{ .AlertmanagerURL }}{{- end }}"
```
AlekSi, line 616: Please pass fully prepared parameters to template to avoid `if` and especially `{{-`

services/supervisord/supervisord_test.go
```diff
@@ -115,3 +119,34 @@ func TestParseStatus(t *testing.T) {
...
+		require.Equal(t, "", params["AlertmanagerURL"])
+	})
+
+	t.Run("simple alertmanager url", func(t *testing.T) {
+		params := map[string]interface{}{}
+		err := addAlertManagerParams("https://some-alertmanager", params)
+		require.NoError(t, err)
+		require.Equal(t, "https://some-alertmanager", params["AlertmanagerURL"])
+	})
+	t.Run("extract username and password from alertmanager url", func(t *testing.T) {
```
AlekSi, line 138: Please consistently separate subtests with a blank line

services/supervisord/supervisord_test.go
```diff
@@ -115,3 +119,34 @@ func TestParseStatus(t *testing.T) {
 		assert.Equal(t, expected, parseStatus(str), "%q", str)
 	}
 }
+
+func TestAddAlertManagerParam(t *testing.T) {
+	t.Parallel()
```
AlekSi, line 124: Blank line after that one

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,362 @@
...
+	svc.l.Debugf("VictoriaMetrics: %s", b)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return errors.Errorf("expected 200, got %d", resp.StatusCode)
+	}
+
+	return nil
+}
```
AlekSi, line 180: Could it be a shared function in Prometheus package?

services/victoriametrics/victoriametrics.go
```diff
@@ -0,0 +1,362 @@
...
+
+	if resp.StatusCode == http.StatusNoContent {
+		return nil
+	}
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return errors.Errorf("%d: %s", resp.StatusCode, b)
```
AlekSi, line 205: Could it be a shared function in Prometheus package?

services/victoriametrics/vmalert.go
```diff
@@ -0,0 +1,204 @@
...
+	if resp.StatusCode == http.StatusOK {
+		return nil
+	}
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	return errors.Errorf("%d: %s", resp.StatusCode, b)
+}
```
AlekSi, line 173: Could it be a shared function in Prometheus package?

services/victoriametrics/vmalert.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return errors.Errorf("expected 200, got %d", resp.StatusCode)
+	}
+
+	svc.l.Debugf("%s", b)
+
+	return nil
+}
```
AlekSi, line 148: Could it be a shared function in Prometheus package?

services/prometheus/alerting_rules_test.go
f41gh7: rename to FormerZero

testdata/victoriametrics/prometheus.external.yml
```diff
@@ -0,0 +1,5 @@
+rule_files:
+  - "/srv/external_rules/rul1.yml"
+  - "/srv/external_rules/rule2.yml"
+global:
+  evaluation_interval: 10s
```
AlekSi, line 5: \n

utils/envvars/parser.go
```diff
@@ -116,6 +116,11 @@ func ParseEnvVars(envs []string) (envSettings *models.ChangeSettingsParams, errs
 			if envSettings.DataRetention, err = parseStringDuration(v); err != nil {
 				err = formatEnvVariableError(err, env, v)
 			}
+		case "VM_CACHE_ENABLE":
```
AlekSi, line 119: It should be ENABLE_VM_CACHE

services/victoriametrics/victoriametrics.go
f41gh7: Well, we are going to remove prometheus package in the next release. It will allow us to make less changes.

AlekSi:
```
// ScrapeTimeout - wraps scrapeTimeout and makes it public for victoriametrics package.
 func ScrapeTimeout(interval time.Duration) config.Duration
```

So we did that to make less changes, but copy&paste code the to make less changes? Why we don't expose that function like we did with ScrapeTimeout?

f41gh7: probably, this function can be unified with `vmalert`. wdyt?

f41gh7: smthng like:
```go
func reloadConfig( ctx, url, wantStatusCode) error
```

services/victoriametrics/vmalert.go
f41gh7: Same as above, Prometheus package is going to be removed.

services/victoriametrics/vmalert.go
f41gh7: Same as above, Prometheus package is going to be removed.

services/supervisord/supervisord.go
f41gh7: yep it looks simpler.

services/prometheus/prometheus.go
```diff
@@ -43,7 +43,12 @@ import (
 	"github.com/percona/pmm-managed/models"
 )
 
-const updateBatchDelay = 3 * time.Second
+const (
+	updateBatchDelay = 3 * time.Second
+	// BasePrometheusConfigPath - basic path with prometheus config,
+	// that user can mount to container.
+	BasePrometheusConfigPath = "/srv/prometheus/prometheus.base.yml"
```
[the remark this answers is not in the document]
f41gh7: for now yes, we are passing it to vmParams as config source.

models/victoriametrics.go
```diff
@@ -0,0 +1,98 @@
...
+
+// loadVMAlertParams - load params and converts it to vmalert flags.
+func (vmp *VictoriaMetricsParams) loadVMAlertParams() error {
+	buf, err := ioutil.ReadFile(vmp.BaseConfigPath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return errors.Wrap(err, "cannot read baseConfigPath for VMAlertParams")
+		}
+
+		return nil
```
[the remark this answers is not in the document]
f41gh7: its `/srv/prometheus/prometheus.base.yml`, that user is able to mount inside container.

 After removing prometheus we cannot read from `/etc/prometheus.yml` and `/etc/victoriametrics.yml` hasn't such config entries.

models/victoriametrics.go
f41gh7: it was a discussion, there is no any cache.

models/victoriametrics.go
```diff
@@ -0,0 +1,98 @@
...
+	// enables victoriametrics services at supervisor config.
+	vmTestEnableEnv = "PERCONA_TEST_ENABLE_VM"
+)
+
+// VictoriaMetricsParams - defines flags and settings for victoriametrics.
+type VictoriaMetricsParams struct {
+	// VMAlertFlags additional flags for VMAlert.
+	VMAlertFlags []string
+	// Enables VictoriaMetrics.
+	Enabled bool
```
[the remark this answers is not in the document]
f41gh7: Comment is invalid.

models/settings_helpers.go
```diff
@@ -152,6 +154,8 @@ func UpdateSettings(q reform.DBTX, params *ChangeSettingsParams) (*Settings, err
 		settings.SaaS.Email = params.Email
 	}
 
+	settings.VictoriaMetrics.CacheEnable = params.VMCacheEnable
```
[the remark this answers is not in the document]
f41gh7: Yep, we want to disable it by default.

models/victoriametrics_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestVictoriaMetricsParams(t *testing.T) {
+	testConf := "../testdata/victoriametrics/prometheus.external.yml"
```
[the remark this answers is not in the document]
f41gh7: copy paste from Prometheus package tests.

models/settings_helpers.go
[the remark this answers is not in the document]
AlekSi: No, my point is that ChangeSettings API can be called with just `{}` (empty JSON object) – it is a supported way to re-read base files, regenerate configuration, etc. With that change, that will always disable cache. Again, that why we have DisableSTT / EnableSTT as two separate booleans instead of just one.

models/victoriametrics.go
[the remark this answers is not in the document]
AlekSi: Then it should be updated :)

models/victoriametrics.go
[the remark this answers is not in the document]
AlekSi: Ok, LGTM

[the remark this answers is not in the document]
AlekSi: (but please add comment explaining why we do that)

services/victoriametrics/victoriametrics.go
AlekSi: Sounds good.

Also, that's something I would not insist on if it would take a lot of time

models/victoriametrics.go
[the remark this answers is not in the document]
f41gh7: done

models/victoriametrics_test.go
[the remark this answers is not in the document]
f41gh7: i mean, it contains external configuration, that user mounted into container. And it tests extraction of `rules` and `evaluationInterval` params

services/supervisord/supervisord_test.go
f41gh7: done

services/supervisord/supervisord_test.go
f41gh7: done

services/victoriametrics/victoriametrics.go
f41gh7: As far as i remember, we discussed it - changes to prometheus config generation shouldn't impact other people work and be minimal. That's why neither ScrapeTimeout nor ScrapeConfig generation were copy-pasted.
 
 And also, wrapping isReady and reload funcs for prometheus package leads to additional changes.

 I've tried to rewrite reloadConfig func and i don't like how it looks.. Lets refactor it next time.

utils/envvars/parser.go
f41gh7: renamed

models/victoriametrics.go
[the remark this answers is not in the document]
f41gh7: added comment

utils/envvars/parser.go
f41gh7: Also, renamed corresponding params.

models/settings.go
```diff
@@ -49,6 +49,10 @@ type Settings struct {
 	// not url.URL to keep username and password
 	AlertManagerURL string `json:"alert_manager_url"`
 
+	VictoriaMetrics struct {
+		CacheEnable bool `json:"cache_enable"`
```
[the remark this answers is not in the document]
f41gh7: changed

models/settings_helpers.go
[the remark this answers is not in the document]
f41gh7: Thanks, get it. Changed logic - added DisableVMCache param.

models/victoriametrics.go
```diff
@@ -0,0 +1,98 @@
...
+	"os"
+	"strconv"
+
+	config "github.com/percona/promconfig"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+
```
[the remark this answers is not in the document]
f41gh7: done

models/victoriametrics_test.go
```diff
@@ -0,0 +1,52 @@
...
+	t.Run("read non exist baseConfigFile", func(t *testing.T) {
+		_ = os.Setenv(vmTestEnableEnv, "true")
```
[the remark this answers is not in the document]
f41gh7: Removed Setenv.

AlekSi (approved):

PR #466, opened 2020-09-17
SAAS-321 Merge PERCONA_TEST_XXX_HOST environment variables

yurkovychv:
created a shared func to validate and get SAAS host.

[SAAS-321](https://jira.percona.com/browse/SAAS-321)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1121

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


utils/envvars/parser.go
```diff
@@ -135,6 +145,37 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+	case govalidator.IsURL(os.Getenv(envSAASTelemetryHost)):
+		h := os.Getenv(envSAASTelemetryHost)
+		logrus.Warnf("SAAS host changed to %s.", h)
+
+		return h
+	case govalidator.IsURL(os.Getenv(envSAASHost)):
+		h := os.Getenv(envSAASHost)
+		logrus.Warnf("SAAS host changed to %s.", h)
+
+		return h
```
artemgavrilov, line 171: It doesn't provide backward compatibility, if I will specify `PERCONA_TEST_AUTH_HOST` it will affect all SaaS related subsystems, not only `services/platform/`. 

As I see the ticket description changed, so if we want to keep old variables we should check that used only new variable or only old variables, not both.

utils/envvars/parser.go
```diff
@@ -135,6 +145,37 @@ func parseStringDuration(value string) (time.Duration, error) {
 	return d, nil
 }
 
+// GetSAASHost validates SAAS host env variables.
+// returns it if it's valid, otherwise returns defaultSAASHost.
+func GetSAASHost() string {
+	switch {
+	case govalidator.IsURL(os.Getenv(envSAASAuthHost)):
```
artemgavrilov, line 152: If `PERCONA_TEST_AUTH_HOST` contains not-url, then it is error, and user should be notified. Probably with app crash.

artemgavrilov (changes requested):

utils/envvars/parser.go
yurkovychv: So in all these cases I should set `PERCONA_TEST_SAAS_HOST` and return it?

AlekSi:
Not sure why you keep reviewing draft PR…

utils/envvars/parser.go
artemgavrilov: `Use old environment variables if they are defined.`

If understand that line form ticket correctly, then you should return old variable value to the specific service if it is defined.

For example `services/platform/` should get value from `PERCONA_TEST_AUTH_HOST`, `services/checks` from `PERCONA_TEST_CHECKS_HOST` etc
Or if `PERCONA_TEST_SAAS_HOST` specified then every service should receive value form that env.

yurkovychv: updated GetSAASHost function accordingly to @AlekSi clarification

utils/envvars/parser.go
```diff
@@ -135,6 +142,33 @@ func parseStringDuration(value string) (time.Duration, error) {
 	return d, nil
 }
 
+// GetSAASHost returns SAAS host env variable value.
+// if it's valid, otherwise returns defaultSAASHost.
+func GetSAASHost(env string) string {
+	if v, b := os.LookupEnv(envSAASHost); b {
```
artemgavrilov, line 148: In go it's common to name bool flags representing success of getting some value from some container `ok`.

So `b` -> `ok` will follow gostyle)

utils/envvars/parser.go
```diff
@@ -135,6 +142,33 @@ func parseStringDuration(value string) (time.Duration, error) {
 	return d, nil
 }
 
+// GetSAASHost returns SAAS host env variable value.
+// if it's valid, otherwise returns defaultSAASHost.
```
artemgavrilov, line 146: The sentence should start from capital letter.

utils/envvars/parser.go
```diff
@@ -135,6 +142,33 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+func GetSAASHost(env string) string {
+	if v, b := os.LookupEnv(envSAASHost); b {
+		if u := govalidator.IsURL(v); u {
+			logrus.Warnf("SAAS host changed to %s.", v)
+
+			return v
+		}
+		logrus.Warnf("environment variable %s has invalid format %s.", env, v)
```
artemgavrilov, line 154: Here we will fallback to the `defaultSAASHost`. I'm not sure that it is a good idea, better to return error that will cause app exit. It will explicitly show that something configured wrong.

utils/envvars/parser.go
```diff
@@ -135,6 +142,33 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+	}
+
+	if v, b := os.LookupEnv(env); b {
```
artemgavrilov, line 157: Same, `b` -> `ok`

utils/envvars/parser.go
```diff
@@ -135,6 +142,33 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+		if u := govalidator.IsURL(v); u {
+			logrus.Warnf("environment variable %s WILL BE REMOVED SOON, please use %s instead", env, envSAASHost)
```
artemgavrilov, line 159: Log messages should start with a capital letter and end with a dot.

artemgavrilov (changes requested):

utils/envvars/parser.go
yurkovychv: Got it, should I throw Fatal here?

artemgavrilov: I think better to return error `func GetSAASHost(env string) (string, error)`, and then service also will return error, then it will be logged in `main.go` with `Fatal`

utils/envvars/parser.go
yurkovychv: oh, forgot about that one, thanks. Done

utils/envvars/parser.go
yurkovychv: Done, thanks

utils/envvars/parser.go
yurkovychv: done

utils/envvars/parser.go
yurkovychv: done thanks

utils/envvars/parser.go
yurkovychv: updated

services/telemetry/telemetry.go
```diff
@@ -83,30 +83,34 @@ type Service struct {
...
 	}
 
 	s.sDistributionMethod, s.tDistributionMethod, s.os = getDistributionMethodAndOS(l)
 
+	u, err := envvars.GetSAASHost(envV2Host)
+	if err != nil {
+		return s, err
+	}
+
+	s.v2Host = u
```
artemgavrilov, line 107: It can be a little bit simpler.
```suggestion
	host, err := envvars.GetSAASHost(envV2Host)
	if err != nil {
		return s, err
	}

	s := &Service{
		db :        db,
		v2Host: host,
		....
```

yurkovychv: as I see `return s, err` will not work in this case

artemgavrilov: Yes, it should be `return nil, err`. It's also common approach in go, to return empty results in case of error.

main.go
```diff
@@ -535,16 +535,27 @@ func main() {
 
 	logs := supervisord.NewLogs(version.FullInfo(), pmmUpdateCheck)
 	supervisord := supervisord.New(*supervisordConfigDirF, pmmUpdateCheck)
-	telemetry := telemetry.NewService(db, version.Version)
+
+	telemetry, err := telemetry.NewService(db, version.Version)
+	if err != nil {
+		l.Fatalf("could not create telemetry service. %s", err)
```
AlekSi, line 541:
```suggestion
		l.Fatalf("Could not create telemetry service: %s", err)
```

The same below

services/checks/checks.go
```diff
@@ -125,7 +124,7 @@ func New(agentsRegistry agentsRegistry, alertmanagerService alertmanagerService,
 		alertsRegistry:      newRegistry(resolveTimeoutFactor * resendInterval),
 
 		l:               l,
-		host:            defaultHost,
+		host:            "",
```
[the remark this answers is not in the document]
AlekSi: Just omit it, it will be initialised by the zero value automatically: https://golang.org/ref/spec#The_zero_value

services/telemetry/telemetry.go
yurkovychv: Done

utils/envvars/parser.go
```diff
@@ -141,6 +148,35 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+			logrus.Warnf("SAAS host changed to %q.", v)
+
+			return v, nil
+		}
+
+		return v, fmt.Errorf("environment variable %q has invalid format %q. Expected {host}:{port}", env, v)
+	}
+
+	if v, ok := os.LookupEnv(env); ok {
+		if ok := govalidator.IsURL(v); ok {
```
AlekSi, line 165: But that's not an URL, it is just `host:port`. Use `net.SplitHostPort` and check that port is present, and there is not error.

AlekSi (changes requested):

utils/envvars/parser.go
yurkovychv: done, thanks

services/checks/checks.go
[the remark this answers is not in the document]
yurkovychv: updated according to Artem's suggestion

utils/envvars/parser.go
```diff
@@ -141,6 +148,44 @@ func parseStringDuration(value string) (time.Duration, error) {
 	return d, nil
 }
 
+// GetSAASHost returns SAAS host env variable value if it's valid.
+// Otherwise returns defaultSAASHost.
+func GetSAASHost(env string) (string, error) {
+	if v, ok := os.LookupEnv(envSAASHost); ok {
+		h, p, err := net.SplitHostPort(v)
+		if err != nil {
+			return v, err
```
artemgavrilov, line 157:
```suggestion
			return nil, err
```

utils/envvars/parser.go
```diff
@@ -141,6 +148,44 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+		}
+		if h != "" && p != "" {
+			logrus.Warnf("SAAS host changed to %q.", v)
+
+			return v, nil
+		}
+
+		return v, fmt.Errorf("environment variable %q has invalid format %q. Expected {host}:{port}", env, v)
```
artemgavrilov, line 165:
```suggestion
		return nil, fmt.Errorf("environment variable %q has invalid format %q. Expected {host}:{port}", env, v)
```

utils/envvars/parser.go
```diff
@@ -141,6 +148,44 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+		}
+
+		if h != "" && p != "" {
+			logrus.Warnf("Environment variable %q WILL BE REMOVED SOON, please use %q instead.", env, envSAASHost)
+			logrus.Warnf("SAAS host changed to %q.", v)
+
+			return v, nil
+		}
+
+		return v, fmt.Errorf("environment variable %q has invalid format %q. Expected {host}:{port}", env, v)
```
artemgavrilov, line 181:
```suggestion
		return nil, fmt.Errorf("environment variable %q has invalid format %q. Expected {host}:{port}", env, v)
```

utils/envvars/parser.go
```diff
@@ -141,6 +148,44 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+	}
+
+	if v, ok := os.LookupEnv(env); ok {
+		h, p, err := net.SplitHostPort(v)
+		if err != nil {
+			return v, err
```
artemgavrilov, line 171:
```suggestion
			return nil, err
```

artemgavrilov (approved):
LGTM with comments

utils/envvars/parser.go
artemgavrilov, line 165: Just stylistic simplification, both conditions check for the error case

```suggestion
		h, p, err := net.SplitHostPort(v)
		if err != nil {
			return v, err
		}
		if h == "" || p == "" {
			return v, fmt.Errorf("environment variable %q has invalid format %q. Expected {host}:{port}", env, v)
		}
		
		logrus.Warnf("SAAS host changed to %q.", v)
		return v, nil
```

utils/envvars/parser.go
yurkovychv: As I see then I should change the return type to `*string` is that correct?

AlekSi: Totally idiomatic way would be `return "", err` so the reader does not have to think too much about if v is empty or not there (cases when both returned value and error are set are rare, but exist: https://golang.org/pkg/io/#Reader)

But that's not that important there

AlekSi: What _is_ important that that we duplicated some logic logic like calling SplitHostPort, etc. See https://github.com/percona/pmm-managed/pull/466/commits/1f09b8baa2a655350d335c141314e0e25f8d2ff8

AlekSi (approved):

utils/envvars/parser.go
```diff
@@ -141,6 +148,38 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+	if v == "" {
+		name, v = fallbackEnv, os.Getenv(fallbackEnv)
+		if v != "" {
+			logrus.Warnf("Environment variable %q WILL BE REMOVED SOON, please use %q instead.", fallbackEnv, envSaaSHost)
+		}
+	}
+
+	if v == "" {
+		logrus.Debugf("Using default SaaS host %q.", defaultSaaSHost)
+		return defaultSaaSHost, nil
```
AlekSi, line 164: @BupycHuk @artemgavrilov I disabled nlreturn linter globally because it wanted add an empty line before returns, and that's, IMHO, worse

utils/envvars/parser.go
yurkovychv: thank you

utils/envvars/parser.go
yurkovychv: this piece of code was updated by Aleksi, thank you

AlekSi (changes requested):
As discussed, we should log "WILL BE REMOVED SOON" message in the entrypoint

services/checks/checks.go
```diff
@@ -42,14 +42,15 @@ import (
...
 )
 
 const (
-	defaultHost            = "check.percona.com:443"
 	defaultRestartInterval = 24 * time.Hour
 	defaultStartDelay      = time.Minute
 
 	// Environment variables that affect checks service; only for testing.
+	// FIXME remove https://jira.percona.com/browse/SAAS-360
 	envHost            = "PERCONA_TEST_CHECKS_HOST"
```
AlekSi, line 54:
```suggestion
	envHost            = "PERCONA_TEST_CHECKS_HOST" // FIXME remove https://jira.percona.com/browse/SAAS-360
```

To make this a comment for this particular line, not the whole section. Please do the same in other files

yurkovychv: Done

AlekSi (approved):

PR #469, opened 2020-09-21
PMM-6566 Start DBaaS on PMM server

JiriCtvrtka:
[PMM-6566](https://jira.percona.com/browse/PMM-6566)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1117

---
- [x] Tests passed.
- [x] Feature build pass.
- [x] (Re)requested review.
- [ ] Update API dependency.


askomorokhov (approved):

BupycHuk (changes requested):

models/settings_helpers.go
```diff
@@ -44,6 +45,8 @@ func GetSettings(q reform.DBTX) (*Settings, error) {
 	}
 
 	s.fillDefaults()
+	s.DBaaS.Enabled = os.Getenv("PERCONA_TEST_DBAAS") == "1"
```
BupycHuk, line 48: Nope, we already set this value in `envvars` package, just use it in `supervisord`

services/supervisord/supervisord_test.go
```diff
@@ -45,6 +46,10 @@ func TestConfig(t *testing.T) {
 			continue
 		}
 
+		if tmpl.Name() == "dbaas-controller" && os.Getenv("PERCONA_TEST_DBAAS") != "1" {
+			continue
```
BupycHuk, line 50: let's check both cases in a separate test.

BupycHuk: btw, you can set the value of Dbaas.Enabled in settings above.

BupycHuk (changes requested):

services/supervisord/supervisord_test.go
```diff
@@ -56,6 +58,45 @@ func TestConfig(t *testing.T) {
...
+
+	for _, enabled := range tests {
+		st := models.Settings{
+			DBaaS: struct {
+				Enabled bool `json:"enabled"`
+			}{
+				Enabled: enabled,
+			},
+		}
+		if !st.DBaaS.Enabled {
```
BupycHuk, line 88: What's the reason to add `false` case and skip it?

JiriCtvrtka: Added expected result for disabled into another file.

services/supervisord/supervisord.go
```diff
@@ -486,6 +487,23 @@ func (s *Service) UpdateConfiguration(settings *models.Settings) error {
 // once we start generating it. See alertmanager service.
 
 var templates = template.Must(template.New("").Option("missingkey=error").Parse(`
+{{define "dbaas-controller"}}
+{{if .PerconaTestDbaas -}}
```
BupycHuk, line 491: I'm not sure that it's a good idea to do it this way, maybe it's better to pass value to autostart and autorestart values?

BupycHuk: btw, why these fields are missed?

JiriCtvrtka: Sorry I didnt realize that it is mandatory. So fields added and value passed into these fields.

BupycHuk (approved):

services/supervisord/supervisord.go
```diff
@@ -486,6 +487,23 @@ func (s *Service) UpdateConfiguration(settings *models.Settings) error {
...
+{{define "dbaas-controller"}}
+[program:dbaas-controller]
+priority = 7
+command = /usr/sbin/dbaas-controller
+user = pmm
+autorestart = {{ .PerconaTestDbaas }}
+autostart = {{ .PerconaTestDbaas }}
+startretries = 10
+startsecs = 1
+stopsignal = INT
```
AlekSi, line 499: Any particular reason we use SIGINT instead of SIGTERM?

services/supervisord/supervisord.go
```diff
@@ -486,6 +487,23 @@ func (s *Service) UpdateConfiguration(settings *models.Settings) error {
 // once we start generating it. See alertmanager service.
 
 var templates = template.Must(template.New("").Option("missingkey=error").Parse(`
+{{define "dbaas-controller"}}
+[program:dbaas-controller]
+priority = 7
```
AlekSi, line 492: Any particular reason we use the same priority as Prometheus?

JiriCtvrtka: Please tell me what priority I should set. I dont know how much important it should be. Thanks.

services/supervisord/supervisord.go
JiriCtvrtka: No particular reason. If is SIGTERM enough here I can change it.

AlekSi: So can we use the same signal as all blocks below for consistency?

services/supervisord/supervisord.go
AlekSi: It defines the order of services startup. If priority is the same, services will be started at the same time. If we don't have a reason to do that, let's use a separate priority. Prometheus and Alertmanager use 7 and 8 below, so – 6?

services/supervisord/supervisord.go
JiriCtvrtka: Sure. I will change it together with priority.

services/supervisord/supervisord.go
JiriCtvrtka: @AlekSi I changed it to 6 because here is space between dash, but if you really thought minus 6 then I will change it again. Thanks.

AlekSi: No, that's dash. I probably used Russian punctuation rules :)

PR #487, opened 2020-10-13
SAAS-280 Add ip_is_private function

palash25:
[SAAS-280](https://jira.percona.com/browse/SAAS-280)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1175

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


palash25:
the CI is not running for some reason, I tried force pushing several times but it didn't trigger the build

AlekSi (changes requested):
Please resolve conflicts, that should trigger CI

AlekSi:
And please create a FB

services/checks/funcs.go
```diff
@@ -80,3 +82,48 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+		return nil, errors.Errorf("expected string argument, got %[1]T (%[1]v)", args[0])
+	}
+
+	ipAddress := net.ParseIP(ip)
+
+	for _, b := range privateAddressBlocks {
+		_, network, err := net.ParseCIDR(b)
+		if err != nil {
+			return nil, errors.Errorf("unable to parse CIDR: %s, reason: %s", b, err)
+		}
```
artemgavrilov, line 121: Can we make it before that function invocation? I don't know, maybe even in `init` or somewhere else?

services/checks/funcs.go
```diff
@@ -80,3 +82,48 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+	if l := len(args); l != 1 {
+		return nil, errors.Errorf("expected 1 argument, got %d", l)
+	}
+
+	ip, ok := args[0].(string)
+	if !ok {
+		return nil, errors.Errorf("expected string argument, got %[1]T (%[1]v)", args[0])
+	}
+
+	ipAddress := net.ParseIP(ip)
```
artemgavrilov, line 115: If we pass not an ip address, for example string "test" as argument, that function returns `false`. But in fact passed argument if wrong at all. Should we check it?

services/checks/funcs_test.go
```diff
@@ -96,3 +96,112 @@ Traceback (most recent call last):
...
+  <builtin>: in ip_is_private
+	`) + "\n",
+			result: nil,
+		},
+		{
+			name: "invalid arg",
+			script: strings.TrimSpace(`
+def check_context(rows, context):
+    func = context.get("ip_is_private", fail)
+    print("result =", func(1))
```
artemgavrilov, line 136: How about `print("result =", func("not an IP address"))` ?

services/checks/funcs.go
palash25: we could add a check, if `ParseIP` returns  nil in such cases we return an error saying "invalid ip"

services/checks/funcs_test.go
palash25: no that would be a string argument and this test case is to check for the case in which we get a wrong type of argument, if I use string here too it reduces the coverage and defeats the purpose of this test case

services/checks/funcs.go
artemgavrilov: Yep, I think to have an error is more clear in that case

services/checks/funcs_test.go
artemgavrilov: Yes, I mean how about another case, when type is correct (strinng) but content is wrong)

docker-compose.yml
```diff
@@ -15,6 +15,9 @@ services:
       - TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}
       - REVIEWDOG_GITHUB_API_TOKEN=${REVIEWDOG_GITHUB_API_TOKEN}
       - PERCONA_TEST_DBAAS=${PERCONA_TEST_DBAAS:-0}
+      - PMM_DEBUG=1
```
palash25, line 18: this wasn't supposed to be committed, will remove these changes

services/checks/funcs.go
palash25: is there a reason why you want it to be outside the function?

services/checks/funcs.go
```diff
@@ -80,3 +82,52 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
 	}
 	return p.String(), nil
 }
+
+// GetAdditionalContext returns additional functions to be used in check scripts.
+func GetAdditionalContext() map[string]starlark.GoFunc {
+	return map[string]starlark.GoFunc{
+		"ip_is_private": ipIsPrivate,
```
AlekSi, line 89: Let's add our two predeclared functions there too. That way all checks can use only context for everything

services/checks/funcs.go
AlekSi: To avoid parsing what is essentially constants (if Go had full contents support) on every function call. I'm with @artemgavrilov there. Let's define privateAddressBlocks as `[]*net.IPNet` and fill it in `init()`.

services/checks/funcs_test.go
```diff
@@ -96,3 +96,132 @@ Traceback (most recent call last):
 	}}
 	assert.Equal(t, expected, res)
 }
+
+func TestAdditionalContext(t *testing.T) {
+	funcs, err := GetFuncsForVersion(1)
```
AlekSi, line 101: predeclaredFuncs

services/checks/funcs_test.go
```diff
@@ -96,3 +96,132 @@ Traceback (most recent call last):
...
+		name   string
+		script string
+		err    string
+		result []check.Result
+	}{
+		{
+			name: "too many args",
+			script: strings.TrimSpace(`
+def check_context(rows, context):
+    func = context.get("ip_is_private", fail)
```
AlekSi, line 115: Let's use the same form as we will use everywhere:
```python
ip_is_private = context.get("ip_is_private", fail)
```

That should even work for predeclared functions

services/checks/funcs.go
```diff
@@ -80,3 +82,52 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+var privateAddressBlocks = []string{
+	// private blocks, see https://tools.ietf.org/html/rfc1918
+	"10.0.0.0/8",
+	"172.16.0.0/12",
+	"192.168.0.0/16",
+	// link-local block, see https://tools.ietf.org/html/rfc3927
+	"169.254.0.0/16",
+	// loop-back block, see https://tools.ietf.org/html/rfc5735
+	"127.0.0.0/8",
+}
```
AlekSi, line 102: What about IPv6?

AlekSi (changes requested):
We should also test with IPv6

services/checks/funcs.go
AlekSi: ⚠️ The only problem is that this error can't be handled in the script itself, and that might be useful. So let's log error using environment's debug logger and return `nil`/`None` in that case. That will require to change that function to a method.

I updated ticket description to mention that,

AlekSi:
Another things that is not covered, I think (or not tested at least): `ip_is_private` should work for networks too. `ip_is_private("10.0.0.0/8")` should return true.

services/checks/funcs.go
```diff
@@ -80,3 +85,79 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+		// private blocks, see https://tools.ietf.org/html/rfc1918
+		"10.0.0.0/8",
+		"172.16.0.0/12",
+		"192.168.0.0/16",
+		// shared address space range, see https://tools.ietf.org/html/rfc6598#section-7
+		"100.64.0.0/10",
+		// IETF protocol assignments range, see https://tools.ietf.org/html/rfc5735#section-4
+		"192.0.0.0/24",
+		// see https://tools.ietf.org/html/rfc5735#section-4
+		"198.18.0.0/15",
```
palash25, line 145: added 3 more ipv4 address ranges that are classified as private networks https://en.wikipedia.org/wiki/Reserved_IP_addresses

had missed these the first time

services/checks/funcs_test.go
```diff
@@ -96,3 +96,189 @@ Traceback (most recent call last):
...
+		{
+			name: "valid ipv6 argument",
+			script: strings.TrimSpace(`
+def check_context(rows, context):
+    func = context.get("ip_is_private", fail)
+
+    return [{
+        "summary": "IP Address Check",
+		"severity": "warning",
+		"description": "is_private: {}".format(func("0:0:0:0:0:0:0:1"))
```
palash25, line 199: test for ipv6 address added

services/checks/funcs_test.go
```diff
@@ -96,3 +96,189 @@ Traceback (most recent call last):
...
+	`),
+			err: "",
+			result: []check.Result{{
+				Summary:     "IP Address Check",
+				Severity:    check.Warning,
+				Description: "is_private: False",
+			}},
+		},
+		{
+			name: "private network",
```
palash25, line 229: tests for networks added

services/checks/funcs.go
```diff
@@ -80,3 +85,79 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+		// IPv6 Unique Local Address prefix analogous to IPv4 private networks, see https://tools.ietf.org/html/rfc4193#section-3
+		"fc00::/7",
+		// IPv6 link-local prefix, see https://tools.ietf.org/html/rfc4291#section-2.4
+		"fe80::/10",
+		// IPv6 loop-block prefix, see https://tools.ietf.org/html/rfc4291#section-2.4
+		"::1/128",
+	}
+
+	for _, b := range privateAddressBlocks {
+		_, network, _ := net.ParseCIDR(b)
```
palash25, line 160: ignoring `err` here since all the cidr provided by us are valid and won't return any error and also because `init` can't handle errors

services/checks/funcs.go
palash25: i can log the error instead of returning it, but i can't understand this part

> That will require to change that function to a method.

which function are you talking about `ipIsPrivate` and change it to a method over what type?

AlekSi: What logger you are going to use?

services/checks/funcs.go
```diff
@@ -80,3 +85,80 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+
+	ipAddress := net.ParseIP(ip)
+	if ipAddress == nil {
+		// check if string was in CIDR notation
+		_, net, err := net.ParseCIDR(ip)
+		if err != nil {
+			return nil, errors.Errorf("invalid ip address: %s", ip)
+		}
+		for _, network := range privateNetworks {
+			if reflect.DeepEqual(net, network) {
```
AlekSi, line 118: I'm pretty sure that DeepEqual will not work there. For example, both `10.0.0.0/8` and `10.0.0.0/9` are not public, because `10.0.0.0/8` is a superset of `10.0.0.0/9`

services/checks/funcs.go
AlekSi: We probably can link straight to Wikipedia instead all those RFCs :)

Will it be easier to use net.IP method like IsLoopback there?

services/checks/funcs.go
AlekSi: > provided by us are valid

How do we know that?

> also because init can't handle errors

panic(err) :)

services/checks/funcs.go
palash25: you were right, changed the logic there and also changed a test case to check for `10.0.0.0/9` returns true now

services/checks/funcs.go
palash25: added panic, although I don't like the idea of having a package that can panic

services/checks/funcs.go
palash25: > We probably can link straight to Wikipedia instead all those RFCs :)

done 

> Will it be easier to use net.IP method like IsLoopback there?

I would rather check for all the address ranges the same way rather than using a special method just for the loopback blocks. If that is what you mean by this comment.

services/checks/funcs.go
palash25: i thought you meant to log it with just a debug log level but I think i misunderstood that you mean "some logger that we already use somewhere"  by this `environment's debug logger` to be used here??? I am not sure :confused:

palash25:
made most of the changes except for one

services/checks/funcs.go
AlekSi: Doh, I renamed it to `print` and forgot about it :)

https://github.com/percona-platform/platform/blob/27c08746e97b7af95712d5361053b968718fc4d2/pkg/starlark/starlark.go#L130-L133

We should use that PrintFunc. But accessing it would require something like:
```go
func (env *Env) ipIsPrivate …
```

But we can't do that there. So yeah, let's use logrus with Error level

palash25: done.

artemgavrilov (approved):

services/checks/funcs.go
```diff
@@ -80,3 +85,79 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+// GetAdditionalContext returns additional functions to be used in check scripts.
+func GetAdditionalContext() map[string]starlark.GoFunc {
+	return map[string]starlark.GoFunc{
+		"ip_is_private":      ipIsPrivate,
+		"parse_version":      parseVersion,
+		"format_version_num": formatVersionNum,
+	}
+}
+
+// ipIsPrivate accepts a single string argument (IP address) and
```
AlekSi, line 98: > (IP address)

Not only it

services/checks/funcs.go
```diff
@@ -80,3 +85,79 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+	if !ok {
+		return nil, errors.Errorf("expected string argument, got %[1]T (%[1]v)", args[0])
+	}
+
+	ipAddress := net.ParseIP(ip)
+	if ipAddress == nil {
+		// check if string was in CIDR notation
+		_, net, err := net.ParseCIDR(ip)
+		if err != nil {
+			log.Errorf("invalid ip/network address: %s", ip)
```
AlekSi, line 117: %q to see spaces, etc

services/checks/funcs.go
```diff
@@ -80,3 +85,79 @@ func formatVersionNum(args ...interface{}) (interface{}, error) {
...
+			return nil, nil
+		}
+		for _, network := range privateNetworks {
+			// check if the two networks intersect
+			// https://stackoverflow.com/questions/34729158/how-to-detect-if-two-golang-net-ipnet-objects-intersect/34729915#34729915
+			if net.Contains(network.IP) || network.Contains(net.IP) {
```
AlekSi, line 123: Nice!

services/checks/funcs.go
AlekSi: There are several methods: IsLoopback, IsLinkLocalMulticast, IsLinkLocalUnicast, etc.
I'm okay with not using them, but please check they all are covered correctly

AlekSi (commented):
LGTM with minor comments. Please resolve them, update FB, and move to QA. Please poke @BupycHuk if you need help merging that PR.

@BupycHuk PTAL

services/checks/funcs.go
```diff
@@ -146,6 +146,7 @@ func init() {
 		"192.0.0.0/24",
 		"198.18.0.0/15",
 		"169.254.0.0/16",
+		"224.0. 0.0/24",
```
palash25: link-local multicast added

services/checks/funcs.go
palash25: added multicast interface-local and multicast link-local

palash25:
made the changes, will move the ticket to qa once the CI is green

services/checks/funcs.go
palash25: fixed the space

AlekSi (approved):

artemgavrilov (approved):

BupycHuk (approved):

PR #514, opened 2020-11-09
PMM-6870 Generate alertmanager config

palash25:
[PMM-6870](https://jira.percona.com/browse/PMM-6870)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1229

---
- [x] Tests passed.
- [x] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/alertmanager/alertmanager.go
```diff
@@ -130,6 +131,33 @@ receivers:
...
 	}
+
+	svc.l.Infof("checking %s content", alertmanagerConfigPath)
+	content, err := ioutil.ReadFile(alertmanagerConfigPath)
+	if err != nil {
+		svc.l.Errorf("Failed to load alertmanager config %s: %s", alertmanagerConfigPath, err)
+	}
+
+	// copy the base config if `/etc/alertmanager.yml` is already present but does not have any config.
+	if string(content) == "---\n" {
```
palash25, line 142: added this check because `/etc/alertmanager.yml` is already present having `---\n` as its content when we do `make env-up`

palash25:
should I add a unit test for the `Run` method and check if the config files are generated. Was there a reason why the unit test was not added in the first place?

services/alertmanager/alertmanager.go
```diff
@@ -29,16 +29,19 @@ import (
 	"github.com/percona/pmm/api/alertmanager/amclient/alert"
 	"github.com/percona/pmm/api/alertmanager/amclient/general"
 	"github.com/percona/pmm/api/alertmanager/ammodels"
+	"github.com/percona/promconfig/alertmanager"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/reform.v1"
+	"gopkg.in/yaml.v2"
```
AlekSi, line 36: @artemgavrilov See? I wonder if we could something about that later

artemgavrilov: I expect to have some free time in this sprint, so I could try to migrate everything to `v3`.

services/alertmanager/alertmanager.go
```diff
@@ -107,10 +110,8 @@ func (svc *Service) createDataDir() {
 	}
 }
 
-// generateBaseConfig generates /srv/alertmanager/alertmanager.base.yml if it is not present.
-//
-// TODO That's a temporary measure until we start generating /etc/alertmanager.yml
-// using /srv/alertmanager/alertmanager.base.yml as a base. See supervisord config.
```
AlekSi, line 113: We should update supervisord config to use /etc/alertmanager.yml instead of /srv/alertmanager/alertmanager.base.yml

services/alertmanager/alertmanager.go
artemgavrilov: Looks like it's not a big deal

services/alertmanager/alertmanager.go
palash25: updated https://github.com/percona/pmm-managed/pull/514/commits/d9edf32bcd500a85a8769da524bdad306447d325

palash25:
looking into the travis failure

palash25:
fixed

services/alertmanager/alertmanager.go
```diff
@@ -130,6 +131,40 @@ receivers:
...
+		svc.l.Infof("checking %s content", alertmanagerConfigPath)
+		content, err = ioutil.ReadFile(alertmanagerConfigPath)
+		if err != nil {
+			svc.l.Errorf("Failed to load alertmanager config %s: %s", alertmanagerConfigPath, err)
+		}
+	}
+
+	// copy the base config if `/etc/alertmanager.yml` is not present or
+	// is already present but does not have any config.
+	if string(content) == "---\n" || os.IsNotExist(err) {
```
artemgavrilov, line 148: It makes sense to change conditions order, at first check file existence, and then file content.

services/alertmanager/alertmanager.go
```diff
@@ -130,6 +131,40 @@ receivers:
...
+		var cfg alertmanager.Config
+		buf, err := ioutil.ReadFile(alertmanagerBaseConfigPath)
+		if err != nil {
+			svc.l.Errorf("Failed to load alertmanager base config %s: %s", alertmanagerBaseConfigPath, err)
```
artemgavrilov, line 152: Shouldn't we call `return` here?

services/alertmanager/alertmanager.go
```diff
@@ -130,6 +131,40 @@ receivers:
...
+		}
+		if err := yaml.Unmarshal(buf, &cfg); err != nil {
+			svc.l.Errorf("Failed to parse alertmanager base config %s: %s.", alertmanagerBaseConfigPath, err)
```
artemgavrilov, line 155: And here?

services/alertmanager/alertmanager.go
```diff
@@ -130,6 +131,40 @@ receivers:
 		err = ioutil.WriteFile(alertmanagerBaseConfigPath, []byte(defaultBase), 0644) //nolint:gosec
 		svc.l.Infof("%s created: %v.", alertmanagerBaseConfigPath, err)
 	}
+
+	var content []byte
+	_, err = os.Stat(alertmanagerBaseConfigPath)
+	svc.l.Debugf("%s status: %v", alertmanagerBaseConfigPath, err)
```
artemgavrilov, line 137: That will be a strange log in case file exists, `/srv/alertmanager/alertmanager.base.yml status: nil`

palash25: removed, I was also using the wrong file path, changed that

services/alertmanager/alertmanager.go
palash25: kept logging the same and added naked `return`s in case of errors to keep the error handling style consistent with other methods in this file

let me know if you wanted something different like returning the error from `generateBasicConfig` and logging/handling it in the `Run` method instead

artemgavrilov (approved):

services/alertmanager/alertmanager.go
```diff
@@ -130,6 +131,47 @@ receivers:
 		err = ioutil.WriteFile(alertmanagerBaseConfigPath, []byte(defaultBase), 0644) //nolint:gosec
 		svc.l.Infof("%s created: %v.", alertmanagerBaseConfigPath, err)
 	}
+
```
BupycHuk, line 134: maybe this part should be extracted to a separate method and called right after `generateBaseConfig`?
it doesn't look like 100% depend on a code above.

palash25: moved 5f791b0

AlekSi (approved):

PR #516, opened 2020-11-10
PMM-6721 pt summary execution performance issue

yareach:
[PMM-6721](https://jira.percona.com/browse/PMM-6721)
To do: The period the pt-summary is cached for must be determined!

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.

### DIFF
--- a/services/management/grpc/actions_server.go
+++ b/services/management/grpc/actions_server.go
@@ -301,6 +301,7 @@ func (s *actionsServer) StartPTSummaryAction(ctx context.Context, psReq *managem
 		return nil, err
 	}
 
+	// If founds an old action record for the agent
 	if sAction, bFound := dicPtSummaryLastAction[agentID]; bFound {
 		// If the time since the last call is less than 30 s
 		if timeNow-sAction.timestamp < ptSummaryRefreshPeriod {
--- a/services/management/grpc/actions_server.go
+++ b/services/management/grpc/actions_server.go
@@ -18,6 +18,7 @@ package grpc
 
 import (
 	"context"
+	"time"
 
 	"github.com/percona/pmm/api/agentpb"
 	"github.com/percona/pmm/api/managementpb"
@@ -37,7 +38,20 @@ type actionsServer struct {
 	l  *logrus.Entry
 }
 
+// structure to keep id with timestamp
+type idWithTimeStamp struct {
+	id        string
+	timestamp int64
+}
+
 var pmmAgent2100 = version.MustParse("2.10.0-HEAD") // TODO: Remove HEAD later once 2.11.0 is released.
+
+// ****** TBD: Define where the period will be stored or configured
+// Period that pt_summary action in seconds is cached for. After this time the pt-summary will be refreshed again.
+const ptSummaryRefreshPeriod = 60
+
+// Dictionary that of agent_id againts a structure of action_id and timestamp
+var dicPtSummaryLastAction = make(map[string]idWithTimeStamp)
 
 // NewActionsServer creates Management Actions Server.
 func NewActionsServer(r *agents.Registry, db *reform.DB) managementpb.ActionsServer {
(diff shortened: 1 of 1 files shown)


services/management/grpc/actions_server.go
```diff
@@ -37,8 +38,21 @@ type actionsServer struct {
...
+type idWithTimeStamp struct {
+	id        string
+	timestamp int64
+}
+
 var pmmAgent2100 = version.MustParse("2.10.0-HEAD") // TODO: Remove HEAD later once 2.11.0 is released.
 
+// ****** TBD: Define where the period will be stored or configured
+// Period that pt_summary action in seconds is cached for. After this time the pt-summary will be refreshed again.
+const ptSummaryRefreshPeriod = 60
```
BupycHuk, line 51: let's make it time.Duration

services/management/grpc/actions_server.go
```diff
@@ -37,8 +38,21 @@ type actionsServer struct {
...
+
+// Dictionary that of agent_id againts a structure of action_id and timestamp
+var dicPtSummaryLastAction = make(map[string]idWithTimeStamp)
```
BupycHuk, line 54: can we include cleanup logic?

BupycHuk: and maybe it's better to make it field of actionsServer structure?

services/management/grpc/actions_server.go
```diff
@@ -260,38 +274,62 @@ func (s *actionsServer) StartMongoDBExplainAction(ctx context.Context, req *mana
...
-	agentID, err := models.FindPmmAgentIDToRunAction(req.PmmAgentId, agents)
+	// Gets the agent by ID
+	agentID, err := models.FindPmmAgentIDToRunAction(psReq.PmmAgentId, psAgents)
 	if err != nil {
 		return nil, err
 	}
 
-	res, err := models.CreateActionResult(s.db.Querier, agentID)
+	// If founds an old action record for the agent
+	if sAction, bFound := dicPtSummaryLastAction[agentID]; bFound {
```
BupycHuk, line 305: looks too complicated

services/management/grpc/actions_server.go
BupycHuk: And it's not thread-safe

BupycHuk (changes requested):
I would suggest to run goroutine which will remove value from dictionary after `ptSummaryRefreshPeriod`. So we will not need separate structure for results and it will simplify checking if we should run pt-summary or not.

btw: why did you rename all variables?

services/management/grpc/actions_server.go
```diff
@@ -260,38 +274,62 @@ func (s *actionsServer) StartMongoDBExplainAction(ctx context.Context, req *mana
...
 
-// StartPTSummaryAction starts pt-summary action.
+// StartPTSummaryAction starts pt-summary action. If the time since the last successful start of pt-summary action is lower than ptSummaryRefreshPeriod,
+// the response from the last action will be used. If the time is longer, the new pt-summary action will be called.
 //nolint:lll
-func (s *actionsServer) StartPTSummaryAction(ctx context.Context, req *managementpb.StartPTSummaryActionRequest) (*managementpb.StartPTSummaryActionResponse, error) {
-	agents, err := models.FindPMMAgentsRunningOnNode(s.db.Querier, req.NodeId)
+func (s *actionsServer) StartPTSummaryAction(ctx context.Context, psReq *managementpb.StartPTSummaryActionRequest) (*managementpb.StartPTSummaryActionResponse, error) {
+	// Gets current timestamp
+	timeNow := time.Now().Unix()
```
AlekSi, line 282: In general, we try to add comments describing _why_ something is done, not _what_, unless the code in question is somewhat complicated. In that particular case, it doesn't look like this comment adds anything. We avoid 90% of those comments :)

image

services/management/grpc/actions_server.go
```diff
@@ -260,38 +274,62 @@ func (s *actionsServer) StartMongoDBExplainAction(ctx context.Context, req *mana
...
+	// Gets pointers to agents running on the node
+	psAgents, err := models.FindPMMAgentsRunningOnNode(s.db.Querier, psReq.NodeId)
 	if err != nil {
 		return nil, status.Errorf(codes.NotFound, "No pmm-agent running on this node")
 	}
 
-	agents = models.FindPMMAgentsForVersion(s.l, agents, pmmAgent2100)
-	if len(agents) == 0 {
+	// Filter by the version
+	psAgents = models.FindPMMAgentsForVersion(s.l, psAgents, pmmAgent2100)
```
AlekSi, line 291: That comment is right there, in the function name

AlekSi (changes requested):
I'm afraid I can't agree with that solution in general. One of the use-cases for pt-summary is when one adds a new Service one the Node. They open the dashboard once, capture pt-summary output, then add a Service to PMM, and reload the dashboard. Fuzzy-rounded pt-summary's output is useful for comparing immediate vmstat changes, etc. Caching breaks this flow.

I don't know yet what is the right solution, but that's not it.

services/management/grpc/actions_server.go
AlekSi: For comparison, I think a comment describing _what_ we do there would be nice – as Nurlan mention, it is somewhat complicated. But that should be a single comment describing the whole block, not Go-to-English translation of each line

services/management/grpc/actions_server.go
yareach: Pretty cat. Reminds me of my sister-in-law who is crazy about cats.

AlekSi: You can say I'm crazy about cats too 😸

services/management/grpc/actions_server.go
yareach: I had to open up the function to understand it. So instead of 1 sec I spent 30. That's why I put there the comment to reduce the investigation time of the reader. Time is expensive.

AlekSi: Then we probably should find a better name for the function to be able to understand what it does from the name. Otherwise, there is high chance that people will read comments and the code in question

services/management/grpc/actions_server.go
yareach: Ok. I'll tell her to name her next cat AlekSi.

services/management/grpc/actions_server.go
AlekSi: Also, it is just a question of familiarity with the code base. We can safely assume that all Go developers know what `time.Now().Unix()` does without checking the source code, right? We assume the same about most of the PMM code from PMM developers after some time.

Then again, I don't imply that we should not add comments at all. It is just that one is not helpful there.

yareach: Yes. An here starts the process: Requirement - design - discussion - approval - implementation - documentation.

AlekSi: Not sure I understand what you mean there.

yareach: Simply said: Decide what what we want to do before we cut the patient. I feel here it is opposite. First we cut.
I was referring to process development cycle. And it would also be good to have a company coding standard document (coding rules). No judge can work without the law.

services/management/grpc/actions_server.go
yareach: To save time I suggest we have a meeting. Can you set up a meeting where the right people will be?

services/management/grpc/actions_server.go
yareach: I can make a large comment before the if block. But my experience is that people skip long texts. That's why I tend to put a one-line comment before every action. What would you prefer as a recipe to cook something: Bullet form of ingredients or a paragraph of text?

services/management/grpc/actions_server.go
AlekSi: But we _do_ have it. Please check Slack

yareach:
> I would suggest to run goroutine which will remove value from dictionary after `ptSummaryRefreshPeriod`. So we will not need separate structure for results and it will simplify checking if we should run pt-summary or not.
> 
> btw: why did you rename all variables?

I didn't rename all variables. It was not clear what was pointer or structure or simple type variable. In Go without the declaration section, one does not see the type from the text. And accessing members of a structure from pointer or variable is the same. You will not get an error. In C you use str.member or pointer->member. Not in Go so you have to be more careful. So I put 'p' for pointer and 's' for structures and 'ps' for pointer to structure.

PR #518, opened 2020-11-12
SAAS-450 Update to yaml v3 

artemgavrilov:
- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1238

- [ ] https://github.com/percona/pmm-agent/pull/208
---
- [x] Tests passed.
- [x] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/agents/rds_test.go
```diff
@@ -90,36 +90,36 @@ func TestRDSExporterConfig(t *testing.T) {
 			`config`: strings.TrimSpace(`
 ---
 instances:
-- region: us-east-1
```
BupycHuk, line 93: does this means that this format will not work?

BupycHuk: I'm woried about backward compatibility

artemgavrilov: It will work, but default formatting rules are changed. In tests we use simple assertion for strings equality, so I formated hardcoded yamls it tests.

For example SaaS returns STT checks formated with `yaml.v3` and before that PR managed unmarshalled them with `yaml.v2`

AlekSi (changes requested):
It looks like this problem is worse than we thought originally – we have custom v2 unmarshallers in several places like https://github.com/prometheus/common/blob/20c99e7aa07352b599bd0517cd5ca945f4af0407/model/labels.go#L107-L118 and https://github.com/percona/promconfig/blob/dfdbe207768d9be9256e116eab3cdd5a068a3705/duration.go#L119-L131.

I created a ticket to track the problem. Let's pause on it for now.

artemgavrilov:
> It looks like this problem is worse than we thought originally – we have custom v2 unmarshallers in several places like https://github.com/prometheus/common/blob/20c99e7aa07352b599bd0517cd5ca945f4af0407/model/labels.go#L107-L118 and https://github.com/percona/promconfig/blob/dfdbe207768d9be9256e116eab3cdd5a068a3705/duration.go#L119-L131.
> 
> I created a ticket to track the problem. Let's pause on it for now.

JFYI `v3` supports the old `Unmarshall` format, so the mentioned code fragments will work.

AlekSi:
> JFYI v3 supports the old Unmarshall format, so the mentioned code fragments will work.

Ohh. I checked the code, but probably missed that. May you show it to me?

artemgavrilov:
> > JFYI v3 supports the old Unmarshall format, so the mentioned code fragments will work.
> 
> Ohh. I checked the code, but probably missed that. May you show it to me?

Interface for v2:
https://github.com/go-yaml/yaml/blob/eeeca48fe7764f320e4870d231902bf9c1be2c08/yaml.go#L40

Checking which version of unmarshaller is present:
https://github.com/go-yaml/yaml/blob/eeeca48fe7764f320e4870d231902bf9c1be2c08/decode.go#L417-L424

AlekSi:
Great! Let's finish it then when we have a chance

AlekSi (approved):

BupycHuk (approved):

palash25 (approved):

PR #525, opened 2020-11-18
PMM-6873 Add rule template collector

palash25:
[PMM-6873](https://jira.percona.com/browse/PMM-6873)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1242

---
- [x] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [x] Update API dependency.


palash25:
sorry for the noise hit the wrong button, ignore the notification for now

services/alertmanager/alertmanager.go
```diff
@@ -18,12 +18,16 @@
...
 import (
+	"bytes"
 	"context"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"syscall"
 
+	saas "github.com/percona-platform/saas/pkg/alert"
```
artemgavrilov, line 29: It shouldn't be in a separate group, as it's external dependency

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +208,71 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
 	svc.l.Infof("%s created", alertmanagerConfigPath)
 }
 
+func (svc *Service) collectRuleTemplates() {
+	rules := make([]saas.Rule, 0)
```
artemgavrilov, line 212: Can be `var rules []saas.Rule`, but it's up to you.

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +208,71 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	svc.rules = rules
+}
+
+func getRuleTemplateFilePaths(pattern string) ([]string, error) {
+	paths, err := filepath.Glob(pattern)
+	if err != nil {
+		return nil, err
+	}
+	return paths, nil
+}
```
artemgavrilov, line 255: Why not use `filepath.Glob` directly? That wrapper doesn't add any benefits as I see, no?

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +208,71 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+
+func (svc *Service) loadRuleTemplates(file string) ([]saas.Rule, error) {
+	data, err := ioutil.ReadFile(file) //nolint:gosec
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read test rule template file")
```
artemgavrilov, line 260: It will be nice to have filen ame in the error log message.

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +208,71 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	}
+
+	// be strict about local files
+	params := &saas.ParseParams{
+		DisallowUnknownFields: true,
+		DisallowInvalidRules:  true,
+	}
+	rules, err := saas.Parse(bytes.NewReader(data), params)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse test rule template file")
```
artemgavrilov, line 270: And here too.

services/alertmanager/alertmanager_test.go
```diff
@@ -36,3 +43,21 @@ func TestAlertmanager(t *testing.T) {
 
 	require.NoError(t, svc.IsReady(context.Background()))
 }
+
+func TestCollect(t *testing.T) {
+	err := os.Setenv("PERCONA_TEST_SHIPPED_RULE_TEMPLATE_PATH", testShippedFilePath)
+	require.NoError(t, err)
+	err = os.Setenv("PERCONA_TEST_USER_DEFINED_RULE_TEMPLATE_PATH", testUserDefinedFilePath)
+	require.NoError(t, err)
```
artemgavrilov, line 51: That can affect other tests that run parallel, probably better just set struct fields directly.

services/alertmanager/alertmanager.go
artemgavrilov: Oh, I see linter warning
https://github.com/percona/pmm-managed/pull/525#discussion_r526421040

In that case, you can move that line down, where you know `shippedFilePaths` size, and preallocate `rules` with the proper length.

services/alertmanager/alertmanager.go
palash25: because I call the wrapper twice

artemgavrilov: I mean, instead of this:
```
	shippedFilePaths, err := getRuleTemplateFilePaths(svc.shippedRuleTemplatePath)
	if err != nil {
		svc.l.Errorf("Failed to get paths of template files shipped with PMM: %s.", err)
		return // keep previously loaded rules
	}
```

you can write this:
```
	shippedFilePaths, err := filepath.Glob(svc.shippedRuleTemplatePath)
	if err != nil {
		svc.l.Errorf("Failed to get paths of template files shipped with PMM: %s.", err)
		return // keep previously loaded rules
	}
```
and there will be no difference.

testdata/ia/shipped/rules.yml
```diff
@@ -0,0 +1,28 @@
+---
+rules:
+    - name: shipped_rules
+      version: 1
+      help: MySQL connections in use
```
AlekSi, line 5: @palash25 @artemgavrilov May you please work together to rename "help" to "summary" there, in platform repo, Check Service, etc?

palash25: will do

testdata/ia/shipped/rules.yml
```diff
@@ -0,0 +1,28 @@
...
+      tiers: [anonymous, registered]
+      expr: |-
+        max_over_time(mysql_global_status_threads_connected[5m]) / ignoring (job)
+        mysql_global_variables_max_connections
+        * 100
+        > [[ .threshold ]]
+      params:
+        - name: threshold
+          help: A percentage from configured maximum
```
AlekSi, line 14: @palash25 "summary" there too

services/alertmanager/alertmanager.go
```diff
@@ -43,20 +46,41 @@ const (
 
 	alertmanagerConfigPath     = "/etc/alertmanager.yml"
 	alertmanagerBaseConfigPath = "/srv/alertmanager/alertmanager.base.yml"
+
+	shippedRuleTemplatePath     = "/tmp/ia1/*.yml"
+	userDefinedRuleTemplatePath = "/tmp/ia2/*.yml"
+
+	envShippedRuleTemplatePath     = "PERCONA_TEST_SHIPPED_RULE_TEMPLATE_PATH"
+	envUserDefinedRuleTemplatePath = "PERCONA_TEST_USER_DEFINED_RULE_TEMPLATE_PATH"
```
AlekSi, line 54: There is no need for them – QA will be able to put files into normal directories for users

services/alertmanager/alertmanager.go
```diff
@@ -43,20 +46,41 @@ const (
...
 
 // Service is responsible for interactions with Alertmanager.
 type Service struct {
-	db *reform.DB
-	l  *logrus.Entry
+	db                          *reform.DB
+	l                           *logrus.Entry
+	shippedRuleTemplatePath     string
+	userDefinedRuleTemplatePath string
+	rules                       []saas.Rule
```
AlekSi, line 63: In practice, we will need to access them by names. Please change to `map[string]saas.Rule` with name in key

AlekSi (changes requested):

testdata/ia/shipped/rules.yml
palash25: have drafted PRs for both platform and checked repos, will request review once the CI is green

testdata/ia/shipped/rules.yml
palash25: yeah I know

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +207,65 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
 	svc.l.Infof("%s created", alertmanagerConfigPath)
 }
 
+// collectRuleTemplates collects IA rule templates from various sources like
+// templates shipped with PMM and defined by the users.
+func (svc *Service) collectRuleTemplates() {
+	shippedFilePaths, err := filepath.Glob(svc.shippedRuleTemplatePath)
```
artemgavrilov, line 213: Looks like proper size here should be `len(svc.shippedRuleTemplatePath + svc.userDefinedRuleTemplatePath)`

palash25: we don't need that anymore, now we use maps

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +195,68 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	if err != nil {
+		svc.l.Errorf("Failed to get paths of template files shipped with PMM: %s.", err)
+		return // keep previously loaded rules
+	}
+
+	for _, path := range shippedFilePaths {
+		rules, err := svc.loadRuleTemplates(path)
+		if err != nil {
+			svc.l.Errorf("Failed to load shipped rule template file: %s, reason: %s.", path, err)
+			return // keep previously loaded rules
```
AlekSi, line 211: Well, if one file was loaded, but another wasn't, we don't keep previously loaded rules – we already collected updateRules below.

And it seems we never remove rules?

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +195,68 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	}
+	rules, err := saas.Parse(bytes.NewReader(data), params)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse test rule template file")
+	}
+
+	return rules, nil
+}
+
+func (svc *Service) updateRules(rules []saas.Rule) {
```
AlekSi, line 254: Looks like we can remove that method – we load all files first, then set `svc.rules`

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +195,68 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
 	svc.l.Infof("%s created", alertmanagerConfigPath)
 }
 
+// collectRuleTemplates collects IA rule templates from various sources like
+// templates shipped with PMM and defined by the users.
+func (svc *Service) collectRuleTemplates() {
```
AlekSi, line 200: Let's keep Alertmanager Service for interaction with Alertmanager only. Our rule templates are not related to it.

Let's move it to Templates service after https://github.com/percona/pmm-managed/pull/527.

AlekSi (changes requested):

services/alertmanager/alertmanager.go
artemgavrilov: Actually, map also has a size)

services/alertmanager/alertmanager.go
```diff
@@ -181,6 +195,68 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse test rule template file")
+	}
+
+	return rules, nil
+}
+
+func (svc *Service) updateRules(rules []saas.Rule) {
+	for _, r := range rules {
+		svc.rules[r.Name] = r
```
artemgavrilov, line 256: So you only add to that map, and never clean it? So once loaded rule will be present in map forever?

services/management/ia/templates_service.go
```diff
@@ -17,15 +17,134 @@
...
+	userDefinedRuleTemplatePath = "/tmp/ia2/*.yml"
 )
 
+// TemplatesService is responsible for interactions with IA rule templates.
 type TemplatesService struct {
-	iav1beta1.UnimplementedTemplatesServer // TODO remove
+	l                           *logrus.Entry
+	shippedRuleTemplatePath     string
+	userDefinedRuleTemplatePath string
+	rules                       map[string]saas.Rule
```
palash25, line 41: do we expect `collectRuleTemplates` to be called concurrently? should I add a lock here?

artemgavrilov (approved):

services/management/ia/templates_service.go
```diff
@@ -17,15 +17,134 @@
...
+	userDefinedRuleTemplatePath = "/tmp/ia2/*.yml"
 )
 
+// TemplatesService is responsible for interactions with IA rule templates.
 type TemplatesService struct {
-	iav1beta1.UnimplementedTemplatesServer // TODO remove
+	l                           *logrus.Entry
+	shippedRuleTemplatePath     string
+	userDefinedRuleTemplatePath string
+	mu                          sync.Mutex
```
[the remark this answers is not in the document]
AlekSi: Linter is actually correct there: it used in `collectRuleTemplates()` that is used in `Run()` which is not used in non-test code

services/management/ia/templates_service.go
```diff
@@ -17,15 +17,134 @@
...
+	panic("Unimplemented")
+}
+
+// DeleteTemplate deletes existing, previously created via API.
+func (svc *TemplatesService) DeleteTemplate(context.Context, *iav1beta1.DeleteTemplateRequest) (*iav1beta1.DeleteTemplateResponse, error) {
+	panic("Unimplemented")
+}
+
+// Run starts collecting IA rule templates.
+func (svc *TemplatesService) Run() {
```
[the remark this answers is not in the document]
AlekSi: Well, that's true. And it is not clear what this method is supposed to do.

AlekSi (approved):

PR #534, opened 2020-11-24
Fix IA APIs registration

AlekSi:


docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       - TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}
       - REVIEWDOG_GITHUB_API_TOKEN=${REVIEWDOG_GITHUB_API_TOKEN}
       - PERCONA_TEST_DBAAS=${PERCONA_TEST_DBAAS:-0}
+      - PERCONA_TEST_IA=1
```
BupycHuk, line 18: should we set 1 here? let's give ability to override it with ENV Var

main.go
```diff
@@ -180,14 +180,16 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps) {
...
-	if enable, _ := strconv.ParseBool(os.Getenv("PERCONA_TEST_IA")); enable {
+	if enable, err := strconv.ParseBool(os.Getenv("PERCONA_TEST_IA")); err == nil && enable {
 		l.Warnf("Enabling experimental IA APIs.")
 		iav1beta1.RegisterAlertsServer(gRPCServer, ia.NewAlertsService())
 		iav1beta1.RegisterChannelsServer(gRPCServer, ia.NewChannelsService())
 		iav1beta1.RegisterRulesServer(gRPCServer, ia.NewRulesService())
 		iav1beta1.RegisterTemplatesServer(gRPCServer, ia.NewTemplatesService())
 	}
 
+	// FIXME Checking settings.DBaaS.Enabled there is not correct: that setting can be changed via API,
```
BupycHuk, line 191: we can't change it via API at the moment, so it's better to update the text that we should fix it with ability to update DBaaS.Enabled via API.

docker-compose.yml
AlekSi: Unlike DBaaS, IA does not require anything else, so there is no use for that configurability. That environment variable will be removed after 2.12 release. We have it _just_ to allow us to merge some PRs now

main.go
AlekSi: Indeed

AlekSi:
Merging to produce the first build with the first IA API. Hopefully those changes are not controversial, but still PTAL

BupycHuk (commented):
LGTM

PR #540, opened 2020-11-27
PMM-6968 Refactoring

AlekSi:


services/management/ia/channels_service.go
```diff
@@ -103,156 +103,164 @@ func (s *ChannelsService) ListChannels(ctx context.Context, request *iav1beta1.L
...
 		}
 	}
 
-	err := s.ia.ChangeChannel(req.ChannelId, params)
-	if err != nil {
-		return nil, err
+	e := s.db.InTransaction(func(tx *reform.TX) error {
+		_, err := models.ChangeChannel(tx.Querier, req.ChannelId, params)
+		return err
+	})
```
artemgavrilov, line 186: I have a question not regarding that PR, but in general. `ChangeChannel` accepts `reform.Querier`, however, it can be transactional and non-transactional. So function interface doesn't descriptive enough, is there any way to explicitly show that function expected to be run inside a transaction? Probably it's more about `reform` lib, maybe something like `reform.TxQuerier`.

artemgavrilov (approved):

services/management/ia/channels_service.go
AlekSi: The idea is that all helpers should use querier, and all API handlers should use transactions. We often skip that when we know that there is only query, but, I think we could add "all API handlers should use transaction" as a general rule

models/channel_helpers.go
```diff
@@ -287,9 +286,13 @@ func ChangeChannel(q *reform.Querier, channelID string, params *ChangeChannelPar
...
 // RemoveChannel removes notification channel with specified id.
 func RemoveChannel(q *reform.Querier, id string) error {
+	if _, err := FindChannelByID(q, id); err != nil {
+		return err
+	}
+
 	err := q.Delete(&Channel{ID: id})
 	if err != nil {
-		return errors.Wrap(err, "failed to delete notifications channel")
+		return errors.Wrap(err, "failed to delete notification channel")
```
artemgavrilov, line 295: `q.Delete` has this in docs `Method returns ErrNoRows if no rows were deleted.`. So the same behavior can be achieved without  TX. Something like that:

```
func RemoveChannel(q *reform.Querier, id string) error {
	if id == "" {
		return status.Error(codes.InvalidArgument, "Empty Channel ID.")
	}
	
	err := q.Delete(&Channel{ID: id})
	if err != nil {
		if err == reform.ErrNoRows {
			return status.Errorf(codes.NotFound, "Channel with ID %q not found.", id)
		}
		return errors.Wrap(err, "failed to delete notification channel")
	}
	return nil
}
```

No?

palash25 (approved):

services/management/ia/channels_service.go
AlekSi: cf93767acee8bdda7281d9a583ef515fe7089dbf

models/channel_helpers.go
AlekSi: Sure, but then we duplicate those checks between Find, Change and Remove

artemgavrilov: Hm, ok. PMMs DB is not a hot place, nothing bad happened because of an extra transaction)

services/management/ia/channels_service.go
artemgavrilov: This is still an implicit contract, but ok. Thanks)

models/channel_helpers.go
AlekSi: Exactly. Plus PostgreSQL's MVCC is a great help for us there – unrelated transactions are running in full speed

PR #541, opened 2020-11-27
PMM-6909 Rule templates API

artemgavrilov:
[PMM-6909](https://jira.percona.com/browse/PMM-6909)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1271

- [ ] https://github.com/Percona-Lab/pmm-api-tests/pull/147
---
- [x] Tests passed.
- [x] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/template_model.go
```diff
@@ -0,0 +1,132 @@
...
+
+type FloatParam struct {
+	HasDefault bool    `json:"has_default"`
+	Default    float32 `json:"default"`
+
+	HasMin bool    `json:"has_min"`
+	Min    float32 `json:"min"`
+
+	HaxMax bool    `json:"hax_max"`
+	Max    float32 `json:"max"`
```
artemgavrilov, line 119: @AlekSi  param structure in API doesn't match param structure in `platform`. https://github.com/percona-platform/platform/blob/main/pkg/alert/parameter.go#L10
There is no way to specify only max in `alert.Rule`, moreover validation in `platform` requires full range and non-nil value.

models/template_model.go
```diff
@@ -0,0 +1,117 @@
...
+type Template struct {
+	Name        string   `reform:"name,pk"`
+	Version     uint32   `reform:"version"`
+	Summary     string   `reform:"summary"`
+	Tiers       Tiers    `reform:"tiers"`
+	Expr        string   `reform:"expr"`
+	Params      Params   `reform:"params"`
+	For         Duration `reform:"for"`
+	Severity    string   `reform:"severity"`
+	Labels      Map      `reform:"labels"`
```
AlekSi, line 23: Please use the same approach as Agent:
* https://github.com/percona/pmm-managed/blob/85459abc452472c051f1070374d89dba2073c9be/models/agent_model.go#L74
* https://github.com/percona/pmm-managed/blob/85459abc452472c051f1070374d89dba2073c9be/models/agent_model.go#L140-L148

models/template_model.go
AlekSi: Let's ignore those fields and even remove from the model for now

services/management/ia/templates_service.go
```diff
@@ -171,19 +290,38 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
...
 					Summary: p.Summary,
-					Unit:    iav1beta1.ParamUnit_PARAM_UNIT_INVALID, // TODO
-					Type:    iav1beta1.ParamType_FLOAT,
-					Value:   nil, // TODO
+					Unit:    convertParamUnit(p.Unit),
+					Type:    convertParamType(p.Type),
+					Value: &iav1beta1.TemplateParam_Float{
+						Float: &iav1beta1.TemplateFloatParam{
+							HasDefault: true,           // TODO remove or fill with valid value.
+							Default:    float32(value), // TODO eliminate conversion.
```
artemgavrilov, line 318: In API values for `Float` have type `float32`, but in `saas.Rule` it's `float64`

services/management/ia/templates_service.go
```diff
@@ -203,17 +341,80 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
...
+	})
+	if e != nil {
+		return nil, e
+	}
+
+	return &iav1beta1.CreateTemplateResponse{}, nil
 }
 
 // UpdateTemplate updates existing template, previously created via API.
 func (svc *TemplatesService) UpdateTemplate(ctx context.Context, req *iav1beta1.UpdateTemplateRequest) (*iav1beta1.UpdateTemplateResponse, error) {
```
artemgavrilov, line 377: Should we after Create/Update refresh in-memory state, or it's up to the user (`reload` parameter in `ListTemplates` request)

models/template_model.go
artemgavrilov: Fixed

artemgavrilov:
I'm still working on API tests, but that PR can be reviewed to save time.

services/management/ia/templates_service.go
```diff
@@ -151,13 +250,33 @@ func (svc *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.
...
+
+func convertParamUnit(u string) iav1beta1.ParamUnit {
+	// TODO: check possible variants.
+	switch u {
+	case "%", "percentage":
+		return iav1beta1.ParamUnit_PERCENTAGE
+	default:
+		return iav1beta1.ParamUnit_PARAM_UNIT_INVALID
+	}
+}
```
artemgavrilov, line 271: There is no any validation for unit on `saas.Rule` side, there it should be just non-empty string. Here we have `enum`.

models/template_model.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Template represents Integrated Alerting rule template.
+//reform:ia_templates
+type Template struct {
+	Name        string        `reform:"name,pk"`
+	Version     uint32        `reform:"version"`
+	Summary     string        `reform:"summary"`
+	Tiers       Tiers         `reform:"tiers"`
+	Expr        string        `reform:"expr"`
+	Params      Params        `reform:"params"`
```
palash25, line 37: shouldn't this be a slice? https://github.com/percona/pmm/blob/PMM-2.0/api/managementpb/ia/templates.pb.go#L446

palash25: nevermind, I missed the type alias

BupycHuk (approved):

models/template_helpers.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/percona-platform/saas/pkg/alert"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"gopkg.in/reform.v1"
+)
+
+func checkUniqueTemplateName(q *reform.Querier, name string) error {
+	if name == "" {
+		panic("empty template name")
```
palash25, line 31: why do you choose to panic here instead of returing an error?

artemgavrilov: Same as in https://github.com/percona/pmm-managed/blob/c2cd5d57da2f4dccfe7b2694e1f75e0f53125ebd/models/agent_helpers.go#L33

artemgavrilov: I personally prefer not to panic) But it's common in `checkXXX` methods here.

models/template_helpers.go
```diff
@@ -0,0 +1,222 @@
...
+		c := s.(*Template)
+
+		templates[i] = *c
+	}
+
+	return templates, nil
+}
+
+// FindTemplateByName finds template by name.
+func FindTemplateByName(q *reform.Querier, name string) (*Template, error) {
```
palash25, line 63: this function is same as the `checkUniqueTemplateName` except for the panic, why don't we drop the panic and reuse the code?

artemgavrilov: Idea is that invocation `checkUniqueTemplateName` with empty rule name is the developer's mistake. For example, for agents, we create IDs on the server side, so they always should be present. In templates name provided by user, however, template `should` have a name, so we check it with `rule.Validate()` and notify user about bad request. So if `checkUniqueTemplateName` invoked with an empty name it's not something that can be handled, it's just serve bug.
On the other side, we have `FindTemplateByName` and here we have a different situation. Name for query provided by the user, so we should just notify him about the wrong request.

artemgavrilov: Also `checkUniqueTemplateName` and `FindTemplateByName` have different errors/switch logic.

palash25: got it, thanks

models/template_helpers.go
AlekSi: Panic there is used to indicate programmer's error – that function should not be called with empty ID; it is caller's responsibility to check that ID is not empty if it comes from the API call, etc. (in practice, that should be covered by protobuf validators, but please check that)

models/template_helpers.go
```diff
@@ -0,0 +1,222 @@
...
+	case reform.ErrNoRows:
+		return nil
+	default:
+		return errors.WithStack(err)
+	}
+}
+
+// FindTemplates returns saved notification rule templates.
+func FindTemplates(q *reform.Querier) ([]Template, error) {
+	structs, err := q.SelectAllFrom(TemplateTable, "")
```
palash25, line 47: nit: `structs` -> `rows`

services/management/ia/templates_service.go
```diff
@@ -338,18 +481,84 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
...
 // CreateTemplate creates a new template.
-func (svc *TemplatesService) CreateTemplate(ctx context.Context, req *iav1beta1.CreateTemplateRequest) (*iav1beta1.CreateTemplateResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method CreateTemplate not implemented")
+func (s *TemplatesService) CreateTemplate(ctx context.Context, req *iav1beta1.CreateTemplateRequest) (*iav1beta1.CreateTemplateResponse, error) {
+	pParams := &alert.ParseParams{
+		DisallowUnknownFields: true,
+		DisallowInvalidRules:  true,
+	}
+
+	fmt.Println(req.Yaml)
```
palash25, line 490: i guess this was being used while debugging please remove

services/management/ia/templates_service.go
```diff
@@ -338,18 +481,84 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
 }
 
 // CreateTemplate creates a new template.
-func (svc *TemplatesService) CreateTemplate(ctx context.Context, req *iav1beta1.CreateTemplateRequest) (*iav1beta1.CreateTemplateResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method CreateTemplate not implemented")
+func (s *TemplatesService) CreateTemplate(ctx context.Context, req *iav1beta1.CreateTemplateRequest) (*iav1beta1.CreateTemplateResponse, error) {
+	pParams := &alert.ParseParams{
```
palash25, line 485: `pParams` -> `parseParams`

services/management/ia/templates_service.go
artemgavrilov: Ah... removed

models/template_helpers.go
artemgavrilov: `structs` is consistent with other models.

https://github.com/percona/pmm-managed/blob/c2cd5d57da2f4dccfe7b2694e1f75e0f53125ebd/models/agent_helpers.go#L155

services/management/ia/templates_service.go
```diff
@@ -307,27 +427,50 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
 			Severity:    managementpb.Severity(r.Severity),
 			Labels:      r.Labels,
 			Annotations: r.Annotations,
-			Source:      iav1beta1.TemplateSource_TEMPLATE_SOURCE_INVALID, // TODO
+			Source:      r.Source,
```
palash25, line 430: nit: i guess this loop can be extracted out of the method, its getting too big, maybe move it to a `makeTemplate` function, something like that

models/template_helpers.go
palash25: fine

services/management/ia/templates_service.go
artemgavrilov: Well, I don't know, the scope of that variable is a few lines. Does it require a more verbose name?

palash25 (approved):
just minor nits

services/management/ia/templates_service.go
palash25: nah, its always better to have self-explantory names, but thats a nit

artemgavrilov:
I updated the model to be more strict. Now it uses its own types for fields that can contain a limited amount of values. Also, I added converters for such types.

services/management/ia/templates_service.go
artemgavrilov: Well, I thought about that. In fact, this loop is almost everything that this method does, so extracting it will just produce a big function. Also, this code is not reusable now. IMHO better to leave it as is.

services/management/ia/templates_service.go
```diff
@@ -49,15 +53,23 @@ const (
...
+	Source iav1beta1.TemplateSource
+}
+
 // TemplatesService is responsible for interactions with IA rule templates.
 type TemplatesService struct {
 	db                   *reform.DB
 	l                    *logrus.Entry
 	builtinTemplatesPath string
 	userTemplatesPath    string
+	rulesFileDir         string
```
palash25, line 69: why this field? I don't see it being used in anyway for now. even in the tests you are just creating a temp dir and setting this field but what is the dir used for?

palash25: I just remembered  its supposed to be used in `dumpRule` here `	path := ruleFileDir + alertRule.Alert + ".yml"`

artemgavrilov: Before that field tests dumped rule files in default dir. These files were caught by running managed instance it devcontainer and that caused the error. (It can't parse this format yet).
So I added that field to force tests to dump files in some temporary directory to eliminate side effects.

palash25 (approved):

models/models.go
```diff
@@ -127,8 +127,8 @@ func getCustomLabels(field []byte) (map[string]string, error) {
 	return m, nil
 }
 
-// setCustomLabels encodes custom labels m to the Node/Service/Agent field.
-func setCustomLabels(m map[string]string, field *[]byte) error {
+// setStringMap encodes string map as model's byte array field.
+func setStringMap(m map[string]string, field *[]byte) error {
```
AlekSi, line 131: ⚠️ `prepareLabels` validates that map keys are valid Prometheus labels. That's not a function any string map.

Let's call it getLabels (annotations are basically labels too)

models/database.go
```diff
@@ -371,6 +371,28 @@ var databaseSchema = [][]string{
...
+			name VARCHAR NOT NULL,
+			version INTEGER NOT NULL,
+			summary VARCHAR NOT NULL,
+			tiers JSONB,
+			expr VARCHAR NOT NULL,
+			params JSONB,
+			"for" BIGINT,
+			severity VARCHAR NOT NULL,
+			labels TEXT,
+			annotations TEXT,
```
AlekSi, line 386: Why TEXT and not JSONB? Can it be NOT NULL? Should it?

models/database.go
```diff
@@ -371,6 +371,28 @@ var databaseSchema = [][]string{
...
+			source VARCHAR NOT NULL,
+			yaml TEXT,
```
AlekSi, line 388: Why not NOT NULL?

models/database.go
artemgavrilov: Other models use `TEXT` for labels, so I used the same approach.
https://github.com/percona/pmm-managed/blob/c2cd5d57da2f4dccfe7b2694e1f75e0f53125ebd/models/database.go#L100

`alert.Rule` validation allows empty labels and annotations, so models follow common validness rules.

AlekSi: If we are going to store JSON anyway, we probably may use JSONB and `{}` for empty set of labels. NULLs are less pleasant to work with

But no strong opinion from me there

models/database.go
artemgavrilov: Was not sure that yaml source always will be present. Will change it.

AlekSi: Well, it will be for all templates stored in the database. SaaS templates will not be in the database at all, for example.

But even in that case we could store an empty string. Again, just to avoid NULLs

models/database.go
artemgavrilov: At first, it was JSONB, I change it to TEXT after that comment https://github.com/percona/pmm-managed/pull/541#discussion_r532482651

Will change to JSONB

artemgavrilov: Ah, I remembered, we can use `JSOB` + Custom type with `Scan/Value` methods, **OR** we can use `TEXT` + `setStringMap`(will rename it) that returns byte slice.

artemgavrilov: @AlekSi so we should choose something

AlekSi: Ok, let's use `TEXT`

models/database.go
artemgavrilov: Done

models/models.go
artemgavrilov: Done

services/management/ia/templates_service.go
```diff
@@ -307,27 +462,50 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
...
+				}
+
+				if p.Range != nil {
+					min, max, err := p.GetRangeForFloat()
+					if err != nil {
+						return nil, errors.Wrap(err, "failed to get range for float parameter")
+					}
+
+					fp.HasMin = true      // TODO remove or fill with valid value.
+					fp.Min = float32(min) // TODO eliminate conversion.,
```
artemgavrilov, line 496: In API values for Float have type `float32`, but in saas.Rule it's `float64`

services/management/ia/templates_service.go
```diff
@@ -338,18 +516,82 @@ func (svc *TemplatesService) ListTemplates(ctx context.Context, req *iav1beta1.L
 }
 
 // CreateTemplate creates a new template.
-func (svc *TemplatesService) CreateTemplate(ctx context.Context, req *iav1beta1.CreateTemplateRequest) (*iav1beta1.CreateTemplateResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method CreateTemplate not implemented")
+func (s *TemplatesService) CreateTemplate(ctx context.Context, req *iav1beta1.CreateTemplateRequest) (*iav1beta1.CreateTemplateResponse, error) {
```
artemgavrilov, line 519: Should Create and Change template methods refresh in-memory state? Or it's up to the user to force it with `Reload` option in `ListTemplates` request

services/management/ia/templates_service.go
```diff
@@ -151,15 +286,35 @@ func (svc *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.
...
+
+func convertParamUnit(u string) iav1beta1.ParamUnit {
+	// TODO: check possible variants.
+	switch u {
+	case "%", "percentage":
+		return iav1beta1.ParamUnit_PERCENTAGE
+	default:
+		return iav1beta1.ParamUnit_PARAM_UNIT_INVALID
+	}
 }
```
artemgavrilov, line 318: There is no any validation for unit on alert.Template side, it should be just non-empty string. Here we have enum.

AlekSi (approved):

PR #544, opened 2020-11-28
PMM-6874 Add template rule converter

palash25:
[PMM-6874](https://jira.percona.com/browse/PMM-6874)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1265

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/management/ia/templates_service_test.go
```diff
@@ -81,3 +86,38 @@ func TestCollect(t *testing.T) {
 		assert.Contains(t, rules, "user2_rule")
 	})
 }
+
+func TestConvertTemplate(t *testing.T) {
```
palash25, line 90: fails with the error `pq: duplicate key value violates unique constraint "pg_database_datname_index"` if I use a `t.Parallel here`

services/management/ia/templates_service.go
```diff
@@ -30,18 +33,27 @@ import (
...
 	builtinTemplatesPath = "/tmp/ia1/*.yml"
 	userTemplatesPath    = "/tmp/ia2/*.yml"
+
+	ruleFileDir = "/tmp/ia1/"
+
+	// TODO remove once we start using real values
+	dummyParamValue = "80"
 )
 
+var paramRegex = regexp.MustCompile(`\[\[.*?\]\]`)
```
AlekSi, line 55: Please use text/template with custom delimiters: https://github.com/percona-platform/pocs/blob/e7b00850eec37585a4f69bb15b911fcc8f08a4c7/ia/rule.go#L37

AlekSi (changes requested):

services/management/ia/templates_service.go
palash25: i didn't know the conversion code was already there, thanks

services/management/ia/templates_service.go
```diff
@@ -151,6 +162,126 @@ func (svc *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.
...
+type ruleFile struct {
+	Group []ruleGroup `yaml:"groups"`
+}
+
+type ruleGroup struct {
+	Name  string `yaml:"name"`
+	Rules []rule `yaml:"rules"`
+}
+
+type rule struct {
```
artemgavrilov, line 174: Why not use `alert.Rule`?

services/management/ia/templates_service.go
```diff
@@ -151,6 +162,126 @@ func (svc *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.
...
+
+		err = dumpRule(rf)
+		if err != nil {
+			return errors.Wrap(err, "failed to dump alert rules")
+		}
+	}
+	return nil
+}
+
+// fills templates found in label and annotaions with values.
```
artemgavrilov, line 242:
```suggestion
// fills templates found in labels and annotations with values.
```

services/management/ia/templates_service.go
```diff
@@ -151,6 +162,126 @@ func (svc *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.
 	return rules, nil
 }
 
+type ruleFile struct {
+	Group []ruleGroup `yaml:"groups"`
+}
+
+type ruleGroup struct {
```
artemgavrilov, line 169: I think that type can be defined inside `dumpRule` func.

services/management/ia/templates_service.go
```diff
@@ -151,6 +162,126 @@ func (svc *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.
...
+
+// dump the transformed IA rules to a file.
+func dumpRule(rule ruleFile) error {
+	b, err := yaml.Marshal(rule)
+	if err != nil {
+		return errors.Errorf("failed to marshal rule %s", err)
+	}
+	b = append([]byte("---\n"), b...)
+
+	filepath := ruleFileDir + rule.Group[0].Rules[0].Alert + ".yml"
```
artemgavrilov, line 268: It's easy to get `panic` here. Better to check that the passed rule is initialized properly.

artemgavrilov, line 268: `filepath` variable collides with the package name, better to name it just `path`

services/management/ia/templates_service.go
palash25: but that type represents a template rule and what we are using is an alert rule, they have different fields so can't be reused here.

services/management/ia/templates_service.go
palash25: any reason why you want that? we can do it, but that way `ruleFile` and `ruleGroup` would have to be moved inside `dumpRule` and I wanted all three types for the alert rule to stay in one place.

palash25:
The FB is failing, will fix it in a few mins

services/management/ia/templates_service.go
artemgavrilov: Probably we need to rename `alert.Rule` into `alert.Template`  😑

services/management/ia/templates_service.go
artemgavrilov: Well, the scope of these structs just one function in fact. `ruleFile` and `ruleGroup` are used only for serialization. So it's nice to keep the scope as small as possible.

palash25: will change it

palash25: @artemgavrilov there is a small problem with that, since I use `ruleFile` in my tests I need it outside the `dumpRule` function

artemgavrilov: Well, ok then. Leave it as is

artemgavrilov (approved):

AlekSi (approved):

PR #552, opened 2020-12-01
PMM-7024 Add built-in IA templates

AlekSi:
[PMM-7024](https://jira.percona.com/browse/PMM-7024)

They are not being used yet, but we need to start somewhere to unblock future work.

### DIFF
--- a/data/iatemplates/mysql_restarted.yml
+++ b/data/iatemplates/mysql_restarted.yml
@@ -1,0 +1,23 @@
+---
+rules:
+  - name: mysql_restarted
+    version: 1
+    summary: MySQL Restarted
+    expr: |-
+      mysql_global_status_uptime
+      < [[ .threshold ]]
+    params:
+      - name: threshold
+        summary: Seconds from the last restart
+        unit: "sec"
+        type: int
+        value: 360
+    for: 10s
+    severity: warning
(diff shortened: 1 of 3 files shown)


adivinho:
Now I have four separate tickets related to alert templates for next sprint.
Could this PR be merged so I would create a separate PR for each ticket then?

adivinho (approved):

data/iatemplates/mysql_restarted.yml
```diff
@@ -0,0 +1,23 @@
...
+    version: 1
+    summary: MySQL Restarted
+    expr: |-
+      mysql_global_status_uptime
+      < [[ .threshold ]]
+    params:
+      - name: threshold
+        summary: Seconds from the last restart
+        unit: "sec"
+        type: int
```
palash25, line 13: i guess this was missed in the review but we only support float params https://github.com/percona-platform/platform/blob/main/pkg/alert/type.go#L16

as a result I can't parse this file in managed cc @AlekSi

palash25: for now I have changed it to float in my PR to make the tests pass https://github.com/percona/pmm-managed/pull/558/files#r535992934

PR #558, opened 2020-12-03
PMM-7024 Support for builtin templates

palash25:
[PMM-7024](https://jira.percona.com/browse/PMM-7024)

- Exrtacted createDataDir to a seperate package
- Created new rules and templates directory
- Generate go code from templates
- Used the generated code to read builtin templates
- Changed the path for user defined templates to `/srv/ia/templates`
- Changed the rules dir to `/etc/ia/rules`

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1283

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


data/iatemplates/mysql_restarted.yml
```diff
@@ -10,7 +10,8 @@ rules:
       - name: threshold
         summary: Seconds from the last restart
         unit: "sec"
-        type: int
+        type: float
+        range: [0, 400]
```
palash25, line 14: hack to make the tests pass

palash25:
travis fails because I had changed one of the templates and regenerated the code

palash25:
travis fails because of a change in a template that changed the generated code, so I squashed all the commits and force pushed so that it won't compare `bindata.go` to its previous diff but it can still somehow do that. Is there a way to skip this?

utils/dir/dir.go
```diff
@@ -0,0 +1,70 @@
...
+		aUID, aGID := int(dataDirSysStat.Uid), int(dataDirSysStat.Gid)
+
+		chownDirStat, err := os.Stat(params.ChownPath)
+		if err != nil {
+			return fmt.Errorf("cannot get stat of %q: %v", params.ChownPath, err)
+		}
+
+		chownDirSysStat := chownDirStat.Sys().(*syscall.Stat_t)
+		bUID, bGID := int(chownDirSysStat.Uid), int(chownDirSysStat.Gid)
+		if aUID != bUID || aGID != bGID {
```
palash25, line 63: there is a problem with this bit, it changes the owners for the dir but not for the parent dir for e.g in case of `/srv/alertmanager/data` the `data` dir has get the same owners set as the prometheus dir that is `pmm:pmm` but its parent `alertmanager` dir has `root:root` set.

But when I do `make env-up` and check the owners of the directories  `alertmanager` and `alertmanager/data` both have `pmm:pmm`, I am not sure how the old code was able to achieve this

AlekSi:
@palash25 The other PR was just merged, please resolve merge conflicts

services/management/ia/templates_service_test.go
```diff
@@ -94,39 +97,62 @@ func TestConvertTemplate(t *testing.T) {
...
+		assert.FileExists(t, builtinRuleFilepath1)
+		assert.FileExists(t, builtinRuleFilepath2)
+		assert.FileExists(t, builtinRuleFilepath3)
+
+		testcases := []struct {
+			path             string
+			alert            string
+			labelsCount      int
+			annotationsCount int
+		}{
```
AlekSi, line 116: `}{{`

services/management/ia/templates_service_test.go
```diff
@@ -94,39 +97,62 @@ func TestConvertTemplate(t *testing.T) {
...
+			{
+				path:             builtinRuleFilepath1,
+				alert:            "mysql_down",
+				labelsCount:      3,
+				annotationsCount: 2,
+			},
```
AlekSi, line 122: `}, {`

Makefile.devcontainer
```diff
@@ -10,6 +10,7 @@ gen:                            ## Generate files.
 	rm -f models/*_reform.go
 	find . -name mock_*.go -delete
 	go generate ./...
+	go-bindata -o services/management/ia/bindata.go -pkg ia data/iatemplates/
```
AlekSi, line 13: Can we add `//go:generate` directive to the file itself?

services/management/ia/templates_service.go
```diff
@@ -40,13 +41,19 @@ import (
...
 	"gopkg.in/reform.v1"
 	"gopkg.in/yaml.v3"
+
+	"github.com/percona/pmm-managed/utils/dir"
 )
 
 const (
-	builtinTemplatesPath = "/tmp/ia1/*.yml"
-	userTemplatesPath    = "/tmp/ia2/*.yml"
+	builtinTemplatesPath = "../../../data/iatemplates/*.yml"
```
AlekSi, line 49: That constant should not exist anymore – we should load templates from the generated Go file

services/management/ia/templates_service.go
```diff
@@ -40,13 +41,19 @@ import (
...
+	userTemplatesPath    = "/srv/ia/templates/*.yml"
+
+	templatesDir  = "/srv/ia/templates"
```
AlekSi, line 52: How it is different from the one above?

services/management/ia/templates_service.go
```diff
@@ -71,6 +78,34 @@ func NewTemplatesService(db *reform.DB) *TemplatesService {
 	}
 }
 
+// Run creates tempaltes and rules dir
+func (svc *TemplatesService) Run() {
+	svc.l.Info("Starting...")
+	defer svc.l.Info("Done.")
+	params := []dir.Params{
```
AlekSi, line 85: `[]dir.Params{{`

services/management/ia/templates_service.go
```diff
@@ -71,6 +78,34 @@ func NewTemplatesService(db *reform.DB) *TemplatesService {
...
+		// created both the dirs with the same owners as prometheus dir
+		{
+			Path:      templatesDir,
+			Perm:      dirPerm,
+			Chown:     true,
+			ChownPath: prometheusDir,
+		},
```
AlekSi, line 92: `}, {`

services/management/ia/templates_service.go
```diff
@@ -87,8 +122,9 @@ func (svc *TemplatesService) getCollected(ctx context.Context) map[string]saas.R
 	return res
 }
 
-// collect collects IA rule templates from various sources like
-// built-in templates shipped with PMM and defined by the users.
+// collect collects IA rule templates from various sources like:
+// - builtin templates: read from the generated code in bindata.go
+// - user-defined templates: read from yaml files created by the user in `/srv/ia/templates`
```
AlekSi, line 127: That will not be rendered nicely in godoc

services/management/ia/templates_service_test.go
```diff
@@ -25,21 +25,22 @@ import (
 	"github.com/stretchr/testify/require"
 	"gopkg.in/reform.v1"
 	"gopkg.in/reform.v1/dialects/postgresql"
-	"gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v2"
```
AlekSi, line 28: No, it should be v3

services/management/ia/templates_service.go
```diff
@@ -71,6 +78,34 @@ func NewTemplatesService(db *reform.DB) *TemplatesService {
 	}
 }
 
+// Run creates tempaltes and rules dir
+func (svc *TemplatesService) Run() {
```
AlekSi, line 82: Run methods should accept context, run in a loop and exit when context is canceled.

It looks looks like this code should be inside `NewTemplatesService`

services/management/ia/templates_service_test.go
```diff
@@ -94,39 +97,62 @@ func TestConvertTemplate(t *testing.T) {
...
+			{
+				path:             userRuleFilepath,
+				alert:            "user_rule",
+				labelsCount:      4,
+				annotationsCount: 2,
+			},
+		}
+
+		for _, tc := range testcases {
+			buf, err := ioutil.ReadFile(tc.path)
```
AlekSi, line 144: Please use subtest there

utils/dir/dir.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+	"syscall"
+)
+
+// Params represent the input for CreateDataDir
+type Params struct {
+	Path      string
+	Perm      os.FileMode
+	Chown     bool
+	ChownPath string
```
AlekSi, line 30: Why we need both Path and ChownPath?

We also should accept user name and group name

AlekSi (changes requested):

utils/dir/dir.go
palash25: > Why we need both Path and ChownPath?

ChownPath is the path of the directory whose user name and group name we want to apply to our newly created directory when we chown

I was using that because it keeps code like this contained in the `CreateDataDir` func

```go
        chownDirStat, err := os.Stat(prometheusDir)
	if err != nil {
		svc.l.Errorf("cannot get stat of %q: %v", prometheusDir, err)
	}

	chownDirSysStat := chownDirStat.Sys().(*syscall.Stat_t)
	uID, gID := int(chownDirSysStat.Uid), int(chownDirSysStat.Gid)
```

> We also should accept user name and group name

I think you meant user and group ID, changed the code to accept those instead of `ChownPath`

services/management/ia/templates_service.go
palash25: removed one of them

utils/dir/dir.go
AlekSi: No, we need names like "pmm", not numerical IDs. https://golang.org/pkg/os/user/ – there are looks function to get IDs by name

Makefile.devcontainer
palash25: moved to `templates_service.go`

palash25:
sorry for the force-push i used interactive rebase while resolving conflicts as it seemed easier that way and completely forgot that the commits will be amended

utils/dir/dir.go
palash25: done, thanks 😃

AlekSi (changes requested):
bindata.go should be in the `data` package. Next to it could a dummy.go with `//go:generate` comment. If possible, asset names should be `iatemplates/mysql_down.yml`, etc.

And CI fails

services/alertmanager/alertmanager.go
```diff
@@ -64,7 +64,17 @@ func (svc *Service) Run(ctx context.Context) {
 	svc.l.Info("Starting...")
 	defer svc.l.Info("Done.")
 
-	svc.createDataDir()
+	err := dir.CreateDataDir(dir.Params{
+		Path:  alertmanagerDataDir,
+		Perm:  dirPerm,
+		Chown: true,
```
AlekSi, line 70: It is always true. Let's remove that parameter

utils/dir/dir.go
```diff
@@ -0,0 +1,84 @@
...
+type Params struct {
+	Path  string
+	Perm  os.FileMode
+	Chown bool
+	User  string
+	Group string
+}
+
+// CreateDataDir creates/updates directories with the given permissions in the persistent volume.
+func CreateDataDir(params Params) error {
```
AlekSi, line 37: Let's change that function to return the first encountered error (mostly for tests), but continue as far as possible with its flow. For example, if chmod can't be done, store that error (if it is the first one), but still continue with chown.

And where are the unit tests?

AlekSi (changes requested):

utils/dir/dir.go
palash25: changed it to store the error and return the first encountered one everywhere except for three places where I am still returning the error right away
- when we create dir as there is no point in continuing to chmod and chown if the dir wasn't created
- when we do user lookup because if we continue despite error it can panic
- group lookup same reason as above

palash25: i tried to make chmod fail with invalid permissions like `77777` to check for the first encountered error but it didn't fail somehow the std lib is converting invalid permissions to valid ones. not sure how else to test for it

utils/dir/dir.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot chown datadir %v", err)
+	}
+	bGID, err := strconv.Atoi(group.Gid)
+	if err != nil {
+		if storedErr == nil {
+			storedErr = fmt.Errorf("cannot chown datadir %v", err)
+		} else {
+			return storedErr
+		}
```
artemgavrilov, line 86: Same as above

utils/dir/dir.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if aUID != bUID || aGID != bGID {
+		if err := os.Chown(params.Path, bUID, bGID); err != nil {
+			if storedErr == nil {
+				storedErr = fmt.Errorf("cannot chown datadir %v", err)
+			} else {
+				return storedErr
+			}
```
artemgavrilov, line 95: Same as above

utils/dir/dir_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			actual := CreateDataDir(tc.params)
+			if tc.err != "" {
+				assert.Equal(t, tc.err, actual.Error())
+			} else {
+				stat, err := os.Stat(tc.params.Path)
+				require.NoError(t, err)
+				assert.True(t, stat.IsDir())
+				assert.Equal(t, tc.params.Perm, stat.Mode().Perm())
+			}
```
artemgavrilov, line 72:
```suggestion
			if tc.err != "" {
				assert.Equal(t, tc.err, actual.Error())
				return
			}
			
			stat, err := os.Stat(tc.params.Path)
			require.NoError(t, err)
			assert.True(t, stat.IsDir())
			assert.Equal(t, tc.params.Perm, stat.Mode().Perm())
			
```

services/management/ia/templates_service_test.go
```diff
@@ -50,17 +50,15 @@ func TestCollect(t *testing.T) {
 	sqlDB := testdb.Open(t, models.SkipFixtures, nil)
 	db := reform.NewDB(sqlDB, postgresql.Dialect, reform.NewPrintfLogger(t.Logf))
 
-	t.Run("bad and missing template paths", func(t *testing.T) {
+	t.Run("bad template paths", func(t *testing.T) {
 		t.Parallel()
 
 		testDir, err := ioutil.TempDir("", "")
 		require.NoError(t, err)
 		defer os.RemoveAll(testDir) //nolint:errcheck
```
artemgavrilov, line 58: It's not used anywhere anymore. But the whole idea was to not create test data in default directories as it affects run `managed` in devcontainer.  When you run tests they create some template/rule files, and then `manage` loads these files.

artemgavrilov (changes requested):

utils/dir/dir_test.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+		err: "cannot chown datadir group: unknown group $",
+	},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := CreateDataDir(tc.params)
+			if tc.err != "" {
+				assert.Equal(t, tc.err, actual.Error())
```
artemgavrilov, line 66: `assert.EqualError(t, actual, tc.err)`

utils/dir/dir.go
```diff
@@ -0,0 +1,99 @@
...
+	Perm  os.FileMode
+	User  string
+	Group string
+}
+
+// CreateDataDir creates/updates directories with the given permissions in the persistent volume.
+func CreateDataDir(params Params) error {
+	// try to create data directory
+	if err := os.MkdirAll(params.Path, params.Perm); err != nil {
+		return fmt.Errorf("cannot create datadir %v", err)
```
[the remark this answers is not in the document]
palash25: i dont know, do we want the stacktrace here? I guess that is the only difference between `errors.Wrap` and `fmt.Errorf`

services/management/ia/templates_service_test.go
palash25: nice catch, I think it would be good to re-introduce it and put our converted rules in the tempDir, I will add it again in the next push thanks

palash25: added the tempdir to `TestConvert` and removed it from `TestCollect` because we are not generating any rules there

utils/dir/dir.go
palash25: bumping this up, cc @artemgavrilov @AlekSi

AlekSi: Sounds like we can just invoke that function with "/srv/alertmanager" before invoking it with "/srv/alertmanager/data"

artemgavrilov (approved):

utils/dir/dir.go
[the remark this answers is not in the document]
artemgavrilov: Ok, probably stacktrace redundant here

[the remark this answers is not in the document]
AlekSi: It is not – there are several equal error messages

utils/dir/dir.go
AlekSi: @palash25 That wasn't done ^

utils/dir/dir.go
AlekSi: >  For example, if chmod can't be done, store that error (if it is the first one), but still continue with chown.

But we don't continue in the current code – we return stored error

services/alertmanager/alertmanager.go
```diff
@@ -64,7 +64,11 @@ func (svc *Service) Run(ctx context.Context) {
 	svc.l.Info("Starting...")
 	defer svc.l.Info("Done.")
 
-	svc.createDataDir()
+	err := dir.CreateDataDir(alertmanagerDataDir, "pmm", "pmm", dirPerm)
```
AlekSi, line 67: Before that, we should call dir.CreateDataDir("/srv/alertmanager")

Please also do the same for /srv/victoriametrics and /srv/victoriametrics/data

services/management/ia/templates_service.go
```diff
@@ -100,40 +112,45 @@ func (s *TemplatesService) getCollected(ctx context.Context) map[string]Template
 	return res
 }
 
-// collect collects IA rule templates from various sources like
-// built-in templates shipped with PMM and defined by the users.
+// collect collects IA rule templates from various sources like:
+// builtin templates: read from the generated code in bindata.go.
+// user-defined templates: read from yaml files created by the user in `/srv/ia/templates` or
```
AlekSi, line 117: That's two separate sources. In API, we call them "user file" and "user API"

services/management/ia/templates_service.go
```diff
@@ -43,14 +43,16 @@ import (
...
-	builtinTemplatesPath = "/tmp/ia1/*.yml"
-	userTemplatesPath    = "/tmp/ia2/*.yml"
+	templatesDir = "/srv/ia/templates"
+	rulesDir     = "/etc/ia/rules"
 
-	ruleFileDir = "/tmp/ia1/"
+	dirPerm = os.FileMode(0o775)
 )
 
 // Template represents alerting rule template with added source field.
```
AlekSi, line 58: We already have 3 other Template type definitions: models.Template, saas/pkg/alert.Template, and iav1beta1.Template. Please remove that one.

services/management/ia/templates_service.go
```diff
@@ -151,6 +168,30 @@ func (s *TemplatesService) collect(ctx context.Context) {
...
+func (s *TemplatesService) loadTemplatesFromAssets(ctx context.Context) ([]alert.Template, error) {
+	paths := data.AssetNames()
+	res := make([]alert.Template, 0, len(paths))
+	for _, path := range paths {
+		data, err := data.Asset(path)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to load rule template file: %s", path)
+		}
+
+		// be strict about local files
```
AlekSi, line 180: They are not "local files" anymore

services/management/ia/templates_service.go
```diff
@@ -273,7 +314,6 @@ func convertSeverity(severity models.Severity) common.Severity {
 	}
 }
 
-// loadFile parses IA rule template file.
 func (s *TemplatesService) loadFile(ctx context.Context, file string) ([]saas.Template, error) {
```
AlekSi, line 317: Please _do_ keep comments for un-exported function and methods

utils/dir/dir.go
palash25: if the chmod error is the first encountered error and there are no errors between the chmod and chown call then we can chown even if chmod wasn't successful, I thought thats what you meant by returning the first encountered error. so I added a check for it everywhere.

utils/dir/dir.go
palash25: will push it in a few mins

palash25: @AlekSi c1d673f

utils/dir/dir.go
```diff
@@ -0,0 +1,87 @@
...
+	// check and fix directory permissions
+	dataDirStat, err := os.Stat(path)
+	if err != nil {
+		storedErr = fmt.Errorf("cannot get stat of %q: %v", path, err)
+	}
+
+	if dataDirStat.Mode()&os.ModePerm != perm {
+		if err := os.Chmod(path, perm); err != nil {
+			if storedErr != nil {
+				return storedErr
```
AlekSi, line 48: If we can't chmod, we should continue with chown logic below, not return previous error:

```go
	if err := os.Chmod(path, perm); err != nil {
		err = errors.Wrapf(err, "cannot chmod path %q", path)
		if storedErr == nil {
			storedErr = err
		}
	}
```

There is even no need to do `if dataDirStat.Mode()&os.ModePerm != perm` check

utils/dir/dir.go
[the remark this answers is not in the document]
palash25: replaced the similar chown errors with `errors.Wrap`

utils/dir/dir.go
palash25: checked it locally the parent dirs have the same user:group now

AlekSi (changes requested):

services/management/ia/templates_service.go
palash25: saas/pkg/alert.Template doesn't have yaml and source fields I guess thats why @artemgavrilov created it here, which one should I reuse here? models.Template or iav1beta1.Template?

AlekSi: > saas/pkg/alert.Template doesn't have yaml and source fields

The whole alert.Template describes that YAML

services/alertmanager/alertmanager.go
palash25: > Please also do the same for /srv/victoriametrics and /srv/victoriametrics/data

added it to victoriametrics package, wasn't sure if it should stay in that one or the vmalert package, let me know if otherwise

services/management/ia/templates_service.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/percona-platform/saas/pkg/alert"
+	saas "github.com/percona-platform/saas/pkg/alert"
```
AlekSi: This is the same package as the one above :) 6798fff6c4e58a947ff291b9bd5f438e520eab02

services/management/ia/templates_service.go
palash25: my point being `alert.Template` doesn't have some of the fields (like source and yaml) that we need while returning all the collected templates in methods like `ListTemplates` so we can't reuse this type in place of the Template type defined in this file

That leaves us with 
- `iav1beta1.Template` which if we try to use will force us to convert `alert.Template` to `iav1beta1.Template` in `loadTemplatesFromAssets` and `loadTemplatesFromFiles` methods and various other places
- we have the same problems if we try to replace it with `models.Template`, it requires too many changes

so it doesn't seem like a quick change, I suggest that we skip it for now and remove this type some other time in some other ticket. @AlekSi let me know what you decide going offline now.

AlekSi (approved):

PR #569, opened 2020-12-09
SAAS-360 Remove PERCONA_TEST_XXX_HOST remnants

yurkovychv:
[SAAS-360](https://jira.percona.com/browse/SAAS-360)

- [FB ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1316

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


utils/envvars/parser.go
```diff
@@ -163,22 +161,22 @@ func parseStringDuration(value string) (time.Duration, error) {
...
 	if v == "" {
-		name, v = fallbackEnv, os.Getenv(fallbackEnv)
-	}
-	if v == "" {
-		logrus.Debugf("Using default SaaS host %q.", defaultSaaSHost)
+		logrus.Infof("Using default SaaS host %q.", defaultSaaSHost)
 		return defaultSaaSHost, nil
 	}
+	if strings.HasPrefix(v, ":") {
+		return "", fmt.Errorf("environment variable %q has invalid format %q. Expected host:[port]", name, v)
```
artemgavrilov, line 171: I suppose the correct format is `host[:port]`

utils/envvars/parser.go
```diff
@@ -163,22 +161,22 @@ func parseStringDuration(value string) (time.Duration, error) {
...
 	name, v := envSaaSHost, os.Getenv(envSaaSHost)
 	if v == "" {
-		name, v = fallbackEnv, os.Getenv(fallbackEnv)
-	}
-	if v == "" {
-		logrus.Debugf("Using default SaaS host %q.", defaultSaaSHost)
+		logrus.Infof("Using default SaaS host %q.", defaultSaaSHost)
 		return defaultSaaSHost, nil
 	}
+	if strings.HasPrefix(v, ":") {
```
AlekSi, line 170: Why we need that validation?

utils/envvars/parser.go
```diff
@@ -163,22 +161,22 @@ func parseStringDuration(value string) (time.Duration, error) {
...
+		return "", fmt.Errorf("environment variable %q has invalid format %q. Expected host:[port]", name, v)
+	}
 
 	host, port, err := net.SplitHostPort(v)
-	if err != nil {
+	if err != nil && strings.Count(v, ":") > 1 {
```
AlekSi, line 175: And that? I think net.SplitHostPort should do everything we need

artemgavrilov (approved):

utils/envvars/parser.go
yurkovychv: this is for eliminating values like `:443`. Because we are assigning value to `host` if `host` is empty

utils/envvars/parser.go
```diff
@@ -163,22 +161,22 @@ func parseStringDuration(value string) (time.Duration, error) {
...
 		return "", err
 	}
 	if host == "" {
-		return "", fmt.Errorf("environment variable %q has invalid format %q. Expected host:[port]", name, v)
+		host = v
```
AlekSi, line 179: Well, the old code would check ":443" and return error there. Not sure why this was changed

utils/envvars/parser.go
yurkovychv: it actually does not. If we pass only hostname without `:[port]`, net.SplitHostPort will fail with empty `host`, `port` and an error. So the code below will not work. Current code will add `443` as port

utils/envvars/parser.go
yurkovychv: The old code did not work if we pass only host without `[:port]`

AlekSi (changes requested):
Let's extract a part of `GetSAASHost` as `func parseSAASHost(v string) (string, error)` and add a unit test for it. `GetSAASHost` then will just call `os.Getenv` and `parseSAASHost`.

utils/envvars/parser.go
yurkovychv: done thanks

utils/envvars/parser_test.go
```diff
@@ -132,6 +128,24 @@ func TestEnvVarValidator(t *testing.T) {
...
 
+	t.Run("Parse SAAS host", func(t *testing.T) {
+		userCase := []struct {
+			value   string
+			err     error
+			respVal string
+		}{
+			{value: "host", err: nil, respVal: "host:443"},
+			{value: ":443", err: fmt.Errorf("environment variable %q has invalid format %q. Expected host[:port]", envSaaSHost, ":443"), respVal: ""},
+			{value: "host:443", err: nil, respVal: "host:443"},
```
BupycHuk, line 139: can we use different port in test to check that we don't just set default value

utils/envvars/parser.go
BupycHuk: Then I think there should be check for `>= 1`, not `> 1`

utils/envvars/parser_test.go
BupycHuk: Can we also add a few other invalid cases like `host:letters`

BupycHuk (changes requested):

utils/envvars/parser_test.go
yurkovychv: sure will update tests with different ports, there is no validation for `host:letters` should I add it @BupycHuk ?

BupycHuk: i thought that `net.SplitHostPort` should be able to validate that, but now I see that it can't. so we can skip it if it's out of scope.

utils/envvars/parser.go
yurkovychv: done

AlekSi:
There are a lot of merge conflicts now

Gopkg.toml
```diff
@@ -14,15 +14,15 @@ required = [
 
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "release/2.13"
```
AlekSi, line 17: PMM-2.0

Gopkg.toml
```diff
@@ -14,15 +14,15 @@ required = [
...
 
 [[constraint]]
   name = "github.com/percona-platform/saas"
   branch = "main"
 
 [[constraint]]
   name = "github.com/percona-platform/dbaas-api"
-  branch = "main"
+  branch = "pmm-2.13"
```
AlekSi, line 25: main

AlekSi (changes requested):
Update branches in Gopkg.toml, then run `dev ensure`

Gopkg.toml
yurkovychv: thanks, trying it

services/management/ia/templates_service.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/percona-platform/saas/pkg/alert"
+	saas "github.com/percona-platform/saas/pkg/alert"
```
BupycHuk, line 34: duplicate?

services/management/ia/templates_service.go
```diff
@@ -42,70 +43,44 @@ import (
...
-	"github.com/percona/pmm-managed/utils/dir"
 )
 
 const (
-	templatesParentDir = "/srv/ia"
-	templatesDir       = "/srv/ia/templates"
-	rulesParentDir     = "/etc/ia"
-	rulesDir           = "/etc/ia/rules"
-	dirPerm            = os.FileMode(0o775)
+	builtinTemplatesPath = "/tmp/ia1/*.yml"
```
BupycHuk, line 50: is it safe to use `/tmp/` directory?

services/management/ia/templates_service.go
yurkovychv: I did not do these changes, my changes are related only to `parser`. I'll recreate a PR to not confuse anybody

AlekSi: That's a result of incorrect merging and conflicts resolving

services/management/ia/templates_service.go
AlekSi: That's a result of incorrect merging and conflicts resolving too

services/management/ia/templates_service.go
yurkovychv: Thanks. Creating new PR right now

PR #589, opened 2020-12-21
PMM-6964 Generate alertmanager config

palash25:
[PMM-6964](https://jira.percona.com/browse/PMM-6964)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1344

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


Gopkg.toml
```diff
@@ -28,3 +28,7 @@ required = [
 [[constraint]]
   name = "github.com/percona/exporter_shared"
   version = "0.7.2"
+
+[[constraint]]
+  name = "github.com/mxschmitt/golang-combinations"
```
AlekSi, line 33: Hard no. As we discussed, we should keep only actually used combination. We can't afford to generate 2^N combinations only to use a few of them.

Instead, we should iterate over rules and keep unique combinations of channels

AlekSi (changes requested):

Gopkg.toml
palash25: > we should keep only actually used combination

by actually used combination do you mean the combination of channelIDs used in the rule as one combination and then each individual channel?

for e.g. if there is a rule with 3 channels `/channel_id/1`, `/channel_id/2 `, `/channel_id/3` the receivers should look like this?

```
receivers:
  - name: empty
 
  - name: /channel_id/1
 
  - name: /channel_id/2

  - name: /channel_id/3
 
  - name: /channel_id/1 + /channel_id/2 + /channel_id/3
```

AlekSi: Only "empty" (which is special) and "/channel_id/1 + /channel_id/2 + /channel_id/3". Three channels in the middle are not used by any existing rule

services/alertmanager/alertmanager_test.go
```diff
@@ -36,3 +44,95 @@ func TestAlertmanager(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+
+	svc := New(db)
+	err = svc.populateConfig(&cfg)
+	require.NoError(t, err)
+
+	assert.Len(t, cfg.Receivers, 1)
+	assert.Equal(t, "empty", cfg.Receivers[0].Name)
+	assert.Equal(t, channel1.ID+" + "+channel2.ID, cfg.Receivers[1].Name)
```
AlekSi, line 132: Two lines above you check that `cfg.Receivers` contains one element with index 0; here the second element is accessed

palash25: thats a typo, that was 2 before. it is supposed to have 2 elements but for some reason the db transactions are failing while creating the rules in the testdb, which is why this assertion also fails.

Gopkg.toml
AlekSi: Please remove that constraint as we are no longer use that package

services/alertmanager/alertmanager.go
```diff
@@ -140,12 +144,17 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
 			return
 		}
 
-		// TODO add custom information to this config.
+		err = svc.populateConfig(&cfg)
+		if err != nil {
+			svc.l.Error(err)
+		}
```
AlekSi, line 150: Missing `return`?

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +166,158 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+
+	cfg.Global = &alertmanager.GlobalConfig{}
+	if settings.IntegratedAlerting.EmailAlertingSettings != nil {
+		cfg.Global.SMTPFrom = settings.IntegratedAlerting.EmailAlertingSettings.From
+		cfg.Global.SMTPHello = settings.IntegratedAlerting.EmailAlertingSettings.Hello
+		cfg.Global.SMTPAuthIdentity = settings.IntegratedAlerting.EmailAlertingSettings.Identity
+		cfg.Global.SMTPAuthUsername = settings.IntegratedAlerting.EmailAlertingSettings.Username
+		cfg.Global.SMTPAuthPassword = settings.IntegratedAlerting.EmailAlertingSettings.Password
+		cfg.Global.SMTPAuthSecret = settings.IntegratedAlerting.EmailAlertingSettings.Secret
+		smarthost := strings.Split(settings.IntegratedAlerting.EmailAlertingSettings.Smarthost, ":")
```
AlekSi, line 188: net.SplitHostPort

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +166,158 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	e = svc.db.InTransaction(func(tx *reform.TX) error {
+		var err error
+		rules, err = models.FindRules(tx.Querier)
+		return err
+	})
+	if e != nil {
+		return errors.Errorf("Failed to retrieve alert rules from database: %s", e)
+	}
+
+	e = svc.db.InTransaction(func(tx *reform.TX) error {
```
AlekSi, line 208: Please get settings, rules, and channels in one transaction

services/alertmanager/alertmanager_test.go
```diff
@@ -36,3 +46,123 @@ func TestAlertmanager(t *testing.T) {
...
+	db := reform.NewDB(sqlDB, postgresql.Dialect, reform.NewPrintfLogger(t.Logf))
+	tx, err := db.Begin()
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, tx.Rollback())
+	}()
+
+	q := tx.Querier
+
+	channel1, err := models.CreateChannel(q, &models.CreateChannelParams{
```
AlekSi, line 67: This channel is created in one transaction. See the other comment below.

The simplest solution is to use db.Querier there and not create test transaction

services/alertmanager/alertmanager_test.go
```diff
@@ -36,3 +46,123 @@ func TestAlertmanager(t *testing.T) {
...
+			Secret:    "secret",
+		},
+		SlackAlertingSettings: &models.SlackAlertingSettings{
+			URL: gofakeit.URL(),
+		},
+	})
+	require.NoError(t, err)
+
+	svc := New(db)
+	err = svc.populateConfig(&cfg)
```
AlekSi, line 145: This method creates its own transaction. Channel created in test transaction is not visible in this method transaction

AlekSi (changes requested):

palash25:
@AlekSi the confluence doc mentions individual channels as receivers which were not generated but were fetched from the database

```
  # Channel with ID /channel_id/1 from pmm-managed database.
  - name: /channel_id/1
    email_configs:
      - to: sue.higgins@example.com
 
  # Channel with ID /channel_id/2 from pmm-managed database.
  - name: /channel_id/2
    pagerduty_configs:
      - routing_key: ms-pagerduty-dev
```

Does this mean all channels that are there in the database or only those individual channels used by the rules. For eg if there are ch1, ch2, ch3 in the db and the rule only uses ch1 and ch2 should all the 3 channels be there in the config or just ch1 and ch2?

AlekSi:
Is this the same question as https://github.com/percona/pmm-managed/pull/589#discussion_r546743879?

`/channel_id/1` alone is not needed unless there is a rule that actually uses that channel alone

AlekSi:
@artemgavrilov PTAL

palash25:
> Is this the same question as #589 (comment)?

no its not the same question that question was about whether to generate individual channel receivers or not, which is clear now, we don't do that, the example in the confluence doc shows two rules both of which use a combination of ch1 and ch2  as their receiver and the receiver field has a receiver `ch1 + ch2` but there  are two other receivers just for ch1 and ch2 separately even though there is no rule that uses just ch1 or ch2 and the comment (`# Channel with ID /channel_id/1 from pmm-managed database.`) on both those receivers says that they were fetched from the DB not generated as one of the combination.

So currently my code only adds the combination of channels used by a rule and preserves the empty receiver but doesn't add any of these receivers 👇🏽 that are mentioned in this doc

```
  # Channel with ID /channel_id/1 from pmm-managed database.
  - name: /channel_id/1
    email_configs:
      - to: sue.higgins@example.com
 
  # Channel with ID /channel_id/2 from pmm-managed database.
  - name: /channel_id/2
    pagerduty_configs:
      - routing_key: ms-pagerduty-dev

```

so this question https://github.com/percona/pmm-managed/pull/589#issuecomment-750298047 is related to those receivers

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if e != nil {
+		return errors.Errorf("Failed to fetch items from database: %s", e)
+	}
+
+	cfg.Global = &alertmanager.GlobalConfig{}
```
artemgavrilov, line 198: It looks like you should allocte new config only if `cfg.Global` is nil, otherwise you will lose user defined values.

AlekSi:
Yes, we should not generate unused configuration. I will update the Confluence page

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+	if settings.IntegratedAlerting.EmailAlertingSettings != nil {
+		cfg.Global.SMTPFrom = settings.IntegratedAlerting.EmailAlertingSettings.From
+		cfg.Global.SMTPHello = settings.IntegratedAlerting.EmailAlertingSettings.Hello
+		cfg.Global.SMTPAuthIdentity = settings.IntegratedAlerting.EmailAlertingSettings.Identity
+		cfg.Global.SMTPAuthUsername = settings.IntegratedAlerting.EmailAlertingSettings.Username
+		cfg.Global.SMTPAuthPassword = settings.IntegratedAlerting.EmailAlertingSettings.Password
+		cfg.Global.SMTPAuthSecret = settings.IntegratedAlerting.EmailAlertingSettings.Secret
```
artemgavrilov, line 205: How to handle case when user already specified these values in base config file?

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+
+// generateReceivers takes the channel map and a unique set of rule combinations and generates a slice of receivers.
+func generateReceivers(chanMap map[string]*models.Channel, recvSet map[string]struct{}) ([]*alertmanager.Receiver, error) {
+	receivers := make([]*alertmanager.Receiver, 0, len(recvSet))
+	for k := range recvSet {
+		recv := &alertmanager.Receiver{
+			Name: k,
+		}
+
+		individualChannels := strings.Split(k, receiverNameSeparator)
```
artemgavrilov, line 256: You can put channels in `recvSet map[string]ChannleIDs` to eliminate that string split

AlekSi:
@palash25 The other PR was merged, please PMM-2.0 into that branch

services/alertmanager/alertmanager.go
AlekSi: For now: log a warning about it and overwrite.

In the the future, we might want to move those settings into each receiver

palash25: i forgot that the base file could be edited, in that case I think we should preserve the changes that the user made in the base file, @AlekSi can you please confirm this?

AlekSi: Nope, see my comment above

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+		for _, ch := range individualChannels {
+			channel := chanMap[ch]
+			switch channel.Type {
+			case models.Email:
+				recv.EmailConfigs = append(recv.EmailConfigs, &alertmanager.EmailConfig{
+					NotifierConfig: alertmanager.NotifierConfig{
+						SendResolved: channel.EmailConfig.SendResolved,
+					},
+					To: channel.EmailConfig.To[0],
```
artemgavrilov, line 265: It should be one `EmailConfig` for each value in `channel.EmailConfig.To`.
Something like this: https://stackoverflow.com/a/53329483

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+						RoutingKey: pdConfig.RoutingKey,
+					})
+					break
+				}
+				recv.PagerdutyConfigs = append(recv.PagerdutyConfigs, &alertmanager.PagerdutyConfig{
+					NotifierConfig: alertmanager.NotifierConfig{
+						SendResolved: pdConfig.SendResolved,
+					},
+					ServiceKey: pdConfig.ServiceKey,
+				})
```
artemgavrilov, line 283: `ServiceKey` also can be empty. To be more precise, only one can be specified, `RoutingKey` or `ServiceKey`.

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+					NotifierConfig: alertmanager.NotifierConfig{
+						SendResolved: channel.WebHookConfig.SendResolved,
+					},
+					URL:       channel.WebHookConfig.URL,
+					MaxAlerts: uint64(channel.WebHookConfig.MaxAlerts),
+					HTTPConfig: promconfig.HTTPClientConfig{
+						BasicAuth: &promconfig.BasicAuth{
+							Username:     channel.WebHookConfig.HTTPConfig.BasicAuth.Username,
+							Password:     channel.WebHookConfig.HTTPConfig.BasicAuth.Password,
+							PasswordFile: channel.WebHookConfig.HTTPConfig.BasicAuth.PasswordFile,
```
artemgavrilov, line 302: `BasicAuth` can be nil

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+						},
+						BearerToken:     channel.WebHookConfig.HTTPConfig.BearerToken,
+						BearerTokenFile: channel.WebHookConfig.HTTPConfig.BearerTokenFile,
+						TLSConfig: promconfig.TLSConfig{
+							CAFile:             channel.WebHookConfig.HTTPConfig.TLSConfig.CaFile,
+							CertFile:           channel.WebHookConfig.HTTPConfig.TLSConfig.CertFile,
+							KeyFile:            channel.WebHookConfig.HTTPConfig.TLSConfig.KeyFile,
+							ServerName:         channel.WebHookConfig.HTTPConfig.TLSConfig.ServerName,
+							InsecureSkipVerify: channel.WebHookConfig.HTTPConfig.TLSConfig.InsecureSkipVerify,
```
artemgavrilov, line 311: `TLSConfig` can be nil

services/alertmanager/alertmanager.go
artemgavrilov: And whole `HTTPConfig` also can be nil

services/alertmanager/alertmanager.go
```diff
@@ -157,6 +169,159 @@ func (svc *Service) updateConfiguration(ctx context.Context) {
...
+						},
+						ProxyURL: channel.WebHookConfig.HTTPConfig.ProxyURL,
+					},
+				})
+			default:
+				return nil, errors.New("Invalid channel type")
```
artemgavrilov, line 317:
```suggestion
				return nil, errors.New("invalid channel type")
```

services/alertmanager/alertmanager_test.go
```diff
@@ -36,3 +47,171 @@ func TestAlertmanager(t *testing.T) {
 
 	require.NoError(t, svc.IsReady(context.Background()))
 }
+
+func TestPopulateConfig(t *testing.T) {
+	t.Run("without receivers and routes", func(t *testing.T) {
+		var cfg alertmanager.Config
+		buf, err := ioutil.ReadFile(alertmanagerBaseConfigPath)
```
artemgavrilov, line 54: Probably it makes sense to create some fake base config, and check that we are not overriding user data.

palash25: didn't we just decide to overwrite user data? https://github.com/percona/pmm-managed/pull/589#discussion_r547989698

artemgavrilov: Yep, but there is possible case when user value is defined, but in settings not. So we should not remove users values, we should only override them if we have another value.

palash25: added one field to the global config to check for overwrite

services/alertmanager/alertmanager.go
palash25: added warning

AlekSi (approved):

PR #599, opened 2020-12-24
PMM-7220 Don't delete template if it is being used by rules

palash25:
[PMM-7220](https://jira.percona.com/browse/PMM-7220)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1350

- [ ] Links to other linked pull requests (optional). https://github.com/Percona-Lab/pmm-api-tests/pull/163
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/template_helpers.go
```diff
@@ -183,7 +183,18 @@ func RemoveTemplate(q *reform.Querier, name string) error {
...
+	rules, err := FindRules(q)
+	if err != nil {
+		return err
+	}
+
+	for _, rule := range rules {
+		if name == rule.TemplateName {
+			return errors.Errorf("failed to delete rule template, as it is being used by rule: %s", rule.ID)
+		}
+	}
```
artemgavrilov, line 195: Better to write helper function `FindRulesByTemplateName` and not to load all rules in memory. Database can find specific rules for you.

palash25:
the latest commit is wrong, working on a fix

models/template_helpers.go
palash25: used reform to find rules by template name it was just two lines so didn't write a separate helper for it

models/template_helpers.go
```diff
@@ -183,6 +183,10 @@ func RemoveTemplate(q *reform.Querier, name string) error {
 		return err
 	}
 
+	if _, err := q.FindOneFrom(RuleTable, "template_name", name); err == nil {
+		return errors.Errorf("failed to delete rule template, as it is being used by a rule")
+	}
```
artemgavrilov, line 188: Error message is true only wher returned error is `reform.ErrNoRows`, otherwise reason can be different.

artemgavrilov (changes requested):

models/template_helpers.go
palash25: but its only checking for when we don't have an error `err == nil` (meaning that there was a rule with the template name), in case there is an error that means there are no rules with that template name and we can carry on deleting the template instead of returning that  error.

artemgavrilov: 🤦🤦🤦 You right

AlekSi: We will also delete the template if there was some temporary error like lost database connection, etc. Artem is right – we should check only for reform.ErrNoRows

artemgavrilov (approved):
There is some questions left. How we will handle deletion of templates added as files?

models/template_helpers.go
palash25: I think the current code can be confusing should I do something like this would this be more readable?

```go

func DeleteTemplate() {
if templateInUse(name) {
   return errors.Errorf("failed to delete rule template, as it is being used by a rule")
}
}


func templateInUse(name string) bool {
  if _, err := q.FindOneFrom(RuleTable, "template_name", name); err != nil {
		return false
   }
   return true
}
```

palash25: ignore the comment, I posted it without refreshing the page

palash25: added error check https://github.com/percona/pmm-managed/pull/599/commits/a91f584ae772c6ee4d6218302c11ddaf6d6aafe5

artemgavrilov (approved):

palash25:
> There is one question left. How we will handle deletion of templates added as files?

those are not in the DB so why do we need to handle their deletion from the API? aren't users supposed to add/delete them as and when required?

artemgavrilov:
> > There is one question left. How we will handle deletion of templates added as files?
> 
> those are not in the DB so why do we need to handle their deletion from the API? aren't users supposed to add/delete them as and when required?

Yes, users can delete templates files when they want, but how we should handle rules that are left? Now as I remember we will just log error about missing template.

AlekSi (approved):
Will wait for CI

PR #600, opened 2020-12-24
PMM-7214 Fix/simplify configuration reload

AlekSi:
[PMM-7214](https://jira.percona.com/browse/PMM-7214)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1349

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/alertmanager/alertmanager.go
```diff
@@ -60,120 +66,262 @@ const (
...
+		case <-ctx.Done():
+			return
 
-	<-ctx.Done()
-}
+		case <-svc.reloadCh:
+			// batch several update requests together by delaying the first one
+			sleepCtx, sleepCancel := context.WithTimeout(ctx, updateBatchDelay)
+			<-sleepCtx.Done()
+			sleepCancel()
```
artemgavrilov, line 121: How that is supposed to work with buffered channel? When I hit `ChangeSettings` twice:
1) first message sent to the  channel
2) first message read from channel and reading goroutine stoped for batch delay
3) second message is sent to the channel
4) after batch delay first message processed
5) second message is read form channel
6) ....

if we will change buffered channel to unbuffered second message will be not sent in channel because write is done with select.

Am I missing something?

AlekSi: If channel will be unbuffered, the following scenario is possible:

1. first message set to the channel
2. first message read from channel and reading goroutine stoped for batch delay
3. after batch delay first message processed
4. while first message is processed, second message is being sent to the channel – and dropped, because Run is not reading from the channel at that moment

artemgavrilov: Then whole that mechanism is broken and should be done a different way (obvoiusly not now). 

We can spawn new goroutine for message processing, or we can accept that extra time spent on first message processin as addittion to the batch delay.

Ok, it is not the proper time for that.

artemgavrilov (approved):

AlekSi:
```
level=info ts=2020-12-24T16:55:33.417Z caller=coordinator.go:119 component=configuration msg="Loading configuration file" file=/etc/alertmanager.yml
level=error ts=2020-12-24T16:55:33.417Z caller=coordinator.go:124 component=configuration msg="Loading configuration file failed" file=/etc/alertmanager.yml err="no route provided in config"
```

main.go
```diff
@@ -451,7 +451,7 @@ func setup(ctx context.Context, deps *setupDeps) bool {
 	deps.vmalert.RequestConfigurationUpdate()
 
 	// Alertmanager is special due to being added to PMM with invalid /etc/alertmanager.yml.
-	deps.alertmanager.GenerateBaseConfigs(ctx)
+	deps.alertmanager.GenerateBaseConfigs()
```
artemgavrilov, line 454: Why not just replace invalid `/etc/alermanager.yml` with valid one at PMM image creation?

AlekSi: Because it should also work for various update methods

services/vmalert/vmalert.go
```diff
@@ -76,33 +76,33 @@ func NewVMAlert(externalRules *ExternalRules, baseURL string) (*Service, error)
...
+	// If you change this and related methods,
+	// please do similar changes in alertmanager and victoriametrics packages.
+
 	svc.l.Info("Starting...")
 	defer svc.l.Info("Done.")
 
+	// reloadCh, configuration update loop, and RequestConfigurationUpdate method ensure that configuration
+	// is reloaded when requested, but several requests are batched together to avoid too often reloads.
+	// That allows the caller to just call RequestConfigurationUpdate when it seems fit.
+	if cap(svc.reloadCh) != 1 {
```
palash25, line 96: this channel is always buffered then why do we have this check?

palash25 (approved):

artemgavrilov (approved):

main.go
artemgavrilov: Ah, got it!

services/vmalert/vmalert.go
AlekSi: To prevent someone from changing buffer size without reading that comment

PR #602, opened 2020-12-24
PMM-7218 Generate rule files

artemgavrilov:
[PMM-7218](https://jira.percona.com/browse/PMM-7218)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1353

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


artemgavrilov:
Now Alerts API is working

services/management/ia/rules_service_test.go
```diff
@@ -0,0 +1,160 @@
...
+				Value: "60",
+			},
+		},
+		ChannelIds: []string{channelID},
+	})
+	require.NoError(t, err)
+	ruleID := resp.RuleId
+	defer func() {
+		e := db.InTransaction(func(t *reform.TX) error {
+			return models.RemoveRule(t.Querier, ruleID)
```
AlekSi, line 123: `testdb.Open` recreates the whole database. There is no need to do manual cleanup.

AlekSi (commented):
Please check two things:
* that we don't generated files for disabled rules – I don't see a check for that;
* that invalid built-in rules (for example, without pmm_ prefix) break tests – test that loads built-in files seems to be removed.

services/management/ia/rules_service_test.go
```diff
@@ -0,0 +1,160 @@
...
+            max_over_time(mysql_global_status_threads_connected[5m]) / ignoring (job)
+            mysql_global_variables_max_connections
+            * 100
+            > 1.22
+          for: 2s
+          labels:
+            baz: faz
+            foo: bar
+            ia: "1"
+            severity: SEVERITY_INFO
```
AlekSi, line 149: Both rule ID and template name should be there

services/management/ia/rules_service_test.go
```diff
@@ -0,0 +1,160 @@
...
+          annotations:
+            description: |-
+                More than 1.22%% of MySQL connections are in use on {{ $labels.instance }}
+                VALUE = {{ $value }}
+                LABELS: {{ $labels }}
+            rule_summary: some testing rule
```
AlekSi, line 155: Just "rule"

services/management/ia/rules_service.go
```diff
@@ -17,38 +17,225 @@
...
+	rulesPath    string // used for testing
+
 }
 
 // NewRulesService creates an API for Integrated Alerting Rules.
-func NewRulesService(db *reform.DB, templates *TemplatesService, vmalert vmAlertService) *RulesService {
+func NewRulesService(db *reform.DB, templates *TemplatesService, vmalert vmAlert, alertManager alertManager) *RulesService {
+	l := logrus.WithField("component", "management/ia/rules")
+
+	err := dir.CreateDataDir(rulesDir, "pmm", "pmm", dirPerm)
```
palash25, line 60: you might want to add another call to change the owners of its parent dir `"/etc/ia/"` last time I was doing it, the owner:group of the parent dir turned out to be different from the child so I added two calls with separate paths in the templates service

services/management/ia/templates_service.go
```diff
@@ -80,21 +70,7 @@ type TemplatesService struct {
 func NewTemplatesService(db *reform.DB) *TemplatesService {
 	l := logrus.WithField("component", "management/ia/templates")
 
-	err := dir.CreateDataDir(templatesParentDir, "pmm", "pmm", dirPerm)
```
palash25, line 83: same as https://github.com/percona/pmm-managed/pull/602/files#r548864528

services/management/ia/rules_service.go
artemgavrilov: Does it caused some problems in you case? And did you find reason why parend dir had differerent owner?

services/management/ia/rules_service_test.go
artemgavrilov: 🤦‍♂️ removed

services/management/ia/rules_service_test.go
artemgavrilov: Added

services/management/ia/rules_service_test.go
artemgavrilov: Renamed

services/management/ia/rules_service.go
palash25: > Does it caused some problems in you case?

not that I can remember, for context you can take a look at this thread https://github.com/percona/pmm-managed/pull/558/#discussion_r536101878

> And did you find reason why parend dir had differerent owner?

no I couldn't

artemgavrilov: Well, lets keep it as is until there is no problems. But I will keep in mind that such case is possible.

palash25 (approved):

AlekSi (commented):
API tests fail

artemgavrilov:
> API tests fail

I recreated FB, and now API tests are green. Previous build was bad for some reason. 
Now there is another porblem.

services/management/ia/rules_service.go
AlekSi: There should not be a problem for `/etc/…` dirs that are internal for container. `/srv/…`, however, can be on an external volume, and there problems could happen

AlekSi (approved):

PR #608, opened 2021-01-05
PMM-7167 Grafana concurrency problem.

JiriCtvrtka:
[PMM-7167](https://jira.percona.com/browse/PMM-7167)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1364

---
- [x] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/supervisord/supervisord.go
```diff
@@ -579,6 +579,9 @@ command =
 		--search.disableCache={{ .VMDBCacheDisable }}
 		--search.maxQueryLen=72KB
 		--search.latencyOffset=5s
+		--search.maxConcurrentRequests=32
+		--search.maxQueueDuration=20s
+		--search.maxQueryDuration=30s
```
askomorokhov, line 584: Looks like 30s is a default value:

https://hub.docker.com/r/victoriametrics/victoria-metrics/

> The maximum duration for search query execution (default 30s)

services/supervisord/supervisord.go
```diff
@@ -579,6 +579,9 @@ command =
 		--search.disableCache={{ .VMDBCacheDisable }}
 		--search.maxQueryLen=72KB
 		--search.latencyOffset=5s
+		--search.maxConcurrentRequests=32
+		--search.maxQueueDuration=20s
```
askomorokhov, line 583: According to the description:

https://hub.docker.com/r/victoriametrics/victoria-metrics/

> The maximum time the request waits for execution when -search.maxConcurrentRequests limit is reached (default 10s)

if we 33 concurrent "slow" requests (may run maxQueryDuration=30s) run in concurrently so the 33rd query needs to wait 30 sec before accept to execution, right?

askomorokhov (commented):
Do we have any justification for these numbers?
How many maximum concurrent queries we need to execute for typical grafana dashboard?

@adivinho, could you please help us with this?

services/supervisord/supervisord.go
JiriCtvrtka: Oh yeah, I forget increase search.maxQueryDuration. Thanks will fix it.

services/supervisord/supervisord.go
JiriCtvrtka: Good point. WIll set maxQueueDuration bigger than maxQueryDuration.

askomorokhov (approved):

testdata/supervisord.d/victoriametrics.ini
```diff
@@ -11,6 +11,9 @@ command =
 		--search.disableCache=true
 		--search.maxQueryLen=72KB
 		--search.latencyOffset=5s
+		--search.maxConcurrentRequests=32
+		--search.maxQueueDuration=90s
```
adivinho, line 15: What's the reason to fetch data after 60 seconds. 60 sec is default Grafana "Query timeout" to VM datasource.
Screenshot_20210111_145256

adivinho (changes requested):

testdata/supervisord.d/victoriametrics.ini
JiriCtvrtka: If query timeout from Grafana to VM is 60s then I can remove line search.maxQueryDuration completly. It was set, because VM default was 30s.   


See: 
https://hub.docker.com/r/victoriametrics/victoria-metrics/
-search.maxQueryDuration duration
        The maximum duration for search query execution (default 30s)

adivinho:
> Do we have any justification for these numbers?
> How many maximum concurrent queries we need to execute for typical grafana dashboard?
> 
> @adivinho, could you please help us with this?

I have tested on out pmmdemo installation by aggressive reloading our home dashboard.
So it was shown about 90 ops for /api/v1/query_range
with default --search.maxConcurrentRequests=8. And there were no errors.

Screenshot_20210111_173339


Errors came up for long time ranges only if few charts are displayed on a screen.

Screenshot_20210111_175953

So amount of query_range requests related to VM internal strategy of splitting requested time range.

If maxConcurrentRequests=8 is enough for one chart then 32 would be good for average four charts on a screen.

testdata/supervisord.d/victoriametrics.ini
adivinho: Let's multiply by two default values maxQueryDuration and maxQueueDuration
--search.maxQueueDuration=20s
--search.maxQueryDuration=60s

services/supervisord/supervisord.go
adivinho: @JiriCtvrtka 
Could you clarify that I got it right?
maxQueryDuration -> lifetime of a general query (e.g. show me a metric for 30 days)
maxQueueDuration -> lifetime of a request as a result of splitting a general query by services, types .... (e.g. 80 requests)

So if we have eight core CPU and spends 1 sec for  each request than a query (80 requests) would be processed by 20 seconds.

The question is next - maxQueueDuration is been started counting since a query is generated (spitted to 80 requests) or when a request tried to be executed and stacked due CPU core occupation?
If the first statement is correct than query maxQueryDuration duration is useless because a whole query would be break after maxQueueDuration time. 

So most probably the second statement is correct. In such case it's not reasonable increasing maxConcurrentRequests to 32.

My suggestion is to just double default values for all involved parameters.
In such case maxQueryDuration will be equal to Grafana Query timeout.

JiriCtvrtka: I checked it and everytime when I changed "refresh interval" then was created around 77 requests (query, query_range) in PMM demo. Not sure if set 16 (double default value) will be enough. Lets wait also for Nurlan opinion.

testdata/supervisord.d/victoriametrics.ini
BupycHuk: Agree with @adivinho

JiriCtvrtka: @adivinho @BupycHuk 

So summary:
--search.maxQueueDuration=20s
--search.maxQueryDuration=60s
--search.maxConcurrentRequests=16

All values doubled then defaults one. OK will change it. Thanks guys.

BupycHuk: let's keep --search.maxConcurrentRequests=32

adivinho (approved):

BupycHuk (approved):

percona-csalguero (approved):

adivinho (approved):

PR #610, opened 2021-01-06
PMM-7226 rule params validation

artemgavrilov:
[PMM-7226](https://jira.percona.com/browse/PMM-7226)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1369

- [ ] https://github.com/Percona-Lab/pmm-api-tests/pull/167
---
- [x] Tests passed.
- [x] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


palash25 (approved):

services/management/ia/rules_service_test.go
```diff
@@ -146,6 +149,148 @@ groups:
...
+				"baz": "faz",
+			},
+			Filters: []*iav1beta1.Filter{{
+				Type:  iav1beta1.FilterType_EQUAL,
+				Key:   "some_key",
+				Value: "'60'",
+			}},
+			ChannelIds: []string{channelID},
+		})
+		tests.AssertGRPCError(t, status.New(codes.InvalidArgument, "Missing parameter threshold."), err)
```
BupycHuk, line 186: looks like we don't check for wrong parameters in this test.

BupycHuk (approved):

services/management/ia/rules_service_test.go
artemgavrilov: We are checking that rule has the same number of parameters as in the template and then we are iterating through parameters defined in the template and compare them with those defined in rule. 

This test about the case when passed amount of parameters invalid, but actual parameters are not.

artemgavrilov: Updated, now proper error returned

artemgavrilov:
I'm added functionality that allows for rule creation to not pass parameters that has a default value in templates.

services/management/ia/rules_service_test.go
```diff
@@ -146,6 +148,148 @@ groups:
...
+				Type:  iav1beta1.FilterType_EQUAL,
+				Key:   "some_key",
+				Value: "'60'",
+			}},
+			ChannelIds: []string{channelID},
+		})
+		tests.AssertGRPCError(t, status.New(codes.InvalidArgument, "Unknown parameters [unknown parameter]."), err)
+	})
+
+	// t.Run("missing parameter", func(t *testing.T) { TODO: PMM-7279 fix once templates will support parameters without default value.
```
BupycHuk, line 188: let's use `t.Skip()` in such cases, so it will be visible in our tests list.

artemgavrilov: Done

services/management/ia/rules_service.go
```diff
@@ -355,6 +352,87 @@ func (s *RulesService) CreateAlertRule(ctx context.Context, req *iav1beta1.Creat
...
+				}
+				unknownParams[rp.Name] = false
+				filled = true
+				res = append(res, rp)
+				break
+			}
+		}
+
+		if !filled {
+			if tp.Value == nil {
```
BupycHuk, line 388: shouldn't this be `tp.DefaultValue`?

services/management/ia/rules_service.go
```diff
@@ -355,6 +352,87 @@ func (s *RulesService) CreateAlertRule(ctx context.Context, req *iav1beta1.Creat
...
+		return nil, status.Errorf(codes.NotFound, "Unknown template %s.", templateName)
+	}
+
+	unknownParams := make(map[string]bool)
+
+	res := make(models.RuleParams, 0, len(t.Params))
+	for _, tp := range t.Params {
+		var filled bool
+		for _, rp := range ruleParams {
+			if _, ok := unknownParams[rp.Name]; !ok {
```
BupycHuk, line 372: for better readability we can extract filling map from the loop whole loop and instead of setting `false` as value just delete the pair from map.

services/management/ia/rules_service.go
```diff
@@ -355,6 +352,87 @@ func (s *RulesService) CreateAlertRule(ctx context.Context, req *iav1beta1.Creat
...
+				filled = true
+				res = append(res, rp)
+				break
+			}
+		}
+
+		if !filled {
+			if tp.Value == nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Parameter %s defined in template %s hasn't "+
+					"default value, so it should be specified in rule", tp.Name, templateName)
```
BupycHuk, line 390:
```suggestion
				return nil, status.Errorf(codes.InvalidArgument, "Parameter %s defined in template %s doesn't have "+
					"default value, so it should be specified in rule", tp.Name, templateName)
```

BupycHuk (approved):
Looks good to me, just a few minor comments

services/management/ia/rules_service.go
artemgavrilov: Fixed

services/management/ia/rules_service.go
artemgavrilov: Agree, done

services/management/ia/rules_service.go
artemgavrilov: `tp.Value` is the default value in fact, but that naming already spread across multiple repos

BupycHuk: ok

BupycHuk (approved):

PR #615, opened 2021-01-14
PMM-7145 Push mode by default.

JiriCtvrtka:
[PMM-7145](https://jira.percona.com/browse/PMM-7145)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1386

PMM: https://github.com/percona/pmm/pull/659
---
- [x] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


JiriCtvrtka:
UI part were fixed in grafana side. Contained in build:
https://github.com/Percona-Lab/pmm-submodules/pull/1386

askomorokhov (approved):

models/agent_helpers.go
```diff
@@ -359,8 +359,7 @@ func CreateNodeExporter(q *reform.Querier, pmmAgentID string, customLabels map[s
 		return nil, err
 	}
 	if !isPushMetricsSupported(pmmAgent.Version) {
-		return nil, errors.Errorf("cannot use push_metrics_enabled with pmm_agent version=%q,"+
-			" it doesn't support it, minimum supported version=%q", pointer.GetString(pmmAgent.Version), PMMAgentWithPushMetricsSupport.String())
+		pushMetrics = false
```
jprukner, line 362: How does user know whats wrong when no error is returned/printed? Or does it fallback to pull mode automatically?

JiriCtvrtka: @jprukner It will continue with pull mode.

jprukner (approved):
@JiriCtvrtka LGTM

models/agent_helpers.go
AlekSi: It should not continue with pull mode if push mode was requested explicitly.

That's not the right place to silently ignore passed parameter: that's a wrong layer, and that breaks Inventory API that should do what it is told to do, not try to be smart and silently "fix" things. Instead, the caller should make sure that correct parameter is passed.

services/management/management.go
```diff
@@ -130,5 +130,5 @@ func validateNodeParamsOneOf(nodeID, nodeName string, addNodeParams *managementp
 // for now only PUSH variant enables pushMode for the agent,
 // later it will be changed and auto will mean pushMode.
```
AlekSi, line 131: Please update that comment

services/management/mongodb.go
```diff
@@ -72,6 +72,10 @@ func (s *MongoDBService) Add(ctx context.Context, req *managementpb.AddMongoDBRe
 
 		mongoDBOptions := models.MongoDBOptionsFromRequest(req)
 
+		if req.AddNode.NodeType == inventorypb.NodeType_REMOTE_NODE {
```
AlekSi, line 75: Why do we check Node type, of all things, to determine if Agent should use pull mode?

AlekSi (changes requested):

services/management/rds.go
```diff
@@ -314,6 +314,17 @@ func (s *RDSService) AddRDS(ctx context.Context, req *managementpb.AddRDSRequest
 			}
 			res.Mysql = invService.(*inventorypb.MySQLService)
 
+			if isPushMode(req.MetricsMode) {
```
BupycHuk, line 317: id we always run agents on pmm-server we can always use pull mode here

models/agent_helpers.go
JiriCtvrtka: Moved.

services/management/management.go
JiriCtvrtka: Updated.

services/management/mongodb.go
JiriCtvrtka: In case of remote we always set pull mode because of "Add Instance" from UI.

services/management/postgresql.go
```diff
@@ -70,6 +70,20 @@ func (s *PostgreSQLService) Add(ctx context.Context, req *managementpb.AddPostgr
 		}
 		res.Service = invService.(*inventorypb.PostgreSQLService)
 
+		switch {
+		case req.AddNode != nil && req.AddNode.NodeType == inventorypb.NodeType_REMOTE_NODE:
```
BupycHuk, line 74: Now at the second sight it looks like it's better to check for `req.PmmAgentId == models.PMMServerAgentID`, because that's the reason why agents added via UI should be pull.
Sorry, for confusing you, again.

BupycHuk: here and in other places.

services/management/rds.go
BupycHuk: let's remove this if and always set metrics mode PULL

BupycHuk: and replace comment in pmm repo that we don't use this value and it's always pull https://github.com/percona/pmm/blob/PMM-2.0/api/managementpb/rds.proto#L128
@AlekSi what do you think?

services/management/node.go
```diff
@@ -119,6 +119,18 @@ func (s *NodeService) Register(ctx context.Context, req *managementpb.RegisterNo
 		if err != nil {
 			return err
 		}
+
```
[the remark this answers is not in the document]
BupycHuk: only here

BupycHuk (changes requested):

services/management/rds.go
AlekSi: That API does not allow one to start rds_exporter on any Node other than PMM Server Node? If yes, then we should always use pull for both pull and auto in that particular case, and return error if caller requested push. We can't silently ignore and "fix" user's error.

But in general, rds_exporter can be started on any Node via Inventory API.

BupycHuk: 1. yes, ok, let's return error and update comment in pmm repo.

services/management/external.go
```diff
@@ -82,6 +82,10 @@ func (e ExternalService) AddExternal(ctx context.Context, req *managementpb.AddE
 		}
 		res.Service = invService.(*inventorypb.ExternalService)
 
+		if isPushMode(req.MetricsMode) {
```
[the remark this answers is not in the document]
AlekSi: How?

services/management/external.go
```diff
@@ -82,6 +82,10 @@ func (e ExternalService) AddExternal(ctx context.Context, req *managementpb.AddE
...
+			req.MetricsMode = managementpb.MetricsMode_PULL
```
AlekSi, line 86: Independently from the comment above, *we must not overwrite mode if it is not auto*. If requested mode is not auto and not supported, we must return error. Everywhere.

services/management/external.go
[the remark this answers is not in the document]
BupycHuk: User can install pmm-agent into the same node as external exporter and add external exporter via `pmm-admin add external --metrics-mode push`. In that case vm-agent will collect metrics from external exporter locally and push to vmdb.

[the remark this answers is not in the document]
BupycHuk: So, here we should have the same check that we will have in other places.

services/management/mongodb.go
```diff
@@ -72,6 +72,20 @@ func (s *MongoDBService) Add(ctx context.Context, req *managementpb.AddMongoDBRe
 
 		mongoDBOptions := models.MongoDBOptionsFromRequest(req)
 
+		switch {
+		case req.AddNode != nil && req.AddNode.NodeType == inventorypb.NodeType_REMOTE_NODE:
+			req.MetricsMode = managementpb.MetricsMode_PULL
```
AlekSi, line 77: The same problem everywhere: we should not overwrite mode before checking that it is set to auto. If user requested push or pull explicitly, and this mode is not supported in this particular case, we should return error.

AlekSi (changes requested):

services/management/external.go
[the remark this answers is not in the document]
AlekSi: You are right. Please check that we document that, and that QA has test cases for that

services/management/mongodb.go
AlekSi: /cc @BupycHuk

BupycHuk: Agree

JiriCtvrtka:
@AlekSi @BupycHuk Please take a look. Now it should contains all required changes, if I didnt miss any.

services/management/mongodb.go
```diff
@@ -72,6 +72,22 @@ func (s *MongoDBService) Add(ctx context.Context, req *managementpb.AddMongoDBRe
 
 		mongoDBOptions := models.MongoDBOptionsFromRequest(req)
 
+		if req.MetricsMode == managementpb.MetricsMode_AUTO {
```
BupycHuk, line 75: Can we extract this block to some function, since we use it in 4 places?

JiriCtvrtka: Yes sure.

models/agent_helpers.go
```diff
@@ -655,11 +656,34 @@ func updateExternalExporterParams(q *reform.Querier, row *Agent) error {
...
 		if agentVersion.Less(PMMAgentWithPushMetricsSupport) {
 			return false
 		}
 	}
 	return true
 }
+
+// SupportedMetricsMode automatically pick metrics mode.
+func SupportedMetricsMode(q *reform.Querier, metricsMode managementpb.MetricsMode, pmmAgentID string) (managementpb.MetricsMode, error) {
+	if metricsMode != managementpb.MetricsMode_AUTO {
```
BupycHuk, line 671: let's add before this
```
if pmmAgentID == PMMServerAgentID && metricsMode == managementpb.MetricsMode_PUSH {
    return errors.Errorf("push metrics mode is not allowed for exporters running on pmm-server")
}
```

JiriCtvrtka: Case added. Also added this method in rds.

BupycHuk (changes requested):
external exporters are missed.

services/management/external.go
```diff
@@ -82,6 +83,20 @@ func (e ExternalService) AddExternal(ctx context.Context, req *managementpb.AddE
 		}
 		res.Service = invService.(*inventorypb.ExternalService)
 
+		agentIDs, err := models.FindPMMAgentsRunningOnNode(tx.Querier, req.RunsOnNodeId)
+		if err != nil {
```
BupycHuk, line 87: if we can't find pmm-agent we should set pull

services/management/external.go
```diff
@@ -82,6 +83,20 @@ func (e ExternalService) AddExternal(ctx context.Context, req *managementpb.AddE
...
+			return errors.Wrapf(err, "cannot find pmm_agent for external exporter with push_metrics")
+		}
+		if len(agentIDs) > 1 {
+			return errors.Errorf("cannot find exact match for pmm_agent for external exporter,"+
```
BupycHuk, line 91: if we have more than one let's set pull as well.

services/management/external.go
BupycHuk: if user didn't set metrics-mode push explicitly

services/management/external.go
BupycHuk: if user didn't set metrics-mode push explicitly

services/management/external.go
BupycHuk: if it's `pull`, just set pull, no need to find pmm-agent.

BupycHuk (changes requested):

services/management/external.go
JiriCtvrtka: Ok, behavior changed.

Case: AUTO && (agents count > 1 or err != nil) => PULL
Case: AUTO && count == 1 && err == nil => check for supported mode

Othervise metrcis-mode will be without change.

BupycHuk (approved):

askomorokhov (approved):

PR #617, opened 2021-01-18
PMM-7328 PMM-7343 force checks download and execution upon enabling STT, fix bug with stale STT results

palash25:
[PMM-7328](https://jira.percona.com/browse/PMM-7328)
[PMM-7343](https://jira.percona.com/browse/PMM-7343)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1405

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


palash25:
@AlekSi do we need an API test for this?

AlekSi:
If we can write it, then yes

palash25:
looking into the FB failure

services/server/server.go
```diff
@@ -596,6 +599,14 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
 		return nil, err
 	}
 
+	if settings.SaaS.STTEnabled {
```
BupycHuk, line 602: it will start checks on each settings update, is it ok? Maybe it's better to start checks only if we have enable STT request?

BupycHuk (changes requested):

services/server/server.go
palash25: you are right, will change it

palash25:
UI tests are failing because of a timeout

services/server/server.go
```diff
@@ -596,6 +599,14 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
 		return nil, err
 	}
 
+	if req.EnableStt {
+		// force checks download and execution when STT is enabled
+		err = s.checksService.StartChecks(ctx)
```
BupycHuk, line 604: looks like it's better to run this in a separate goroutine.

palash25: changed

palash25:
looking into FB failure

services/server/server.go
```diff
@@ -596,6 +599,16 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
 		return nil, err
 	}
 
+	if req.EnableStt {
```
artemgavrilov, line 602: I think it's better to move it below `s.UpdateConfiguration` and `isAgentsStateUpdateNeeded`. Then we will be sure that we start checks only when an update is successful.

artemgavrilov (approved):

services/server/server.go
palash25: done, waiting for the FB to complete

artemgavrilov (approved):

Dasio (approved):

services/server/server.go
```diff
@@ -600,6 +603,16 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
```
BupycHuk, line 606: Can we check that before STT was disabled? Because frontend always passes all fields.

artemgavrilov: We can get old settings before applying an update, and then compare STT status changes. Is it worth it?

BupycHuk: I think yes, because when user updates some other setting (e.g. metrics resolution) he don't expect that force check downloads will be executed

artemgavrilov: Done, also I fixed https://jira.percona.com/browse/PMM-7343 in this PR

BupycHuk (approved):

BupycHuk (changes requested):
Travis CI tests are failed. Probably you missed updating mocks

artemgavrilov:
Sorry, was concentrated on FB

BupycHuk (approved):

PR #625, opened 2021-01-26
Revert "PMM-7145 Push mode by default."

BupycHuk:
Reverts percona/pmm-managed#615


AlekSi:
Why?

BupycHuk:
To check if our tests are failing because of this PR and yes, it's because of metrics mode were changed to push

AlekSi:
So we probably should delete it?

BupycHuk:
Yes

PR #626, opened 2021-01-26
PMM-7232 Disabling IA continues sending

yareach:
[PMM-7232](https://jira.percona.com/browse/PMM-7232)

- [x] https://github.com/Percona-Lab/pmm-submodules/pull/1508

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/checks/checks.go
```diff
@@ -214,7 +214,12 @@ func (s *Service) resendAlerts(ctx context.Context) {
 	defer t.Stop()
 
 	for {
-		s.alertmanagerService.SendAlerts(ctx, s.alertsRegistry.collect())
+		settings, err := models.GetSettings(s.db)
+
+		// Only sends if IAs enabled in the settings
+		if (err == nil) && settings.IntegratedAlerting.Enabled {
+			s.alertmanagerService.SendAlerts(ctx, s.alertsRegistry.collect())
+		}
```
artemgavrilov, line 222: `alertsRegistry` contains only STT alerts, IA alerts sent to `alertmanager` directly from `vmalert`. Moreover now disabling IA will also disables STT, but it's two independent features, except they both use alertmanager. 

Tickets original text proposes more o less right solution. All generated files should be removed (but what we will do with user-created files?) and vmalert should be asked to reload the configuration.

artemgavrilov (changes requested):

services/checks/checks.go
yareach: Reworked. Please check.

services/server/server.go
```diff
@@ -611,6 +618,49 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+	}
+
+	return nil
+}
+
+// getAlertRuleIds lists the alert rules ids into an array.
+// Error is not returned, when occurres an empty array is returned.
+func (s *Server) getAlertRuleIds() []string {
+	var rules []*models.Rule
+	err := s.db.InTransaction(func(tx *reform.TX) error {
```
oter, line 644: why you are not returning this error?

services/server/server.go
```diff
@@ -611,6 +618,49 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+		var err error
+		rules, err = models.FindRules(tx.Querier)
+
+		return err
+	})
+
+	if err == nil {
+		ruleIds := make([]string, len(rules))
```
oter, line 652: minor: in Go it's common to have initialisms or acronyms consistent case, so `ruleIDs` instead of `ruleIds`. The same for the `alertRuleIds` above.
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

services/server/server.go
yareach: Do we have any plans with the error code? I can add it there, but as it is now, it will return an empty string in case of error which means it will not do anything.
Shall I add it?

services/server/server.go
yareach: I though that ID is for Identification Document where Id just an abbreviation of Identification.
I got you. Ids could also be understood as Idstein where they used to burn witches.
So I'll change it.

artemgavrilov (changes requested):
It doesn't work, I still see alerts. That's because you don't clean up actual files in directory `/etc/ia/rules`, you only removed rules from DB. But what is more important you should not delete anything from DB, in that case, we just losing user data.

When the user creates a rule, it's saved to DB, then we read all rules from the table and generate rule files in `/etc/ia/rules`. Then we ask `vmalert` to refresh its configuration, `vmlarts` reloads all rules from our directory and starts to execute them. So our problem is that after IA disabling `vmalert` continues to execute rules. So all that we need on IA disabling we should clean up `/etc/ia/rules` and ask `vmalert` to reload the configuration and on IA enabling we should regenerate all rules files from DB records. So the ticket is about `/etc/ia/rules` directory state.

Code for rule file generation: https://github.com/percona/pmm-managed/blob/787c43114f2bcdcce597e1081cceb2f9ea1f8b98/services/management/ia/rules_service.go#L96-L127

Please ask me questions if it's not clear to you how IA works or what you should do in that PR.

services/server/server.go
BupycHuk, line 644: Any reason to do this one command in transaction?
We shouldn't ignore errors, at least log them or better to return them to user.

services/server/server.go
```diff
@@ -611,6 +618,49 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
+func (s *Server) getAlertRuleIds() []string {
+	var rules []*models.Rule
+	err := s.db.InTransaction(func(tx *reform.TX) error {
+		var err error
+		rules, err = models.FindRules(tx.Querier)
+
+		return err
+	})
+
+	if err == nil {
```
BupycHuk, line 651: this check doesn't make sense because in Go if you return error you don't return result. 
and go usually ignores loops if you try to loop nil.

BupycHuk (changes requested):
Any reason to complicate logic by "Finding all Rules and then deleting by IDs" instead of just deleting all rules?

services/server/server.go
yareach: Does not exist anymore.

yareach: Does not exist anymore.

services/server/server.go
yareach: Does not exist anymore.

services/server/server.go
yareach: Does not exist anymore.

models/rule_helpers.go
```diff
@@ -221,6 +221,24 @@ func RemoveRule(q *reform.Querier, id string) error {
 	return nil
 }
 
+// RemoveRules removes alert rules whose ids are present in the ids array.
+// In case of success nil is returned, otherwise an error.
+func RemoveRules(q *reform.Querier, ids []string) error {
```
artemgavrilov, line 226: This func is unused and can be deleted

services/server/server.go
```diff
@@ -65,7 +66,11 @@ type Server struct {
 	platformService      platformService
 	awsInstanceChecker   *AWSInstanceChecker
 	grafanaClient        grafanaClient
-	l                    *logrus.Entry
+	RulesService         *ia.RulesService
+	AlertsService        *ia.AlertsService
```
BupycHuk, line 70: `AlertsService` and `TemplatesService` is not used in `Server`, let's remove it from here

services/server/server.go
```diff
@@ -65,7 +66,11 @@ type Server struct {
 	platformService      platformService
 	awsInstanceChecker   *AWSInstanceChecker
 	grafanaClient        grafanaClient
-	l                    *logrus.Entry
+	RulesService         *ia.RulesService
```
BupycHuk, line 69: let's keep all fields private.

main.go
```diff
@@ -180,11 +180,10 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps) {
 	managementpb.RegisterSecurityChecksServer(gRPCServer, managementgrpc.NewChecksServer(checksSvc))
 
 	iav1beta1.RegisterChannelsServer(gRPCServer, ia.NewChannelsService(deps.db, deps.alertmanager))
-	templatesSvc := ia.NewTemplatesService(deps.db)
-	templatesSvc.Collect(ctx)
-	iav1beta1.RegisterTemplatesServer(gRPCServer, templatesSvc)
-	iav1beta1.RegisterRulesServer(gRPCServer, ia.NewRulesService(deps.db, templatesSvc, deps.vmalert, deps.alertmanager))
-	iav1beta1.RegisterAlertsServer(gRPCServer, ia.NewAlertsService(deps.db, deps.alertmanager, templatesSvc))
+	deps.server.TemplatesService.Collect(ctx)
```
BupycHuk, line 183: templateService and other services can be passed as separate field in deps instead of exposing them

services/management/ia/rules_service.go
```diff
@@ -92,8 +92,26 @@ type rule struct {
...
+				break
+			}
+		}
+
+		s.vmalert.RequestConfigurationUpdate()
+		s.alertManager.RequestConfigurationUpdate()
+	}
+
+	return err
+}
```
artemgavrilov, line 111:
```suggestion
func (s *RulesService) RemoveVMAlertRulesFiles() {
	matches, err := filepath.Glob(s.rulesPath + "/*.yml")
	if err != nil {
		s.l.Errorf("Failed to clean old alert rule files: %+v", err)
		return
	}

	for _, match := range matches {
		if err = os.RemoveAll(match); err != nil {
			s.l.Errorf("Failed to remove old rule file: %+v", err)
		}
	}
}
```


It can be just copy-pasted lines 127-137 from `WriteVMAlertRulesFiles`. Also we can use `RemoveVMAlertRulesFiles` inside `WriteVMAlertRulesFiles` for cleanup.

services/management/ia/rules_service.go
```diff
@@ -92,8 +92,26 @@ type rule struct {
 	Annotations map[string]string   `yaml:"annotations,omitempty"`
 }
 
+// RemoveVMAlertRulesFiles removes all generated rules files (*.yml) on the ia path
+func (s *RulesService) RemoveVMAlertRulesFiles() error {
+	ruleFiles, err := filepath.Glob(rulesDir + "/*.yml")
+	if err == nil {
```
artemgavrilov, line 98: This is quiet unusual for golang. Better just return error ASAP:
```
if err != nil {
    return err
}
```

services/management/ia/rules_service.go
```diff
@@ -92,8 +92,26 @@ type rule struct {
...
+
+		for _, file := range ruleFiles {
+			if err = os.RemoveAll(file); err != nil {
+				break
+			}
+		}
+
+		s.vmalert.RequestConfigurationUpdate()
+		s.alertManager.RequestConfigurationUpdate()
```
artemgavrilov, line 107: We can remove it from here, it will be invoked anyway on line:
https://github.com/percona/pmm-managed/blob/5d330f5af2b646a0cf647315d236137edd3aca4d/services/server/server.go#L599

services/server/server.go
```diff
@@ -93,6 +98,9 @@ type Params struct {
 	PlatformService      platformService
 	AwsInstanceChecker   *AWSInstanceChecker
 	GrafanaClient        grafanaClient
+	RulesService         *ia.RulesService
+	AlertsService        *ia.AlertsService
+	TemplatesService     *ia.TemplatesService
```
artemgavrilov, line 103: It seems that you don't need these two services here.

artemgavrilov (commented):
Yep, it works! But some improvements can be added.

BupycHuk (changes requested):

artemgavrilov:
Also CI is failing

services/server/server.go
yareach: I wanted to keep them private, but that I will have to create functions to access them from outside. Do you prefer having a public function that will return the the rules service?

BupycHuk: what is the reason to get `rulesService` from `Server`?

BupycHuk (changes requested):
still didn't get what is the reason to get `RulesService` from `Server`

BupycHuk:
CI are still failing

services/server/server.go
yareach: We need to call the its instance functions when deleting or regenerating the rules files: https://github.com/percona/pmm-managed/blob/PMM-7232-Disabling-IA-continues-sending/services/server/server.go#L581

BupycHuk: It can be private there to call.

yareach: It then fails in main.go when registered.
image

BupycHuk: don't get `rulesService` from `deps.server`, in main.go you have direct access to `rulesService`

services/server/server.go
```diff
@@ -47,6 +47,7 @@ import (
 	"gopkg.in/reform.v1"
 
 	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/services/management/ia"
```
BupycHuk, line 50: We also try to not import one package from another to avoid risks of circle dependencies. Instead of that we use interfaces and mocks. Please check other services in the same file.

yareach: I was forgotten there. For some reason it was not removed automatically on save.

BupycHuk: it's imported because you use it https://github.com/percona/pmm-managed/pull/626/files#diff-a990b29ab7e542cbefea97ccf6055f35a8f1486f1af794294dc542bf9e0e0b25R69

services/server/server.go
askomorokhov: I do not see any big reason to make field public too.

main.go
```diff
@@ -664,15 +676,17 @@ func main() {
...
-		l:            logrus.WithField("component", "setup"),
+		sqlDB:            sqlDB,
+		dbUsername:       *postgresDBUsernameF,
+		dbPassword:       *postgresDBPasswordF,
+		supervisord:      supervisord,
+		vmdb:             vmdb,
+		vmalert:          vmalert,
+		alertmanager:     alertmanager,
+		server:           server,
+		l:                logrus.WithField("component", "setup"),
```
BupycHuk, line 687: let's keep logger as the last element here

main.go
```diff
@@ -96,6 +96,9 @@ const (
 	cleanOlderThan = 30 * time.Minute
 )
 
+// Pointer to the rules service needed at the time of registration
+var rulesSvc *ia.RulesService
```
BupycHuk, line 100: no reason to use global variable here.

BupycHuk (changes requested):
let's write some tests

main.go
yareach: It is not global, it is only accessible within the module.
The variable is initialized in main() and used in runGRPCServer(). How would you like to do that?

BupycHuk: Like we do with other services, please read the code.

artemgavrilov: @yareach pass it form `main` to `runGRPCServer()` with `gRPCServerDeps`. Same as we do for `alertsService` for example.

yareach:
> let's write some tests

@BupycHuk Can you please specify "some tests"? A function I add "RemoveVMAlertRulesFiles" is called from the WriteVMAlertRulesFiles function which is tested. Is it necessary to write a test for it?

BupycHuk (changes requested):
1. I don't see test where we check if file is deleted
2. What about tests for Settings update?
3. Each settings update request will regenerate rule files even if we don't enable or disable IA. @artemgavrilov can point to a right PR.

services/server/server.go
```diff
@@ -572,6 +575,14 @@ func (s *Server) ChangeSettings(ctx context.Context, req *serverpb.ChangeSetting
...
 		}
 
+		if req.DisableAlerting {
+			// Deletes the alert rules files in case of disabling
+			s.rulesService.RemoveVMAlertRulesFiles()
+		} else if req.EnableAlerting {
+			// Regenerates the rules files in case of enabling
+			s.rulesService.WriteVMAlertRulesFiles()
+		}
+
```
artemgavrilov, line 585: Let's do this things only when alerts state actually changed (UI always sends full set of setting)

Similar changes: https://github.com/percona/pmm-managed/pull/617/files#diff-a990b29ab7e542cbefea97ccf6055f35a8f1486f1af794294dc542bf9e0e0b25R611

services/management/ia/rules_service.go
```diff
@@ -92,8 +92,22 @@ type rule struct {
 	Annotations map[string]string   `yaml:"annotations,omitempty"`
 }
 
+// RemoveVMAlertRulesFiles removes all generated rules files (*.yml) on the ia path.
+func (s *RulesService) RemoveVMAlertRulesFiles() {
```
artemgavrilov, line 96: Let's add some tests for that function in `rules_service_test.go`. There you can find similar tests for `WriteVMAlertRulesFiles()`

yareach: I added check that the files are delete after they are created. Please, check if it is sufficient.

services/server/server.go
yareach: Done

services/management/ia/rules_service_test.go
```diff
@@ -146,6 +147,13 @@ groups:
 `, ruleID, ruleID)
 
 		assert.Equal(t, expected, string(file))
+
+		// Removes the rules files
+		rules.RemoveVMAlertRulesFiles()
+
+		matches, err := filepath.Glob(testDir + "/*.yml")
```
artemgavrilov, line 154: Missing error check

artemgavrilov (approved):
LGTM with comment

Please prepare feature build

artemgavrilov:
Also please resolve merge conflicts

Dasio (commented):
Can we add to test to check if after disabling IA, it removes vm alert rules files?

JiriCtvrtka:
@yareach Please take a look on these errors:
`services/server/server.go:628:6: undefined: oldSettings
services/server/server.go:628:37: undefined: newSettings
services/server/server.go:638:5: undefined: oldSettings
services/server/server.go:638:37: undefined: newSettings
services/server/server.go:649:31: undefined: newSettings`

Build failed.

yareach:
> Can we add to test to check if after disabling IA, it removes vm alert rules files?

The check is here: https://github.com/percona/pmm-managed/pull/626/files#diff-483bfd6b08e669264eeba849f634889e1bba188808fd7fae99f958d4c52b9a4dR152

Dasio:
> > Can we add to test to check if after disabling IA, it removes vm alert rules files?
> 
> The check is here: https://github.com/percona/pmm-managed/pull/626/files#diff-483bfd6b08e669264eeba849f634889e1bba188808fd7fae99f958d4c52b9a4dR152

No, it just tests that, after calling `rules.RemoveVMAlertRulesFiles()` rules are removed from filesystem. Not that it is called when IA is being disabled.

services/server/server_test.go
```diff
@@ -219,8 +219,40 @@ func TestServer(t *testing.T) {
...
+		require.NoError(t, err)
+		require.NotNil(t, s)
+
+		rs.On("RemoveVMAlertRulesFiles")
+		s, err = server.ChangeSettings(ctx, &serverpb.ChangeSettingsRequest{
+			DisableAlerting: true,
+		})
+		require.NoError(t, err)
+		require.NotNil(t, s)
+		rs.AssertExpectations(t)
```
Dasio, line 256: double assert expectations

Dasio (approved):

services/management/ia/rules_service_test.go
```diff
@@ -142,6 +143,14 @@ groups:
...
+
+		assert.Zero(t, len(matches))
```
artemgavrilov, line 152:
```suggestion
		assert.Empty(t, matches)
```

artemgavrilov (changes requested):
Please remove `pmm-managed` binary from PR. Also, don't close feature build PR until your changes are merged (plz recreate it)

Other things LGTM

artemgavrilov (approved):

services/management/ia/rules_service.go
```diff
@@ -92,8 +92,22 @@ type rule struct {
...
+	matches, err := filepath.Glob(s.rulesPath + "/*.yml")
+	if err != nil {
+		s.l.Errorf("Failed to clean old alert rule files: %+v", err)
```
BupycHuk, line 99: it's better to return some error here, not just log and go ahead

BupycHuk (approved):

services/management/ia/rules_service.go
artemgavrilov: Done

PR #629, opened 2021-01-26
PMM-7306 Handle disabled collectors

askomorokhov:
[PMM-7306](https://jira.percona.com/browse/PMM-7306)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1436

https://github.com/percona/pmm/pull/660
https://github.com/percona/pmm-admin/pull/131

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [x] (Re)requested review.
- [ ] Update API dependency.

image

image


JiriCtvrtka:
@askomorokhov Please rerun Travis.

services/agents/agents.go
```diff
@@ -40,3 +43,28 @@ func redactWords(agent *models.Agent) []string {
...
 	return words
 }
+
+func filterOutCollectors(prefix string, args, disabledCollectors []string) []string {
+	isDisabled := func(arg string) bool {
+		for _, disabledCollector := range disabledCollectors {
+			// DisableCollector valuse should  match collector flag till end of string or till `=` sign.
+			// Examples:
+			// 1. if we pass `meminfo` then only "--collector.meminfo" but not "--collector.meminfo_numa"
+			// 2. if we pass `netstat.field` then "--collector.netstat.fields=^(.*_(InErrors|InErrs|InCsumErrors)..." shold be disabled.
```
JiriCtvrtka, line 53: Small typo. "Should be disabled".

JiriCtvrtka (approved):

services/agents/agents.go
```diff
@@ -40,3 +43,28 @@ func redactWords(agent *models.Agent) []string {
 	}
 	return words
 }
+
+func filterOutCollectors(prefix string, args, disabledCollectors []string) []string {
+	isDisabled := func(arg string) bool {
```
BupycHuk, line 48: maybe it's better to create `map[string]struct{}` at the beginning and just check if key is exists instead of looping n*m times.

services/agents/proxysql.go
```diff
@@ -44,6 +44,8 @@ func proxysqlExporterConfig(service *models.Service, exporter *models.Agent, red
 		args = append(args, "-web.telemetry-path="+*exporter.MetricsPath)
 	}
 
+	args = filterOutCollectors("--collect", args, exporter.DisabledCollectors)
```
BupycHuk, line 47:
```suggestion
	args = filterOutCollectors("-collect", args, exporter.DisabledCollectors)
```

BupycHuk (changes requested):
we also should update scrape configs.

services/agents/agents.go
askomorokhov: Yes, think on this as well - but without more complex refactoring this does not much bring much effort.
And as addition, we cannot match key by it part (`strings.HasPrefix`) - so it still requires iterations.

BupycHuk: as far as I see all collectors has only two options:
boolean flag like `--collect.binlog_size`
value flags like `--collect.custom_query.hr.directory=/usr/local/percona/pmm2/collectors/custom-queries/mysql/high-resolution`
for boolean flags we can just concatinate prefix with disabled collector and just compare two values
for value flags we can extract first part before `=` symbol and use it to compare. 
what do you think?

askomorokhov: Yes, true. Good point - will change. 👍

models/database.go
```diff
@@ -419,6 +419,9 @@ var databaseSchema = [][]string{
 	26: {
 		`ALTER TABLE ia_rules ALTER COLUMN channel_ids DROP NOT NULL`,
 	},
+	27: {
+		`ALTER TABLE agents ADD COLUMN disabled_collectors VARCHAR []`,
```
BupycHuk, line 423:
```suggestion
		`ALTER TABLE agents ADD COLUMN disabled_collectors VARCHAR[]`,
```

services/agents/agents.go
```diff
@@ -40,3 +43,32 @@ func redactWords(agent *models.Agent) []string {
...
+			delete(argsMap, key)
+		}
+	}
+
+	enabledArgs := []string{}
+	for _, arg := range argsMap {
+		enabledArgs = append(enabledArgs, arg)
+	}
+
+	return enabledArgs
```
BupycHuk, line 73:
```suggestion
	disabledCollectorsMap := make(map[string]string)
	for _, disabledCollector := range disabledCollectors {
		key := fmt.Sprintf("%s%s", prefix, disabledCollector)
		disabledCollectorsMap[key] = struct{}
	}

	enabledArgs := []string{}
	for _, arg := range args {
		flagName := strings.Split(arg, "=")[0]
		if _, ok := disabledCollectorsMap[flagName]; !ok {
			enabledArgs = append(enabledArgs, arg)
		}
	}

	return enabledArgs
```

BupycHuk: The same in the other place. But I'm almost sure that we can extract this to utils and call it from both places.

BupycHuk (changes requested):

BupycHuk (approved):

PR #640, opened 2021-02-01
PMM-7380 DB Model

Dasio:
[PMM-7380](https://jira.percona.com/browse/PMM-7380)

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/backup_model.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+//go:generate reform
+
+// BackupLocationType represents BackupLocation type as stored in database.
+type BackupLocationType string
+
+// BackupLocation types
+const (
+	S3BackupLocationType NodeType = "s3"
```
artemgavrilov, line 32: Seems that type here should be `BackupLocationType`

```suggestion
	S3BackupLocationType BackupLocationType = "s3"
```

models/backup_model.go
```diff
@@ -0,0 +1,78 @@
...
+	S3BackupLocationType BackupLocationType = "s3"
+	FSBackupLocationType BackupLocationType = "fs"
+)
+
+// BackupLocation represents destination for backup.
+//reform:backup_locations
+type BackupLocation struct {
+	ID          string             `reform:"id,pk"`
+	Name        string             `reform:"name"`
+	Description *string            `reform:"description"`
```
artemgavrilov, line 41: Why it's a pointer?

models/backup_model.go
```diff
@@ -0,0 +1,78 @@
...
+	Type        BackupLocationType `reform:"type"`
+	Endpoint    *string            `reform:"endpoint"`
+	AccessKey   *string            `reform:"access_key"`
+	SecretKey   *string            `reform:"secret_key"`
+	Path        *string            `reform:"path"`
```
artemgavrilov, line 46: I think it better to create separate config models and store them as jsons. Same as https://github.com/percona/pmm-managed/blob/787c43114f2bcdcce597e1081cceb2f9ea1f8b98/models/channel.go#L46-L49

models/backup_model.go
```diff
@@ -0,0 +1,78 @@
...
+
+	CreatedAt time.Time `reform:"created_at"`
+	UpdatedAt time.Time `reform:"updated_at"`
```
atymchuk, line 49: @artemgavrilov We will have an audit requirement very soon, maybe we should provision for it now?

models/backup_model.go
Dasio: I thought when it's option then it's pointer.
https://github.com/percona/pmm-managed/blob/PMM-2.0/models/agent_model.go#L98

models/backup_model.go
artemgavrilov: Yep, I think it's a good idea 👍

atymchuk: I take it back - this should go to the Audit component, which will be designed later down the road.

Dasio: Auditing is out of scope of this issue/feature.
Would be handled separetely.

models/backup_model.go
artemgavrilov: Let's make this filed `string`, we don't need to distinguish between nil and empty sting here. And it's easier to use simple strings.

artemgavrilov (approved):

oter (approved):

PR #646, opened 2021-02-04
PMM-7467 Fix bug with angle brackets escaping in rule expressions

artemgavrilov:
[PMM-7467](https://jira.percona.com/browse/PMM-7467)

https://github.com/Percona-Lab/pmm-api-tests/pull/182

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1464

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.

### DIFF
--- a/services/management/ia/templates_service.go
+++ b/services/management/ia/templates_service.go
@@ -19,12 +19,12 @@ package ia
 import (
 	"bytes"
 	"context"
-	"html/template"
 	"io/ioutil"
 	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
+	"text/template"
 	"time"
 
 	"github.com/percona-platform/saas/pkg/alert"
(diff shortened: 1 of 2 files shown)


services/management/ia/templates_service.go
```diff
@@ -19,12 +19,12 @@ package ia
 import (
 	"bytes"
 	"context"
-	"html/template"
 	"io/ioutil"
 	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
+	"text/template"
```
AlekSi, line 27: Epic 😆 Please check other places too

artemgavrilov: Already checked, seems that it the only place where we execute expression)

Dasio (approved):
What about already existing templates. Can we fix it somehow?

Or we don't care about it sine it's beta.

artemgavrilov:
> What about already existing templates. Can we fix it somehow?
> 
> Or we don't care about it sine it's beta.

That bug affects rules, not templates. If user created a rule that has `<` it expression, then he cannot use it anyway. So yes, he should just recreate the rule, because as you said it's beta and the fix will be too complex for that.

BupycHuk (approved):

oter (approved):

PR #377, opened 2020-04-29
PMM-5777 Add external exporter to inventory API.

BupycHuk:

PR #657, opened 2021-02-11
PMM-7366 Track consistency for IA entities

artemgavrilov:
[PMM-7366](https://jira.percona.com/browse/PMM-7366)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1493

- [ ] https://github.com/Percona-Lab/pmm-api-tests/pull/186
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/management/ia/alerts_service.go
```diff
@@ -116,12 +116,19 @@ func (s *AlertsService) ListAlerts(ctx context.Context, req *iav1beta1.ListAlert
 				return err
 			})
 			if e != nil {
+				// Only FindRuleByID from transaction above can return codes.NotFound
+				if st, ok := status.FromError(e); ok && st.Code() == codes.NotFound {
```
artemgavrilov, line 120: When a rule deleted associated alert may be still present for several seconds (alertmanager delay). So instead of returning error better just to skip such alerts.

services/management/ia/alerts_service.go
```diff
@@ -116,12 +116,19 @@ func (s *AlertsService) ListAlerts(ctx context.Context, req *iav1beta1.ListAlert
...
+					s.l.Warnf("The related alert rule most likely was removed: %s", st.Message())
+					continue
+				}
```
artemgavrilov, line 123: When a rule deleted associated alert may be still present for several seconds (alertmanager delay). So instead of returning error better just to skip such alerts.

services/management/ia/alerts_service.go
```diff
@@ -116,12 +116,19 @@ func (s *AlertsService) ListAlerts(ctx context.Context, req *iav1beta1.ListAlert
...
+
 				return nil, e
 			}
 
 			template, ok := s.templatesService.getTemplates()[r.TemplateName]
 			if !ok {
-				return nil, status.Errorf(codes.NotFound, "Failed to find template with name: %s", r.TemplateName)
+				s.l.Warnf("Failed to find template with name: %s", r.TemplateName)
+				continue
```
artemgavrilov, line 131: This case possible when a rule created from the template that was created by the user as a file under `/srv/ia/templates/`. So when the user removes the template file the associated rule will be broken and the rule file will be removed soon. But the alert will be present for several seconds (alertmanager delay), better just to skip it.

services/management/ia/alerts_service.go
```diff
@@ -116,12 +116,19 @@ func (s *AlertsService) ListAlerts(ctx context.Context, req *iav1beta1.ListAlert
 				return err
 			})
 			if e != nil {
+				// Only FindRuleByID from transaction above can return codes.NotFound
```
yareach, line 119: This comment should be rephrased. What did you wanted to say here?

services/management/ia/alerts_service.go
```diff
@@ -116,12 +116,19 @@ func (s *AlertsService) ListAlerts(ctx context.Context, req *iav1beta1.ListAlert
...
+				if st, ok := status.FromError(e); ok && st.Code() == codes.NotFound {
+					s.l.Warnf("The related alert rule most likely was removed: %s", st.Message())
```
yareach, line 121: Words swap: The related alert rule WAS MOST LIKELY removed.

services/management/ia/alerts_service.go
artemgavrilov: The transaction above has two operations `models.FindRuleByID` and `models.FindChannelsByIDs`. Only first one can return NotFound error, so we can log message about a `rule`, not about `rule or channel`

models/channel_helpers.go
```diff
@@ -324,9 +324,30 @@ func RemoveChannel(q *reform.Querier, id string) error {
 		return err
 	}
 
-	err := q.Delete(&Channel{ID: id})
+	inUse, err := channelInUse(q, id)
```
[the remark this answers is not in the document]
artemgavrilov: This is harder to read, as for me:
```
	if inUse, err := channelInUse(q, id); err != nil {
		return err
	} else if inUse {
		return status.Errorf(codes.FailedPrecondition, "Failed to delete notification channel %s, as it is being used by some rule.", id)
	}
```

models/channel_helpers.go
```diff
@@ -324,9 +324,30 @@ func RemoveChannel(q *reform.Querier, id string) error {
...
+	if err != nil {
+		return err
+	}
+
+	if inUse {
+		return status.Errorf(codes.FailedPrecondition, "Failed to delete notification channel %s, as it is being used by some rule.", id)
+	}
+
+	err = q.Delete(&Channel{ID: id})
```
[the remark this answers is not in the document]
artemgavrilov: Done

services/management/ia/alerts_service.go
artemgavrilov: Fixed

models/rule_helpers.go
```diff
@@ -91,7 +91,8 @@ type CreateRuleParams struct {
 // CreateRule persists alert Rule.
 func CreateRule(q *reform.Querier, params *CreateRuleParams) (*Rule, error) {
 	id := "/rule_id/" + uuid.New().String()
-	if err := checkUniqueRuleID(q, id); err != nil {
+	var err error
```
Dasio, line 94: why declare it here and not directly in if statement?
same in remove rule

Dasio (approved):
LGTM, just i prefer
```go
if err := do(); err != nil {
}
```
than
```go
var err error
if err = do(); err != nil {
}
```

models/rule_helpers.go
artemgavrilov: Because there are other similar statements in that function, and each will allocate a new error.  By predeclaring `err` variable we can just reuse it each time.

artemgavrilov: In terms of you examples we choose between
```
if err := do1(); err != nil {
}
if err := do1(); err != nil {
}
```
and
```
var err error
if err = do1(); err != nil {
}
if err = do2(); err != nil {
}
```

Dasio: It's micro-optimization.
Anyway we are doing it at many places, like https://github.com/percona/pmm-managed/blob/PMM-2.0/models/agent_helpers.go#L331
Also in go srcs, e.g this test https://github.com/golang/go/blob/master/src/os/example_test.go#L32

artemgavrilov: In general, I'm trying to minimize allocations when it doesn't reduce code clearness. I agree that this is not important here, so I can change, it's not that big a deal)

AlekSi: The bigger issue with `if err :=` is shadowing – there are several variables with the same name. See also Go issue #377

services/management/ia/alerts_service.go
oter, line 120: Maybe it's time to define ErrNotFound in the models pkg and start using it instead of grpc status error?

artemgavrilov: Currently, we have a custom string for each NotFound error. If we will define package-level error, then its message always will be the same.

oter: For such cases you can use `errors.Wrap()` e.g. `errors.Wrap(ErrNotFound, "item not found")`. As a result you will get `item not found: not found`.

artemgavrilov: In that case, the returned error will be not a gRPC error. Also, the message is "inside-out".

services/management/ia/alerts_service.go
yareach: What about this? The codes.NotFound code can be returned just only by the FindRulesByID method/function (used above).

artemgavrilov: To be honest, I don't see much difference, but ok. Done.

services/management/ia/alerts_service.go
oter: And then you handle model error and convert it to appropriate gRPC error. 
My main point here is it's a bit weird to handle gRPC error that returned from the storage layer.

oter (approved):

services/management/ia/alerts_service.go
artemgavrilov: Well, maybe you right, but then we should refactor all other similar places. Let's leave it for the next time, such refactoring should be done in separate PR.

oter: resolving thread for now

BupycHuk (approved):

PR #658, opened 2021-02-12
PMM-7382 Change Backup Location

Dasio:
[PMM-7382](https://jira.percona.com/browse/PMM-7382)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1497

- API - https://github.com/percona/pmm/pull/682
- API tests - https://github.com/Percona-Lab/pmm-api-tests/pull/190
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


Dasio:
Forgot about submodules, will prepare it at monday with api-tests.

models/location_helpers.go
```diff
@@ -111,18 +111,93 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
 	return locations, nil
 }
 
+// FindBackupLocationByID finds Location by ID.
```
yareach, line 114: Is Location name? If not, it would be better: finds a location by its ID.

models/location_helpers.go
```diff
@@ -111,18 +111,93 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
+	}
+
+	if configCount > 1 {
+		return status.Error(codes.InvalidArgument, "Only one config is allowed.")
+	}
+
+	return nil
+}
+
+// FillLocation fills provided location based on config.
```
yareach, line 167: I'd clarify this comment: sets the pointed location's data according to the backup locations config.

models/location_helpers.go
```diff
@@ -111,18 +111,93 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
 	Name        string
 	Description string
 
-	PMMClientConfig *PMMClientLocationConfig
-	PMMServerConfig *PMMServerLocationConfig
-	S3Config        *S3LocationConfig
+	BackupLocationsConfigs
 }
 
 // CreateBackupLocation persists backup location.
```
yareach, line 195: "persists" should be replaced by something like "creates", "makes", "defines".

models/location_helpers.go
```diff
@@ -138,51 +213,61 @@ func CreateBackupLocation(q *reform.Querier, params CreateBackupLocationParams)
...
 
-	switch {
-	case params.S3Config != nil:
-		if err := checkS3Config(params.S3Config); err != nil {
-			return nil, err
-		}
-		row.Type = S3BackupLocationType
-		row.S3Config = params.S3Config
+	// remove previous configuration
+	row.Type = ""
```
yareach, line 248: What about writing a short function wipe the row configuration?

services/management/backup/locations_service.go
```diff
@@ -93,6 +93,40 @@ func (s *LocationsService) AddLocation(ctx context.Context, req *backupv1beta1.A
...
+	}
+
+	if req.S3Config != nil {
+		params.S3Config = &models.S3LocationConfig{
+			Endpoint:  req.S3Config.Endpoint,
+			AccessKey: req.S3Config.AccessKey,
+			SecretKey: req.S3Config.SecretKey,
+		}
+	}
+	if req.PmmServerConfig != nil {
```
yareach, line 110: Add a line to keep your style of if blocks separated by a line.

AlekSi:
@Dasio please resolve merge conflicts

artemgavrilov (approved):

models/location_helpers.go
```diff
@@ -111,18 +111,93 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
+		}
+	}
+	if c.PMMClientConfig != nil {
+		configCount++
+		if err := checkPMMClientLocationConfig(c.PMMClientConfig); err != nil {
+			return err
+		}
+	}
+
+	if configCount > 1 {
```
oter, line 160: in case all the configs are nil Validate() will return nil.

models/location_helpers.go
```diff
@@ -111,18 +111,93 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
+// BackupLocationsConfigs groups all backup locations configs.
+type BackupLocationsConfigs struct {
+	PMMClientConfig *PMMClientLocationConfig
+	PMMServerConfig *PMMServerLocationConfig
+	S3Config        *S3LocationConfig
+}
+
+// Validate checks if there is exactly one config with required fields.
+func (c BackupLocationsConfigs) Validate() error {
+	configCount := 0
```
oter, line 140: maybe better to check that only one config is set and then validate that config? As a user I can accidentally specify 2 configs and let's say some field in the second config is wrong, so I will try to fix it and only after that I will get an error that I'm setting more than one config.

models/location_helpers.go
```diff
@@ -111,18 +111,91 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
 	return locations, nil
 }
 
+// FindBackupLocationByID finds a Backup Location by it's ID.
```
yareach, line 114: Remove the apostrophe (by its ID).

models/location_helpers.go
BupycHuk: Agree with @oter

models/location_helpers.go
```diff
@@ -111,18 +111,91 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
+
 // CreateBackupLocationParams are params for creating new backup location.
 type CreateBackupLocationParams struct {
 	Name        string
 	Description string
 
-	PMMClientConfig *PMMClientLocationConfig
-	PMMServerConfig *PMMServerLocationConfig
-	S3Config        *S3LocationConfig
+	BackupLocationsConfigs
```
BupycHuk, line 190:
```suggestion
	BackupLocationConfig
```

BupycHuk: We expect that there will be only one config

models/location_helpers.go
```diff
@@ -138,51 +211,57 @@ func CreateBackupLocation(q *reform.Querier, params CreateBackupLocationParams)
...
+	// remove previous configuration
+	row.Type = ""
+	row.PMMClientConfig = nil
+	row.PMMServerConfig = nil
+	row.S3Config = nil
 
-	case params.PMMClientConfig != nil:
-		if err := checkPMMClientLocationConfig(params.PMMClientConfig); err != nil {
+	if params.Name != "" && params.Name != row.Name {
+		row.Name = params.Name
```
BupycHuk, line 250: let's move it after the if statement below to make it clear.

models/location_helpers.go
```diff
@@ -138,51 +211,57 @@ func CreateBackupLocation(q *reform.Querier, params CreateBackupLocationParams)
...
+	}
 
-	case params.PMMServerConfig != nil:
-		if err := checkPMMServerLocationConfig(params.PMMServerConfig); err != nil {
-			return nil, err
-		}
-		row.Type = PMMServerBackupLocationType
-		row.PMMServerConfig = params.PMMServerConfig
+	// remove previous configuration
+	row.Type = ""
```
BupycHuk, line 244: so we always expect that we have some config in params even when we want to change only name?

services/management/backup/locations_service_test.go
```diff
@@ -167,5 +168,74 @@ func TestListBackupLocations(t *testing.T) {
 		assert.Condition(t, checkLocation(res2.LocationId, req2))
 
 	})
+}
+
+func TestUpdateBackupLocation(t *testing.T) {
```
BupycHuk, line 173: `Update` or `Change`? let's use same naming everywhere

BupycHuk (changes requested):

models/location_helpers.go
Dasio: Yes, I was following similar funcition in IA channels.

BupycHuk: Then there is possibility to remove all configs by just sending empty request to this endpoint.
https://github.com/percona/pmm-managed/pull/658/files#diff-72bd90d93d5f3b4bf4a97c3640b8638d111ee250ffd6c77cd54565860912ff5bR161.

And that's where we have a problem in our teams structure and communication: In other change endpoints we change only that fields that we got in request.

models/location_helpers.go
```diff
@@ -111,18 +111,91 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
+		err = checkPMMServerLocationConfig(c.PMMServerConfig)
+	}
+
+	if c.PMMClientConfig != nil {
+		configCount++
+		err = checkPMMClientLocationConfig(c.PMMClientConfig)
+	}
+
+	if configCount == 0 {
+		return status.Error(codes.InvalidArgument, "Missing location type.")
```
BupycHuk, line 158: maybe it's better to pass `missing location config`?

models/location_helpers.go
BupycHuk: > In other change endpoints we change only that fields that we got in request. 
So let's keep the same logic in new Change API endpoints to not confuse our customers..

models/location_helpers.go
```diff
@@ -231,7 +237,8 @@ type ChangeBackupLocationParams struct {
 
 // ChangeBackupLocation updates existing location by specified locationID and params.
 func ChangeBackupLocation(q *reform.Querier, locationID string, params ChangeBackupLocationParams) (*BackupLocation, error) {
-	if err := params.Validate(); err != nil {
+	configCount, err := params.Validate()
```
oter, line 240: not sure you need to know how many configs are set, is set or not would be enough.

oter (approved):

models/location_helpers.go
```diff
@@ -111,18 +111,97 @@ func FindBackupLocations(q *reform.Querier) ([]*BackupLocation, error) {
...
+		err = checkPMMServerLocationConfig(c.PMMServerConfig)
+	}
+
+	if c.PMMClientConfig != nil {
+		configCount++
+		err = checkPMMClientLocationConfig(c.PMMClientConfig)
+	}
+
+	if configCount > 1 {
+		return configCount == 1, status.Error(codes.InvalidArgument, "Only one config is allowed.")
```
BupycHuk, line 158: `configCount == 1` is always false here

BupycHuk (approved):
LGTM, can we also add one more test to check that config will not be overwritten if it's not exist in request?

PR #669, opened 2021-02-18
PMM-7231 IA Disable channel

Dasio:
[PMM-7231](https://jira.percona.com/browse/PMM-7231)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1537

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/alertmanager/alertmanager.go
```diff
@@ -359,20 +359,33 @@ func (svc *Service) populateConfig(cfg *alertmanager.Config) error {
 		cfg.Global = &alertmanager.GlobalConfig{}
 	}
 
-	// make sure that "empty" receiver is there
-	var emptyFound bool
-	for _, r := range cfg.Receivers {
-		if r.Name == "empty" {
-			emptyFound = true
-			break
+	findReceiver := func(name string) int {
```
BupycHuk, line 362: findReceiverIdx?

services/alertmanager/alertmanager.go
```diff
@@ -359,20 +359,33 @@ func (svc *Service) populateConfig(cfg *alertmanager.Config) error {
...
 			Name: "empty",
 		})
 	}
 
+	disabledReceiver := &alertmanager.Receiver{
+		Name: "disabled",
+	}
+	disabledReceiverIndex := findReceiver("disabled")
+	if disabledReceiverIndex >= 0 {
+		cfg.Receivers[disabledReceiverIndex] = disabledReceiver
```
BupycHuk, line 384: why do we update it here?

services/alertmanager/alertmanager.go
BupycHuk: based on the question below looks like we can return boolean here

services/alertmanager/alertmanager.go
Dasio: so user can't add his "disabled" receiver by using file.
Same thing as he can add "empty" (default) receive.

BupycHuk: Yeah, but why do we replace it?

Dasio: Any better solution how to forbid user to add this receiver?
Check after yaml is parsed? And what then?

BupycHuk: ok, so we let users to override empty, but don't let to override disabled, right?
If it is, then everything is ok, but we should document it.

artemgavrilov: I agree, better to add a comment here.

artemgavrilov (approved):
Please add comments about `disabled` receivers/

Other stuff LGTM

oter (approved):

services/alertmanager/alertmanager.go
```diff
@@ -359,20 +359,32 @@ func (svc *Service) populateConfig(cfg *alertmanager.Config) error {
...
+	if findReceiverIdx("empty") == -1 {
 		cfg.Receivers = append(cfg.Receivers, &alertmanager.Receiver{
 			Name: "empty",
 		})
 	}
 
+	disabledReceiver := &alertmanager.Receiver{
+		Name: "disabled",
+	}
+	// Override if there is any user defined receiver `disabled, needs to be empty
```
BupycHuk, line 381:
```suggestion
	// Override if there is any user defined receiver `disabled`, needs to be empty
```

BupycHuk (approved):

PR #678, opened 2021-02-23
PMM-7379 Edit alert rule name

Dasio:
[PMM-7379](https://jira.percona.com/browse/PMM-7379)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1549

API - https://github.com/percona/pmm/pull/692
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/rule_helpers.go
```diff
@@ -192,6 +193,9 @@ func ChangeRule(q *reform.Querier, ruleID string, params *ChangeRuleParams) (*Ru
 	row.Severity = params.Severity
 	row.Filters = params.Filters
 	row.Params = params.RuleParams
+	if params.Summary != "" {
```
Dasio, line 196: I'm not sure about this one.
Everything else is set from request.
But If I do it here, it will be breaking change.

BupycHuk: it's interesting case. In all other endpoints we change only fields that we get from request, but doesn't change other ones. And we should keep this behaviour.
IA is still private API, so you can change behaviour.

BupycHuk (approved):
Can we create a ticket to change behaviour for other fields?

Dasio:
> Can we create a ticket to change behaviour for other fields?

I can. And also what about bool parameters, or to clear fields? I asked Fabio about it, and they are sending values each time. So with this way it is problematic to clear fields. I would suggest try to look on proto wrappers so we can distinguish nil and empty values.
 But then also, what about slice or maps? These ones should be probably required.

BupycHuk:
we add new fields `remove_*` to remove value from the fields which can be removed. API will be used not only by our front-end, but also can be used by custom scripts.

BupycHuk:
BTW please check why Travis CI is faling

AlekSi:
Please check and update/add comments in .proto files – I think they say something about required fields, how absent values are treated, etc

Dasio:
> BTW please check why Travis CI is faling

Tests are failing, but not related to this PR, issue was raised ons lack. Wrong password..

artemgavrilov (approved):

oter (approved):

PR #683, opened 2021-02-25
PMM-7339 Run individual checks

artemgavrilov:
[PMM-7339](https://jira.percona.com/browse/PMM-7339)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1567

- [x] https://github.com/percona/pmm/pull/696
---
- [x] Tests passed.
- [x] Feature build pass.
- [ ] (Re)requested review.
- [x] Update API dependency.


services/checks/checks.go
```diff
@@ -451,8 +452,36 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
+	}
+
+	em := make(map[string]struct{}, len(enable))
+	for _, e := range enable {
+		em[e] = struct{}{}
+	}
+
+	for _, c := range checks {
+		if _, ok := dm[c.Name]; ok {
+			s.l.Debugf("Skipping disabled mySQL check %s", c.Name)
```
Dasio, line 469: it's generic check, or? why is there mysql

Dasio (approved):

services/checks/checks.go
artemgavrilov: It was copy-pasted. Fixed 👍

services/checks/checks.go
```diff
@@ -451,8 +452,36 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
 	}
 }
 
-// executeChecks runs all available checks for all reachable services.
-func (s *Service) executeChecks(ctx context.Context) error {
+func (s *Service) filterChecks(checks []check.Check, disable, enable []string) []check.Check {
+	var res []check.Check
+	dm := make(map[string]struct{}, len(disable))
+	for _, e := range disable {
+		dm[e] = struct{}{}
```
oter, line 459: IMO `dm` and `em` looks confusing, maybe specify full name e.g `disabledMap`?

services/checks/registry.go
```diff
@@ -70,10 +92,12 @@ func (r *registry) getCheckResults() []sttCheckResult {
 	r.rw.RLock()
 	defer r.rw.RUnlock()
 
-	checkResults := make([]sttCheckResult, 0, len(r.checkResults))
-	checkResults = append(checkResults, r.checkResults...)
+	results := make([]sttCheckResult, 0, len(r.checkResults))
```
oter, line 95: why not using `append` or `copy`?

services/checks/registry_test.go
```diff
@@ -83,4 +83,197 @@ func TestRegistry(t *testing.T) {
...
+					"bar":       "foo",
+					"severity":  "notice",
+					"stt_check": "1",
+				},
+			},
+		}
+
+		collectedAlerts := r.collect()
+		require.Len(t, collectedAlerts, 1)
+		require.Equal(t, 1, cap(collectedAlerts))
```
oter, line 158: why checking capacity?

services/checks/registry.go
artemgavrilov: `r.checkResult` is a map

services/checks/checks.go
artemgavrilov: Renamed

services/checks/registry_test.go
artemgavrilov: It was copied from another test. Removed everywhere.

services/checks/checks.go
```diff
@@ -451,8 +452,36 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
+		disableMap[e] = struct{}{}
+	}
+
+	enableMap := make(map[string]struct{}, len(enable))
+	for _, e := range enable {
+		enableMap[e] = struct{}{}
+	}
+
+	for _, c := range checks {
+		if _, ok := disableMap[c.Name]; ok {
```
BupycHuk, line 468: should we show some error message or log if check was disabled in settings, but requested to run in API call?

BupycHuk: I think we shouldn't ignore that silently

BupycHuk (approved):
Overall looks good to me, but one thing that could be improved

services/checks/checks.go
```diff
@@ -451,8 +452,36 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
 	}
 }
 
-// executeChecks runs all available checks for all reachable services.
-func (s *Service) executeChecks(ctx context.Context) error {
+func (s *Service) filterChecks(checks []check.Check, disable, enable []string) []check.Check {
```
palash25, line 455: a comment explaining this would be nice since we will have 2 filterChecks functions after PMM-7336 is merged
something like 
``` 
// filterChecks  filters the list of checks received from the UI to be executed from the checks disabled in settings
```

oter (approved):

artemgavrilov:
Sorry, it was a hardcore merge with the parent branch, so implementation changed. So a new review required.

services/checks/checks.go
artemgavrilov: The whole function was rewritten during merge x_X

services/checks/checks.go
artemgavrilov: Well, there is only one function now, but it's not so easy to understand.

artemgavrilov:
Build failed because of https://github.com/percona/pmm-managed/pull/701#pullrequestreview-618426868

services/checks/checks.go
```diff
@@ -474,49 +477,82 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
+	}
+
+	for _, c := range checks {
+		// If empty group passed, which means `any group`
+		// or check has required interval
+		// or check has empty interval and required interval is `standard`.
+		if group == "" || c.Interval == group || (group == check.Standard && c.Interval == "") {
+			// If check enabled explicitly or all checks enabled by passing empty `enable` slice.
+			if _, ok := enableMap[c.Name]; ok || len(enableMap) == 0 {
+				// Filter disabled checks.
```
Dasio, line 503: I would move this if at begin of loop. Seems for me easier to read than 3 nested ifs.
(mean disable map)

services/checks/checks.go
```diff
@@ -474,49 +477,82 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
+	enableMap := make(map[string]struct{}, len(enable))
+	for _, e := range enable {
+		enableMap[e] = struct{}{}
+	}
+
+	for _, c := range checks {
+		// If empty group passed, which means `any group`
+		// or check has required interval
+		// or check has empty interval and required interval is `standard`.
+		if group == "" || c.Interval == group || (group == check.Standard && c.Interval == "") {
```
Dasio, line 500: Maybe also negate this condition and add continue.
"Align the happy path to the left edge"

Dasio (approved):

services/checks/checks.go
artemgavrilov: Then we will have a warning message for a check that wasn't specified. So this order covers situation "I specified check, but it's disabled"

services/checks/checks.go
artemgavrilov: Yes, I know this function in not so easy to read. I tried different approaches, even split it to different functions. If we will nagate conditions it will be harder to understand the logic, so nested `if` is a lesser evil.

palash25:
> Build failed because of #701 (review)

build should work now, SAAS-487 changes were merged, please update

services/checks/checks.go
palash25, line 500: lets rename `group` to something like `intervalBucket` otherwise this becomes confusing to read because I might start thinking that groups and intervals are different things

services/checks/checks.go
```diff
@@ -474,49 +477,82 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
...
+	postgreSQLCheckResults := s.executePostgreSQLChecks(ctx, postgreSQLChecks)
 	checkResults = append(checkResults, postgreSQLCheckResults...)
 
-	mongoDBCheckResults := s.executeMongoDBChecks(ctx, interval, disabledChecks)
+	mongoDBChecks := s.filterChecks(s.getMongoDBChecks(), group, disabledChecks, checkNames)
+	mongoDBCheckResults := s.executeMongoDBChecks(ctx, mongoDBChecks)
 	checkResults = append(checkResults, mongoDBCheckResults...)
 
-	s.alertsRegistry.set(interval, checkResults)
+	if len(checkNames) != 0 {
```
BupycHuk, line 539: Is it possible that we run the same checks for different groups?

BupycHuk (approved):

services/checks/checks.go
artemgavrilov: Nope, a single check can be a member of a single group.

services/checks/checks.go
artemgavrilov: Renamed to `intervalGroup`

PR #697, opened 2021-03-16
PMM-7345 Expose metrics for all available databases in PG server

adivinho:
[PMM-7345](https://jira.percona.com/browse/PMM-7345)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1612

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.

Screenshot_20210316_191832


services/agents/postgresql.go
```diff
@@ -47,6 +47,8 @@ func postgresExporterConfig(service *models.Service, exporter *models.Agent, red
 		"--collect.custom_query.mr.directory=/usr/local/percona/pmm2/collectors/custom-queries/postgresql/medium-resolution",
 		"--collect.custom_query.hr.directory=/usr/local/percona/pmm2/collectors/custom-queries/postgresql/high-resolution",
 
+		"--auto-discover-databases",
+		"--exclude-databases=template0,template1,postgres",
```
BupycHuk, line 51: do we really should exclude template0, template1 and postgres?

BupycHuk (changes requested):
Tests are failing

services/agents/postgresql.go
adivinho: Yes we do. Please check my comment in the ticket.

Also here are some snippets from postgresql documentation: 
_"CREATE DATABASE actually works by copying an existing database. By default, it copies the standard system database named template1. Thus that database is the "template" from which new databases are made. 
...
There is a second standard system database named template0. This database contains the same data as the initial contents of template1, that is, only the standard objects predefined by your version of PostgreSQL."_

BupycHuk: ok, then. let's fix tests (only the one related to this ticket) and go on

adivinho: Done

BupycHuk (approved):

PR #707, opened 2021-03-23
SAAS-479 Add STT and IA information to telemetry

palash25:
[SAAS-479](https://jira.percona.com/browse/SAAS-479)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1711

- [ ] Links to other linked pull requests (optional).  https://github.com/percona-platform/saas/pull/39   https://github.com/percona-platform/platform/pull/208   https://github.com/percona-platform/telemetryd/pull/161   https://github.com/percona/pmm-managed/pull/707
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


palash25:
CI seems to be failing due to unrelated reasons, have run the tests locally and they pass

palash25:
please see the telemetryd PR too, I am not able to add reviewers to it for some reason https://github.com/percona-platform/telemetryd/pull/161

oter (approved):

services/telemetry/telemetry.go
```diff
@@ -256,7 +257,7 @@ func (s *Service) sendV1Request(ctx context.Context, data []byte) error {
 	return nil
 }
 
-func (s *Service) makeV2Payload(serverUUID string) (*reporter.ReportRequest, error) {
+func (s *Service) makeV2Payload(serverUUID string, sttEnabled, iaEnabled bool) (*reporter.ReportRequest, error) {
```
Dasio, line 260: maybe wrap it up to struct?

atymchuk: @palash25 Seems like David's comment was not resolved...

palash25: i will push it before merging

artemgavrilov: @palash25 `push -> re-review -> merge`, issues should be addressed and approved by commenter

artemgavrilov: And it seems that this function can accept just `settings`

palash25: done

services/telemetry/telemetry.go
```diff
@@ -267,6 +268,8 @@ func (s *Service) makeV2Payload(serverUUID string) (*reporter.ReportRequest, err
 		Version:            s.pmmVersion,
 		UpDuration:         ptypes.DurationProto(time.Since(s.start)),
 		DistributionMethod: s.tDistributionMethod,
+		SttEnabled:         wrapperspb.Bool(settings.SaaS.STTEnabled),
```
Dasio, line 271: with this approach, it never be nil. Do we need still wrapper around it? Just asking, because I don't know

Dasio (approved):

services/telemetry/telemetry.go
palash25: this new code won't send nils, but the older PMM versions won't send the two field, so when on telemetryd we unmarshall an event received from older pmm servers which don't send these two fields, they will be filled as nil and we can perform nil checks in that case https://github.com/percona-platform/telemetryd/pull/161/files#diff-a9ef45466a6e12ccc15711a5dd88db525d8928020c2e093f5e467fc7e7ffe436R300-R308  . at least that was my idea behind it. correct me if I am wrong or mixing things up

palash25: ok so this wrapperspb has a weird behaviour, it doesn't set the value if it is `false`  `wrapperspb.Bool(false)`, only sets the value for true  https://github.com/golang/protobuf/issues/438

weird

palash25: will have to do this to be able to set false

```
if settings.SaaS.STTEnabled {
		event.SttEnabled = &wrapperspb.BoolValue{Value: true}
	} else {
		event.SttEnabled = &wrapperspb.BoolValue{Value: false}
	}

	if settings.IntegratedAlerting.Enabled {
		event.IaEnabled = &wrapperspb.BoolValue{Value: true}
	} else {
		event.IaEnabled = &wrapperspb.BoolValue{Value: false}
	}
```

then it does set false, but if I try to print event with a field set to false it still wont show `stt_enabled:{}`

i hate this wrapper

artemgavrilov: It won't help, `&wrapperspb.BoolValue{Value: false}` == `wrapperspb.Bool(false)`

Problem not with the function, but with type
```
type BoolValue struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// The bool value.
	Value bool `protobuf:"varint,1,opt,name=value,proto3" json:"value,omitempty"`
}
```

Field value has `omitempty` annotation, and for the `bool` fails is equal to emtpy.

artemgavrilov: I suppose you need some enum `ENABLED, DISABLED, UNKNOWN` instead of bool here.

services/telemetry/telemetry.go
```diff
@@ -268,6 +269,19 @@ func (s *Service) makeV2Payload(serverUUID string) (*reporter.ReportRequest, err
...
+	} else {
+		event.SttEnabled = &wrapperspb.BoolValue{Value: false}
+	}
+
+	if settings.IntegratedAlerting.Enabled {
+		event.IaEnabled = &wrapperspb.BoolValue{Value: true}
+	} else {
+		event.IaEnabled = &wrapperspb.BoolValue{Value: false}
+	}
+	s.l.Debugf("Event: %+v", event)
```
artemgavrilov, line 283: No need to do that

artemgavrilov (changes requested):

services/telemetry/telemetry.go
artemgavrilov: >  it doesn't set the value if it is false

That's ok, in protobuf bool `false` == `empty`, and that's because you used `wrapperspb`, to have a pointer to bool wrapper instead of plain `bool`.

palash25: ok I will remove this, and test it once on the dev deployment

palash25: works

services/telemetry/telemetry.go
palash25: done, remvoed

palash25:
tests had probably failed because of re-deployments on saas dev, have triggerred a travis rebuild

artemgavrilov (approved):

PR #712, opened 2021-03-30
PMM-7338 Add API to change default check execution interval

palash25:
[PMM-7338](https://jira.percona.com/browse/PMM-7338)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1644  https://github.com/Percona-Lab/pmm-api-tests/pull/211

- [ ] Links to other linked pull requests (optional). https://github.com/percona/pmm/pull/714
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


palash25:
will fix the linter warnings in the next push

services/checks/checks.go
```diff
@@ -416,6 +422,82 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+			s.cim.Lock()
+			defer s.cim.Unlock()
+			s.checksIntervalMap[c.Name] = c.Interval
+			break
+		}
+	}
+	return nil
+}
+
+func (s *Service) updateCheckInterval(newCheck check.Check) {
```
artemgavrilov, line 456: This function updates the whole check, not just the interval. So it should have a different name or different signature.

services/checks/checks.go
```diff
@@ -416,6 +422,82 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+			}
+
+			s.updateCheckInterval(c)
+
+			// since we re-run checks at regular intervals using a call
+			// to s.StartChecks which in turn calls s.collectChecks
+			// to load/download checks, we must record any changes
+			// to check intervals in this map so that they can be re-applied
+			// once the checks have been re-loaded on restarts.
+			s.cim.Lock()
```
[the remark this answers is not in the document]
palash25: thanks, will remove defer

services/checks/checks.go
palash25: renamed

services/checks/checks.go
```diff
@@ -416,6 +422,82 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+			defer s.cim.Unlock()
+			s.checksIntervalMap[c.Name] = c.Interval
```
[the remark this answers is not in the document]

services/checks/checks.go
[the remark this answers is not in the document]
palash25: removed since we are persisting to DB now

services/checks/checks.go
```diff
@@ -416,6 +417,111 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
 
+// ChangeInterval changes a check's interval to the value received from the UI.
+func (s *Service) ChangeInterval(checkName string, interval managementpb.SecurityCheckInterval) error {
+	checks := s.GetAllChecks()
+	if len(checks) == 0 {
+		return errors.New("no checks loaded")
+	}
+
+	for _, c := range checks {
+		if c.Name == checkName {
```
Dasio, line 428: I prefer negate condition and continue. So code below is not nested.

services/checks/checks.go
```diff
@@ -416,6 +417,111 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+			s.l.Error(c.Name)
```
Dasio, line 429: why log this?

Dasio (dismissed):

services/checks/checks.go
palash25: no need, I guess I forgot to remove it after debuging thanks

services/checks/checks.go
palash25: done thanks

services/management/checks.go
```diff
@@ -103,20 +105,27 @@ func (s *ChecksAPIService) ListSecurityChecks(ctx context.Context, req *manageme
...
+// ChangeSecurityChecks enables/disables Security Thread Tool checks by names or changes its execution interval.
 func (s *ChecksAPIService) ChangeSecurityChecks(ctx context.Context, req *managementpb.ChangeSecurityChecksRequest) (*managementpb.ChangeSecurityChecksResponse, error) {
 	var enableChecks, disableChecks []string
+	changeIntervalParams := make([]checks.ChangeIntervalParams, 0, len(req.Params))
 	for _, check := range req.Params {
 		if check.Enable && check.Disable {
 			return nil, status.Errorf(codes.InvalidArgument, "Check %s has enable and disable parameters set to the true.", check.Name)
 		}
 
+		changeIntervalParams = append(changeIntervalParams, checks.ChangeIntervalParams{
```
BupycHuk, line 124: We should update interval only if it's passed.

BupycHuk: let's use `map` instead of `checks.ChangeIntervalParams`, so you don't need to loop again.

models/checks_state_helper_test.go
```diff
@@ -0,0 +1,90 @@
...
+	sqlDB := testdb.Open(t, models.SkipFixtures, nil)
+	defer func() {
+		require.NoError(t, sqlDB.Close())
+	}()
+	db := reform.NewDB(sqlDB, postgresql.Dialect, reform.NewPrintfLogger(t.Logf))
+
+	t.Run("create", func(t *testing.T) {
+		tx, err := db.Begin()
+		require.NoError(t, err)
+		defer func() {
```
BupycHuk, line 41: let's use `t.Cleanup` instead of `defer` everywhere

services/checks/checks.go
```diff
@@ -419,6 +420,124 @@
...
+		// update interval change if already present
+		if cs != nil {
+			cs, err := models.ChangeCheckState(s.db.Querier, c.Name, models.Interval(c.Interval))
+			if err != nil {
+				return err
+			}
+			s.l.Debugf("Updated interval change for check: %s in DB", cs.Name)
+			break
+		}
+		return err
```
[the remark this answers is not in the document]
BupycHuk: it's critical ^

services/checks/checks.go
```diff
@@ -927,6 +1046,18 @@ func (s *Service) collectChecks(ctx context.Context) {
 	}
 
 	checks = s.filterSupportedChecks(checks)
+	for i, c := range checks {
+		cs, err := models.FindCheckStateByName(s.db.Querier, c.Name)
+		if err != nil {
+			// we use warn because this might not always be a failure, its possible
+			// that there were no interval changes to be applied in the first place.
+			s.l.Warnf("Unable to re-apply interval changes to check: %s : %+v", c.Name, err)
```
BupycHuk, line 1054: maybe then there is no reason to warn at all?

services/checks/checks.go
```diff
@@ -927,6 +1046,18 @@ func (s *Service) collectChecks(ctx context.Context) {
 	}
 
 	checks = s.filterSupportedChecks(checks)
+	for i, c := range checks {
+		cs, err := models.FindCheckStateByName(s.db.Querier, c.Name)
```
BupycHuk, line 1050: Wouldn't it better to just get all `CheckStates` and convert to map? to not have n queries to DB.

services/checks/checks.go
```diff
@@ -419,6 +420,124 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+	switch newCheck.Type {
+	case check.MySQLSelect:
+		fallthrough
+	case check.MySQLShow:
+		s.cm.Lock()
+		defer s.cm.Unlock()
+		for i, oldCheck := range s.mySQLChecks {
+			if oldCheck.Name == newCheck.Name {
+				s.mySQLChecks[i] = newCheck
+				s.l.Infof("Changed check interval for: %s, from: %s to: %s", oldCheck.Name, oldCheck.Interval, newCheck.Interval)
```
BupycHuk, line 503: actually you don't change interval here, you replace whole check.

services/checks/checks.go
```diff
@@ -419,6 +420,124 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+	}
+
+	return nil
+}
+
+// updateCheck updates a check with an updated interval in the appropriate check slice.
+func (s *Service) updateCheck(newCheck check.Check) {
+	switch newCheck.Type {
+	case check.MySQLSelect:
+		fallthrough
```
BupycHuk, line 496: interesting case about `fallthrough` platform team uses it, but Core and Server integration teams uses comma.
https://github.com/percona/pmm-managed/blob/1b79357f4ef7037243e532c07d3cb8f6ceee7ee4/services/victoriametrics/prometheus.go#L146

BupycHuk (changes requested):

BupycHuk:
Tests are failing

services/checks/checks.go
palash25: updated the message

services/checks/checks.go
palash25: done

services/checks/checks.go
palash25: removed

models/checks_state_helper_test.go
palash25: done

services/management/checks.go
palash25: done

services/checks/checks.go
palash25: forgot to remove this, will do in a few mins

palash25: actually I thought better than removing it, is to log it if we receive an error other than `reform.ErrNoRows` from the DB

let me know what you think

services/checks/checks.go
BupycHuk: it''s still here.

palash25: missed one out of three messages, sorry. updated

services/checks/checks.go
```diff
@@ -419,6 +420,118 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+		return errors.New("no checks loaded")
+	}
+
+	checkMap := make(map[string]check.Check, len(checks))
+	for _, ch := range checks {
+		checkMap[ch.Name] = ch
+	}
+
+	checkStateMap, err := models.FindCheckStates(s.db.Querier)
+	if err != reform.ErrNoRows && err != nil {
```
artemgavrilov, line 436:
```suggestion
	if err != nil && err != reform.ErrNoRows {
```

services/checks/checks.go
```diff
@@ -419,6 +420,118 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+		// to check intervals in the DB so that they can be re-applied
+		// once the checks have been re-loaded on restarts.
+		if _, ok := checkStateMap[c.Name]; !ok {
+			cs, err := models.CreateCheckState(s.db.Querier, c.Name, models.Interval(c.Interval))
+			if err != nil {
+				return err
+			}
+			s.l.Debugf("Saved interval change for check: %s in DB", cs.Name)
+			continue
+		}
```
artemgavrilov, line 473: It can fail because you grab existing without transaction, so some other user can create a check state. So I think all DB interaction should be wrapped into the transaction.

services/checks/checks.go
```diff
@@ -419,6 +420,118 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+			c.Interval = check.Frequent
+		case managementpb.SecurityCheckInterval_RARE:
+			c.Interval = check.Rare
+		case managementpb.SecurityCheckInterval_SECURITY_CHECK_INTERVAL_INVALID:
+			return errors.New("invalid security check interval")
+		default:
+			return errors.New("unknown security check interval")
+		}
+
+		s.updateCheck(c)
```
artemgavrilov, line 459: In memory state should be updated after everything persisted in DB, otherwise, in case of DB error, you will have inconsistent state.

services/checks/checks.go
```diff
@@ -419,6 +420,118 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+	if len(checks) == 0 {
+		return errors.New("no checks loaded")
+	}
+
+	checkMap := make(map[string]check.Check, len(checks))
+	for _, ch := range checks {
+		checkMap[ch.Name] = ch
+	}
+
+	checkStateMap, err := models.FindCheckStates(s.db.Querier)
```
artemgavrilov, line 435: Not sure, bet maybe it makes sense to not load all states. We know check names that we want to update.

palash25: agreed it was done after this review comment https://github.com/percona/pmm-managed/pull/712#discussion_r606407886 

right now we just change interval for a single check so we don't need to load all but in case this feature is turned into a bulk action on the UI in the future, we won't have to change the backend and it will still work

BupycHuk: Yeah, but it was requested in another place, not here.

palash25: you are right I made the change in the wrong place. will revert this and make that change in collect checks

models/checks_state_model.go
```diff
@@ -0,0 +1,36 @@
...
+// Available check execution intervals.
+const (
+	Standard Interval = "standard"
+	Frequent Interval = "frequent"
+	Rare     Interval = "rare"
+)
+
+// ChecksState represents any changes to an STT check loaded in pmm-managed.
+//reform:checks_state
+type ChecksState struct {
```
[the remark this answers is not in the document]
palash25: no I prefer checksstate over setting and I can't think of anything better

services/checks/checks.go
palash25: done, now it loads all the checkstates in `collectChecks` and `ChangeInterval` is back to the way it was retrieving one state at a time using name

services/checks/checks.go

services/checks/checks.go

models/checks_state_helper.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		return cs, nil
+	case reform.ErrNoRows:
+		return nil, err
+	default:
+		return nil, errors.WithStack(err)
+	}
+}
+
+// FindCheckStateByName finds ChecksState by name.
```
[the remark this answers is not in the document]
palash25: done

models/checks_state_model.go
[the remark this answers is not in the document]
artemgavrilov: But it is not a state, this is overwritten parameter/configuration. I mean current name doesn't reflect the content.

services/checks/checks.go
```diff
@@ -419,6 +420,124 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+
+// updateCheck updates a check with an updated interval in the appropriate check slice.
+func (s *Service) updateCheck(newCheck check.Check) {
+	switch newCheck.Type {
+	case check.MySQLSelect:
+		fallthrough
+	case check.MySQLShow:
+		s.cm.Lock()
+		defer s.cm.Unlock()
+		for i, oldCheck := range s.mySQLChecks {
```
artemgavrilov, line 500: How about making `s.mySQLChecks` map instead of slice? It will allow us to eliminate that loop.

Same for other checks slices.

services/checks/checks.go
```diff
@@ -927,6 +1049,22 @@ func (s *Service) collectChecks(ctx context.Context) {
...
+		return
+	}
+
+	for i, c := range checks {
+		interval, ok := checkStateMap[c.Name]
+		if !ok {
+			continue
+		}
+		c.Interval = check.Interval(interval)
+		checks[i] = c
```
artemgavrilov, line 1065: Minor:
```suggestion
		if ok {
			c.Interval = check.Interval(interval)
			checks[i] = c
		}
```

services/management/checks.go
```diff
@@ -126,6 +133,17 @@ func (s *ChecksAPIService) ChangeSecurityChecks(ctx context.Context, req *manage
...
 
+	if len(changeIntervalParams) != 0 {
+		err := s.checksService.ChangeInterval(changeIntervalParams)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to change security check interval")
+		}
+		// enabling/disabling checks and changing interval are two seperate
+		// features on the UI that shrare the same API endpoint.
+		// if check interval was changed without error then return early.
+		return &managementpb.ChangeSecurityChecksResponse{}, nil
```
artemgavrilov, line 144: But why? Why user can enable check and set interval simultaneously?

services/checks/checks.go
artemgavrilov: Or even: 
```
		if interval, ok := checkStateMap[c.Name]; ok {
		...
		}
```

services/management/checks.go
palash25: i explained it in the comment, according to the UI design these were two different things
but on second thoughts maybe it will be better to remove that so that we don't have to change the backend in case the UI allows users to do those two things together in the future

services/checks/checks.go
```diff
@@ -419,6 +420,124 @@ func (s *Service) EnableChecks(checkNames []string) error {
 	return nil
 }
 
+// ChangeInterval changes a check's interval to the value received from the UI.
+func (s *Service) ChangeInterval(params map[string]managementpb.SecurityCheckInterval) error {
```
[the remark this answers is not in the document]

services/checks/checks.go
palash25: thats a good idea, removed a lot of unnecessary code thanks

services/checks/checks.go

services/management/checks.go
palash25: done

services/checks/checks.go
```diff
@@ -419,6 +433,98 @@ func (s *Service) EnableChecks(checkNames []string) error {
...
+				s.l.Debugf("Updated interval change for check: %s in DB", cs.Name)
+				return nil
+			}
+			return err
+		})
+		if e != nil {
+			return e
+		}
+
+		s.updateCheck(c)
```
BupycHuk, line 482: I suggest to include it into transaction

BupycHuk (approved):

models/checks_state_model.go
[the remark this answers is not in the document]
palash25: forgot to rename this, please continue with the review, I will rename this shortly

services/checks/checks.go

models/checks_state_model.go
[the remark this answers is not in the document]
palash25: @artemgavrilov renamed

BupycHuk (dismissed):
let's merge as is, we will cleanup later.

models/check_settings_helper_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/reform.v1"
+	"gopkg.in/reform.v1/dialects/postgresql"
+
+	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/utils/testdb"
+)
+
+func TestChecksState(t *testing.T) {
```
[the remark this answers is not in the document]
palash25: done

artemgavrilov (dismissed):

BupycHuk (approved):

BupycHuk (approved):

artemgavrilov (approved):

PR #726, opened 2021-04-13
Refactoring

artemgavrilov:
[PMM-XXXX](https://jira.percona.com/browse/PMM-XXXX)

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/location_helpers.go
```diff
@@ -239,11 +234,11 @@ func (c BackupLocationConfig) Validate() (bool, error) {
 		err = checkPMMClientLocationConfig(c.PMMClientConfig)
 	}
 
-	if configCount > 1 {
-		return false, status.Error(codes.InvalidArgument, "Only one config is allowed.")
+	if configCount != 1 {
```
artemgavrilov, line 237: API description requires exactly one config to be present.

services/management/backup/locations_service.go
```diff
@@ -97,6 +101,19 @@ func (s *LocationsService) AddLocation(ctx context.Context, req *backupv1beta1.A
...
+	if params.S3Config != nil {
+		bucketLocation, err := getBucketLocation(params.S3Config)
+		if err != nil {
+			return nil, err
+		}
+
+		params.S3Config.BucketLocation = bucketLocation
+	}
+
 	loc, err := models.CreateBackupLocation(s.db.Querier, params)
```
artemgavrilov, line 117: Main idea

models/location_helpers.go
Dasio: But when u want change location, you don't have to specify config.

artemgavrilov: I don't get, you mean when user wants to change location name or description?

API should be `send only fields that should be changed` or `send whole object`, but it was a mix of these two approaches. So user should have sent `name` and `description` each time, but configs only when he wants to change it.

Dasio: No, user can send empty `name` or `description`, then it's not updated.

And with this change, if user don't want change config, so he don't send anything. It will fail because of `configCount == 0`

artemgavrilov: Hmm, it differs from how we handle IA entities. So as I see location `description` field is optional when you create record, but once you set description you are not able to remove it.

Dasio: Yes, we can try use wrappers for this, when it's new API. So we can check empty vs not sent.

artemgavrilov: I slightly remember that this was discussed somewhere. It seems that different API parts have different behavior, which is not good.

Well, let's discuss it at some meeting. Please ignore this for now.

services/management/backup/locations_service.go
```diff
@@ -215,16 +257,46 @@ func convertLocation(location *models.BackupLocation) (*backupv1beta1.Location,
...
+}
+
+func (s *LocationsService) checkBucket(c *models.S3LocationConfig) error {
+	url, err := models.ParseEndpoint(c.Endpoint)
+	if err != nil {
+		return status.Errorf(codes.InvalidArgument, "%s", err)
+	}
+
+	secure := true
+	if url.Scheme == "https" {
```
oter, line 290:
```suggestion
	if url.Scheme == "http" {
```

services/management/backup/locations_service.go
```diff
@@ -215,16 +257,46 @@ func convertLocation(location *models.BackupLocation) (*backupv1beta1.Location,
...
+		secure = false
+	}
+
+	exists, err := s.s3.BucketExists(url.Host, secure, c.AccessKey, c.SecretKey, c.BucketName)
+	if err != nil {
+		return err
```
oter, line 296: The user will see `Internal error` if credentials not valid. And, it looks like, without any descriptive message from s3 provider.

models/location_helpers.go
```diff
@@ -81,7 +80,28 @@ func checkPMMClientLocationConfig(c *PMMClientLocationConfig) error {
...
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "%s", err)
+		}
+	}
+
+	return parsedURL, nil
+}
+
+// checkS3Config checks S3 config and returns the flag that indicates if
+// secure connection should be used and the parsed host.
```
artemgavrilov, line 103:
```suggestion
// checkS3Config checks S3 config.
```

PR #730, opened 2021-04-16
PMM-7938 migrate to go modules

Dasio:
[PMM-7938](https://jira.percona.com/browse/PMM-7938)

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


.devcontainer/setup.py
```diff
@@ -71,7 +71,6 @@ def install_tools():
     """Installs Go developer tools."""
 
     run_commands([
-        "curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh",
         "curl https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /root/go/bin",
         "curl https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh | sh -s -- -b /root/go/bin",
```
artemgavrilov, line 75: It seems that this also can be replaced with `go install -modfile=tools/go.mod ...`

go.mod
```diff
@@ -0,0 +1,51 @@
+module github.com/percona/pmm-managed
+
+go 1.16
+
```
artemgavrilov, line 4: Add please comments similar to https://github.com/percona-platform/template/blob/4b2f2acce3592549eb136486c6720ae71022e0a9/go.mod#L5-L9

for `saas` and `pmm` dependencies, please

services/agents/channel/channel_test.go
```diff
@@ -230,7 +234,7 @@ func TestServerExitsWithUnknownErrorIntercepted(t *testing.T) {
 	assert.NoError(t, err)
 
 	_, err = stream.Recv()
-	assert.Equal(t, status.Error(codes.Internal, "Internal server error."), err)
+	assert.Equal(t, status.Error(codes.Internal, "Internal server error.").Error(), err.Error())
```
artemgavrilov, line 237: tests.AssertGRPCError()

tools/tools.go
```diff
@@ -0,0 +1,32 @@
...
+// +build tools
+
+package tools
+
+import (
+	_ "github.com/kevinburke/go-bindata/go-bindata"
+	_ "github.com/vektra/mockery/cmd/mockery"
+	_ "golang.org/x/tools/cmd/goimports"
+	_ "gopkg.in/reform.v1/reform"
+
```
artemgavrilov, line 27: Why empty line?

go.mod
Dasio: We can't use replace directive because pmm is not migrated to go modules

.devcontainer/setup.py
Dasio: Should I merge it? And add 
```
 github.com/go-delve/delve/cmd/dlv@latest \
                golang.org/x/tools/gopls@latest \
                mvdan.cc/gofumpt@latest \
                mvdan.cc/gofumpt/gofumports"
```
also to tools ?

artemgavrilov: Better to install everything that possible with `tools`, there we can easily control versions.

artemgavrilov: We can remove `run_commands` and install all that we need with `make init` as we did it in saas repos.

https://github.com/percona-platform/authed/blob/7c0fcc523941a56d8d337f21c5e8d0175864805b/Makefile#L11-L16

go.mod
artemgavrilov: Ok, then only for `saas` for now

.devcontainer/setup.py
Dasio: CI failed:
`/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exec: "gcc": executable file not found in $PATH`

go.mod
```diff
@@ -0,0 +1,59 @@
...
+// Use for local development, but do not commit:
+// replace github.com/percona-platform/platform => /local-path/platform
```
artemgavrilov, line 6:
```suggestion
// replace github.com/percona-platform/saas => ../saas
```
I assume it's a common case when all percona repos located in the same folder. So better to have `../saas` as replace target. In 95% cases developer should just uncomment this line

artemgavrilov (approved):

.devcontainer/setup.py
artemgavrilov: The problem already solved as I see.

go.mod
Dasio: but saas is in percona platform, so maybe then `../../saas`

artemgavrilov: My modules projects located outside gopath, just in folder `~/projects/{pmm-managed,saas,pmm-agent}`.
I thought it's a more or less common approach. Now modules are enabled by default, so all modules projects work in Gopath without setting `GO111MODULES` variable, so I'm not sure what option is better. But at least this `replace` pattern already used in platform repos.

Makefile.devcontainer
```diff
@@ -52,14 +71,12 @@ fuzz-grafana:                   ## Run fuzzer for services/grafana package.
 
 check:                          ## Run required checkers and linters.
 	go run .github/check-license.go
-	dep check
-	go-sumtype ./vendor/... ./...
```
BupycHuk, line 56: we should still keep this check.

BupycHuk (commented):
let's add a comment about `github.com/percona-platform/dbaas-api`

docker-compose.yml
```diff
@@ -34,3 +34,6 @@ services:
     volumes:
       - .:/root/go/src/github.com/percona/pmm-managed
       - ./Makefile.devcontainer:/root/go/src/github.com/percona/pmm-managed/Makefile:ro
+      - go-modules:/root/go/pkg/mod # Put modules cache into a separate volume
+volumes:
+  go-modules:
```
oter, line 39: let's add new line at the end.

Makefile.devcontainer
Dasio: how? we removed vendor folder.
We removed this check from previos repositories

BupycHuk: there are 2 folders :)

Dasio: Do we use it somewhere? 
Maybe I don't get usage, but based on README from go-sumtype
Comment is needed
```
//go-sumtype:decl MySumType
```
And I can't see it anywhere in our codebase

askomorokhov (approved):

Makefile.devcontainer
palash25: so I checked and we need it, those comments are only there in pmm repo because the types for which go-sumtype is used for are in the pmm repo, but these types are used in switch cases for example

`inventorypb.Node` type in pmm repo has the directive comment  https://github.com/percona/pmm/blob/b23824e4facf723fdbb08b0ea28f30c9f51e386e/api/inventorypb/nodes.go#L3

and the switch case for that is in pmm-managed here https://github.com/percona/pmm-managed/blob/PMM-2.0/services/management/management.go#L100

so go-sumtype is needed in this repo to check for exhaustivness of such types and there are many such types, so we should keep it

jprukner (approved):

JiriCtvrtka (approved):

Makefile.devcontainer
palash25: sorry the second link was wrong, here is the usage of the NodeTypes in switch statements  https://github.com/percona/pmm-managed/blob/PMM-2.0/services/inventory/grpc/nodes_server.go#L67-L78

if you remove any one case from it then go-sumtype would fail  

```
/root/go/src/github.com/percona/pmm-managed/services/inventory/grpc/nodes_server.go:66:3: exhaustiveness check failed for sum type 'Node': missing cases for RemoteAzureDatabaseNode
make[1]: *** [check] Error 1
```

so it works and we should keep it

palash25: now that we are removing vendor we should change the path for the dependency we give in the go-sumtype command `go-sumtype ./...  <path-to-dependencies>`

BupycHuk (approved):

oter (approved):

Makefile.devcontainer
Dasio: Is it needed to check depedencies? If yes, how to do it. It isn't straightforward

palash25: yes the dependency (or the pmm repo) has the go-sumtype directive so we should supply the path to that to actually run `go-sumtype`

but that might be difficult without a vendor folder
i found this command of off the internet ` go list -m -f '{{if not (or .Indirect .Main)}}{{.Path}}{{end}}' all` which lists all modules used by a go modules project but it only lists them as import paths and not the actual paths

i guess we can skip it for now, let the command be as it is and merge this. 

I will try to find a solution for this. If I can't then we can remove it in the future. Lets merge it for now

palash25: there is one more solution for this, that is having the `vendor` folder

palash25 (approved):

Makefile.devcontainer
Dasio: I don't want keep vendor folder just because of  linter.

BupycHuk: If there is no other way, let's go forward without this linter.

artemgavrilov: `A simple utility for running exhaustiveness checks on type switch statements.`

As I understand this linter checks that all options specified in switch statement. So in dependencies we have types that "prepared" for that, but in this repo we have switch statements. So we don't need vendor or something else, this linter will work for that repo.

PR #746, opened 2021-04-30
PMM-7567 MongoDB backup

artemgavrilov:
[PMM-7567](https://jira.percona.com/browse/PMM-7567)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1727
- [x] https://github.com/percona/pmm/pull/726
- [ ] https://github.com/percona/pmm-agent/pull/256
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


Gopkg.toml
```diff
@@ -14,7 +14,7 @@ required = [
 
 [[constraint]]
   name = "github.com/percona/pmm"
-  branch = "PMM-2.0"
+  branch = "PMM-7567-mongo-backup"
```
artemgavrilov, line 17: TODO

models/artifact_helpers.go
```diff
@@ -137,6 +155,10 @@ func CreateArtifact(q *reform.Querier, params CreateArtifactParams) (*Artifact,
 		return nil, errors.WithStack(err)
 	}
 
+	if err := checkUniqueArtifactName(q, params.Name); err != nil {
+		return nil, err
+	}
```
artemgavrilov, line 160: Added constraint for artifact name uniqueness. It should be also DB constraint, but we already released schema without it. Probably we can cleanup artifacts table in migration and apply constraint until backups are in tech preview state.

models/agent_model.go
```diff
@@ -220,7 +220,7 @@ type DBConfig struct {
 
 // Valid returns true if config is valid.
 func (c DBConfig) Valid() bool {
-	return c.User != "" && (c.Address != "" || c.Socket != "")
+	return c.Address != "" || c.Socket != ""
```
Dasio, line 223: can be user unspecified?

Dasio (approved):

models/artifact_helpers.go
BupycHuk: Why artifact name should be unique? don't we add datetime to artifact name?

BupycHuk (approved):

models/agent_model.go
artemgavrilov: Yes, at least for `mongoDB`. For MySQL not sure, need to check.

models/artifact_helpers.go
artemgavrilov: That restriction was suggested by @rnovikovP to simplify MVP. For now we use name exactly how it was specified by user, but it seems that we need really unique names for backups, because backup tooling will use name to determine what to restore. So we should be more or less sure that there is no second backup with same name.

P.s. backup name is the `folder` name in S3, so two backups with the same name will just put their files in the same directory, but it can cause problems.

models/agent_model.go
artemgavrilov: @Dasio regarding [xtrabackup](https://www.percona.com/doc/percona-xtrabackup/8.0/xtrabackup_bin/xbk_option_reference.html) and [mysql](https://linux.die.net/man/1/mysql) documentation user is optional, in case when it's not specified defaults used (root/emtpy pass)

artemgavrilov: I tested it. Yes, user can be empty also for MySQL. So I made changes in MySQL backup job In pmm agent.

PR #748, opened 2021-05-05
PMM-8013: Fix artifacts and history items listing

oter:
[PMM-8013](https://jira.percona.com/browse/PMM-8013)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1804

---
- [x] Tests passed.
- [x] Feature build pass.
- [x] (Re)requested review.


models/artifact_helpers_test.go
```diff
@@ -29,6 +30,8 @@ import (
 	"github.com/percona/pmm-managed/utils/testdb"
 )
 
+// TODO: Add a test for the case when service removed, artifacts that belong to the service should be removed too
```
artemgavrilov, line 33: I'm not sure that we should remove such artifacts.

models/service_helpers.go
```diff
@@ -262,19 +262,40 @@ func RemoveService(q *reform.Querier, id string, mode RemoveMode) error {
...
+			for _, a := range agents {
+				if _, err := RemoveAgent(q, a.AgentID, RemoveCascade); err != nil {
+					return err
+				}
+			}
+			for _, a := range artifacts {
+				if err := RemoveArtifact(q, a.ID); err != nil {
+					return err
+				}
+			}
```
artemgavrilov, line 296: Not sure that it's a good idea. Maybe we can just set empty `service_id` for such artifacts?

Dasio: I think I agree with Artem because of this use case.
1. User will backup service1
2. User deletes service1
3. User want restore backup from step(1) to new service.

models/artifact_helpers_test.go
```diff
@@ -158,18 +214,6 @@ func TestArtifactValidation(t *testing.T) {
 		params   models.CreateArtifactParams
 		errorMsg string
 	}{
-		{
```
BupycHuk, line 161: Why this test is deleted?

oter: this is a duplicate of https://github.com/percona/pmm-managed/pull/748/files#diff-5a898b5b01f7ff28425211fc14a0224a0c34ecb231119a14a8066316aa836767R93

models/service_helpers.go
oter: reworked, so now I'm returning empty ServiceID and ServiceName and keeping the artifacts for removed services

models/database.go
```diff
@@ -508,6 +508,10 @@ var databaseSchema = [][]string{
 	37: {
 		`ALTER TABLE agents ALTER COLUMN max_query_log_size TYPE BIGINT`,
 	},
+	38: {
+		`ALTER TABLE artifacts ADD FOREIGN KEY (location_id) REFERENCES backup_locations (id)`,
+		`ALTER TABLE artifacts DROP CONSTRAINT artifacts_service_id_check`,
```
artemgavrilov, line 513: I don't see such constraint in DB migrations. What is it?

models/database.go
```diff
@@ -508,6 +508,10 @@ var databaseSchema = [][]string{
 	37: {
 		`ALTER TABLE agents ALTER COLUMN max_query_log_size TYPE BIGINT`,
 	},
+	38: {
+		`ALTER TABLE artifacts ADD FOREIGN KEY (location_id) REFERENCES backup_locations (id)`,
```
artemgavrilov, line 512: How we will handle cases when the user already has some data in artifacts table and it does not satisfy this constraint?

BupycHuk: let's remove existed artifacts if they not satisfy this constraint

oter: done

models/location_helpers.go
```diff
@@ -358,6 +358,21 @@ func RemoveBackupLocation(q *reform.Querier, id string, mode RemoveMode) error {
...
+		return status.Errorf(codes.FailedPrecondition, "backup location with ID %q has artifacts.", id)
+	}
+
+	for _, a := range artifacts {
+		if err := RemoveArtifact(q, a.ID); err != nil {
+			return err
+		}
+	}
+
 	// @TODO - force delete https://jira.percona.com/browse/PMM-7475
```
Dasio, line 376: you basically did this ticket, no?

models/location_helpers.go
```diff
@@ -358,6 +358,21 @@ func RemoveBackupLocation(q *reform.Querier, id string, mode RemoveMode) error {
...
+	artifacts, err := FindArtifacts(q, &ArtifactFilters{LocationID: id})
+	if err != nil {
+		return err
+	}
+
+	if len(artifacts) != 0 && mode == RemoveRestrict {
+		return status.Errorf(codes.FailedPrecondition, "backup location with ID %q has artifacts.", id)
+	}
+
+	for _, a := range artifacts {
```
Dasio, line 370: can't we remove all artifacts at once?

models/location_helpers.go
oter: well looks like yes :) it is left to handle `force` flag

artemgavrilov (approved):

models/artifact_helpers.go
```diff
@@ -32,9 +32,35 @@ var (
...
+		if _, err := FindServiceByID(q, filters.ServiceID); err != nil {
+			return nil, err
+		}
+		conditions = append(conditions, fmt.Sprintf("service_id = %s", q.Placeholder(1)))
+		args = append(args, filters.ServiceID)
+	}
+	if filters != nil && filters.LocationID != "" {
+		conditions = append(conditions, fmt.Sprintf("location_id = %s", q.Placeholder(1)))
+		args = append(args, filters.LocationID)
+	}
```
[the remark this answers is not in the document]
oter: should be vice versed, thanks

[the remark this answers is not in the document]
oter: fixed

models/location_helpers.go
oter: done

models/location_helpers.go
oter: We could, but I cannot find this approach in our codebase, we always remove item-by-item...

PR #757, opened 2021-05-13
DLV Debugger

Dasio:
Dev command for attaching delve
Compile pmm-managed with extra debug flags

I'm not sure about `make install`:
It does `go install` with debugging flags
But then make release
Which creates binaries to PMM_RELEASE_PATH. But `make run` copy binary from this PMM_RELEASE_PATH, copy to /sbin and run it.
So I think, we don't use then binary with these extra flags

### DIFF
--- a/Makefile.devcontainer
+++ b/Makefile.devcontainer
@@ -45,6 +45,11 @@ install-race:                   ## Install pmm-managed binaries with race detect
 	go build -race -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-init ./cmd/pmm-managed-init
 	go build -race -v $(PMM_LD_FLAGS) -o $(PMM_RELEASE_PATH)/pmm-managed-starlark ./cmd/pmm-managed-starlark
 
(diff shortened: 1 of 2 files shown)


BupycHuk (approved):

Makefile.devcontainer
```diff
@@ -117,3 +124,6 @@ travis: ci-reviewdog test-race test-cover test-crosscover
 ci-reviewdog:
 	golangci-lint run -c=.golangci-required.yml --out-format=line-number | reviewdog -f=golangci-lint -level=error -reporter=github-pr-check
 	golangci-lint run -c=.golangci.yml --out-format=line-number | reviewdog -f=golangci-lint -level=error -reporter=github-pr-review
+
+dlv/attach:
+	dlv --listen=:2345 --headless=true --api-version=2 --accept-multiclient attach $(shell pgrep pmm-managed)
```
askomorokhov, line 129: Please add a line at EOF

askomorokhov (approved):

BupycHuk:
oops

PR #773, opened 2021-06-02
PMM-8191 Scheduling service

Dasio:
[PMM-8191](https://jira.percona.com/browse/PMM-8191)

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


Dasio:
I'll write later also tests. But to speed up review process, I converted it to PR.

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,218 @@
...
+			if err == nil || err == context.Canceled {
+				break
+			} else {
+				l.Error(err)
+			}
+			if retry <= 0 {
+				break
+			}
+			retry--
+			time.Sleep(retryInterval)
```
artemgavrilov, line 176: Better to use `select` with timer and ctx.Done()

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/percona/pmm-managed/models"
+
+	"github.com/go-co-op/gocron"
+	"github.com/sirupsen/logrus"
+	"gopkg.in/reform.v1"
+)
+
+type jobsDeps struct {
+}
```
artemgavrilov, line 17: That's where we will keep dependencies on the other services?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,218 @@
...
+			l.WithField("duration", time.Since(t)).Debug("Ended  job")
+			if err == nil || err == context.Canceled {
+				break
+			} else {
+				l.Error(err)
+			}
+			if retry <= 0 {
+				break
+			}
+			retry--
```
artemgavrilov, line 175: I wonder, do we need to track retries id DB? Because now we will lose retries counter in case of a server restart.

services/scheduler/scheduler.go
Dasio: Yeah, I didn't came up with better solution. So eg. pass agentsService to scheduler and then to this jobs.

services/scheduler/scheduler.go
Dasio: Good question. Now I hold retries just for new schedule job. So, add also current retry attemp?

models/schedule_job_models.go
```diff
@@ -0,0 +1,100 @@
...
+	Type           ScheduleJobType  `reform:"type"`
+	Data           *ScheduleJobData `reform:"data"`
+	Retries        uint             `reform:"retries"`
+	RetryInterval  time.Duration    `reform:"retry_interval"`
+	CreatedAt      time.Time        `reform:"created_at"`
+	UpdatedAt      time.Time        `reform:"updated_at"`
+}
+
+// ScheduleJobData holds result data for different job types.
+type ScheduleJobData struct {
```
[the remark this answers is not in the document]
Dasio: so let's rename also `ScheduleJob` to `ScheduledTask` ?

models/schedule_job_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+	default:
+		return nil, errors.WithStack(err)
+	}
+}
+
+type ScheduleJobsFilter struct {
+	Disabled *bool
+}
+
+// FindScheduleJobs returns all scheduled satisfying filter.
```
oter, line 51:
```suggestion
// FindScheduleJobs returns all scheduled jobs satisfying filter.
```

services/scheduler/scheduler.go
oter: yes, I think we need current retry attempt in the DB.

models/schedule_job_models.go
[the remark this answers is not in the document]
artemgavrilov: Yep, I think so. Just to not get confused if the future.

services/scheduler/scheduler.go
artemgavrilov: Not sure, but maybe it will be more explicit if job/task will accept required deps in constructor

```
func NewEchoJob(message string, someSrv service.Service, etc)
```

And when we retore this task from DB schedule service will invoke this constructor again to rebuild task.

services/scheduler/scheduler.go
artemgavrilov: But does it make sense to continue retries when the server was shut down for 2 days, for example?

services/scheduler/job.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Job interface {
+	Do(ctx context.Context) error
+	Type() models.ScheduleJobType
+	Data() models.ScheduleJobData
+}
+
+type EchoJob struct {
+	jobsDeps
+	models.EchoJobData
```
artemgavrilov, line 18: Can we have simpe `message string` here and convert `EchoJob` to the model right before persisting? IMHO it will be more explicit and simpler.

services/scheduler/scheduler.go
Dasio: I can do this. But still, I'll need to pass every service to scheduler.

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,316 @@
...
+		if err != nil {
+			return err
+		}
+
+		scheduledTask, err = models.ChangeScheduledTask(tx.Querier, id, models.ChangeScheduledTaskParams{
+			NextRun: scheduleJob.NextRun(),
+			LastRun: scheduleJob.LastRun(),
+		})
+		if err != nil {
+			s.l.WithField("id", id).Errorf("failed to set next run for new created task")
```
oter, line 89: can we return an error here and discard the scheduled job?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,316 @@
...
+			_, err = models.ChangeScheduledTask(s.db.Querier, id, models.ChangeScheduledTaskParams{
+				Running: pointer.ToBool(true),
+			})
+
+			err = task.Do(ctx)
+			l.WithField("duration", time.Since(t)).Debug("Ended task")
+			if err == nil {
+				succeeded = true
+				break
+			} else {
```
oter, line 235: could omit else stmt since you are doing break.

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,316 @@
...
+		retriesRemaining := retry
+		succeeded := false
+		for {
+			t := time.Now()
+			l.Debug("Starting task")
+			_, err = models.ChangeScheduledTask(s.db.Querier, id, models.ChangeScheduledTaskParams{
+				Running: pointer.ToBool(true),
+			})
+
+			err = task.Do(ctx)
```
oter, line 230: as I see in the case of `retriesRemaining` == n you will perform n+1 actual attempts 😄

oter: let's also cover this case in the test

Dasio: job should run at least one time.
If retries == 0,  it will run just one time.
If retries == 1, it will run max two times. One run, and one retry.

Also I don't know how to test this. Any idea? I can schedule running job with minimum frequency 1min. And then this job would be long.

oter: I got the idea, then it's ok. 

Regarding the testing, you could set `retryInterval` to 1ms and have a mock for Do() function, which will return an error couple of times and a successful last call. On the mock's side expect all these calls. Also, you could check `retriesRemaining` in the DB.

Dasio: But still. next run can take up to 60s, before `Do()` function is run.

services/scheduler/scheduler.go
Dasio: do you think it's better to discard it, just because of next run wasn't saved to DB?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+		l.WithField("duration", time.Since(t)).Debug("Ended task")
+
+		_, err = models.ChangeScheduledTask(s.db.Querier, id, models.ChangeScheduledTaskParams{
+			Running: pointer.ToBool(false),
+		})
+
+		if err != nil {
+			l.Errorf("failed to change running status: %v", err)
+		}
```
BupycHuk, line 268: it makes no sense to change scheduled task here because we anyway update it in taskFinished.

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+	l := s.l.WithField("id", id)
+
+	dbTask, err := models.FindScheduledTaskByID(s.db.Querier, id)
+	if err != nil {
+		return
+	}
+
+	params := models.ChangeScheduledTaskParams{
+		Succeeded: pointer.ToUint(dbTask.Succeeded),
+		Failed:    pointer.ToUint(dbTask.Failed),
```
BupycHuk, line 291: it's unnecessary here, if it's nil you won't change it anyway

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+		Running:   pointer.ToBool(false),
+	}
+
+	if taskErr == nil {
+		params.Succeeded = pointer.ToUint(dbTask.Succeeded + 1)
+		params.Error = pointer.ToString("")
+	} else {
+		params.Failed = pointer.ToUint(dbTask.Failed + 1)
+		params.Error = pointer.ToString(taskErr.Error())
```
BupycHuk, line 300: don't we store it in `job` or `artifact`?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+func (s *Service) taskFinished(id string, taskErr error) {
+	var job *gocron.Job
+	for _, j := range s.scheduler.Jobs() {
+		if len(j.Tags()) > 0 && j.Tags()[0] == id {
+			job = j
+			break
+		}
+	}
```
BupycHuk, line 281: let's use some map to store taskID -> Job link

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Reload removes job from scheduler and add it again from DB.
+func (s *Service) Reload(id string) error {
```
BupycHuk, line 152: Why do we need it?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+		l.Debug("Starting task")
+		_, err = models.ChangeScheduledTask(s.db.Querier, id, models.ChangeScheduledTaskParams{
+			Running: pointer.ToBool(true),
+		})
+
+		if err != nil {
+			l.Errorf("failed to change running state: %v", err)
+		}
+
+		taskErr := task.Run(ctx)
```
BupycHuk, line 256: task might be outdated(changed) to this moment, so it's better to reload it first.

BupycHuk: my suggestion to convert DBTask into Task in `wrapTask` and keep using it here.

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/AlekSi/pointer"
+
+	"github.com/percona/pmm-managed/models"
```
BupycHuk, line 27: please fix goimports

BupycHuk (changes requested):

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,327 @@
...
+		dbTask := dbTasks[i]
+		fn := s.wrapTask(task, dbTask.ID)
+		j := s.scheduler.Cron(dbTask.CronExpression).SingletonMode()
+		if !dbTask.StartAt.IsZero() {
+			j = j.StartAt(dbTask.StartAt)
+		}
+		if _, err := j.Tag(dbTask.ID).Do(fn); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
Dasio: I wanted to convert all db tasks first, if no errors occurs, then I reset internal scheduler and add it again.

services/scheduler/scheduler.go
Dasio: Yes. Error is in job. But scheduler doesn't necessary run only jobs. But I can remove it.

services/scheduler/scheduler.go
Dasio: Then it's another abstraction over library jobs. And internal if I do e.g remove job by tag. It works exactly the same way as here.

But okay. :+1:

services/scheduler/scheduler.go
Dasio: For changing existing tasks. E.g. change cron expression.

models/database.go
```diff
@@ -508,6 +508,24 @@ var databaseSchema = [][]string{
 	37: {
 		`ALTER TABLE agents ALTER COLUMN max_query_log_size TYPE BIGINT`,
 	},
+	38: {
+		`CREATE TABLE schedule_jobs (
```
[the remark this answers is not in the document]
Dasio: it's otudated. It's already `scheduled_tasks`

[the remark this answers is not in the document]
BupycHuk: yeah, sorry, I started review before I went to vacation :)

services/scheduler/scheduler.go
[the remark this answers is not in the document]
BupycHuk: When do we call `loadFromDB`? Only on startup?
If so, then scheduler is already clear.

services/scheduler/scheduler.go
BupycHuk: If you will have it, you will be able to just update job instead of removing and readding it.

Dasio: There is no API for updating job.

services/scheduler/scheduler.go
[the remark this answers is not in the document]
Dasio: Yeah, you are right. I'll wrap it one loop :+1:

services/scheduler/scheduler.go
Dasio: If DB entry is changed, `Reload` should be called.
If I add load from DB to wrap task, then there will be DB select many times unnecessary, even though nothing changed.

What do you think @artemgavrilov @oter.

services/scheduler/scheduler.go
BupycHuk: https://github.com/go-co-op/gocron/blob/master/scheduler.go#L900

Dasio: I looked on to it. And I can't update function(+argument, even though we don't pass any). So if something changed inside. For example in PrintTask message. We can't reflect this change.

models/scheduled_tasks_helpers.go
```diff
@@ -0,0 +1,246 @@
...
+		crossJoin = true
+		andConds = append(andConds, "value ->> 'service_id' = "+q.Placeholder(idx))
+		args = append(args, filters.ServiceID)
+		idx++
+	}
+	if filters.LocationID != "" {
+		crossJoin = true
+		andConds = append(andConds, "value ->> 'location_id' = "+q.Placeholder(idx))
+		args = append(args, filters.LocationID)
+		idx++
```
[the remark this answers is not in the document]
Dasio: I'm not sure why. Because it's last and then never used again?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,329 @@
...
+	s.jobsMx.Lock()
+	delete(s.jobs, id)
+	s.jobsMx.Unlock()
+
+	err := s.db.InTransaction(func(tx *reform.TX) error {
+		if err := models.RemoveScheduledTask(tx.Querier, id); err != nil {
+			return err
+		}
+
+		s.mx.Lock()
```
oter, line 150: let's move this block out of the transaction, you need to remove the task after successful commit of transaction.

models/scheduled_tasks_helpers.go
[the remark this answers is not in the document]
oter: yup, because it's last.

services/scheduler/scheduler.go
oter: I agree with @BupycHuk

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,330 @@
...
+	if job != nil {
+		nextRun := job.NextRun().UTC()
+		lastRun := job.LastRun().UTC()
+		params.NextRun = &nextRun
+		params.LastRun = &lastRun
+	} else {
+		l.Errorf("failed to find scheduled task: %v", err)
+	}
+
+	_, err = models.ChangeScheduledTask(s.db.Querier, id, params)
```
oter, line 313: I think transaction is need for `FindScheduledTaskByID` and `ChangeScheduledTask` calls.

services/scheduler/scheduler.go
artemgavrilov: I agree with Max, if we failed to save it here then it means something wrong, and better to return the error to the user.

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,340 @@
...
+
+// Reload removes job from scheduler and add it again from DB.
+func (s *Service) Reload(id string) error {
+	dbTask, err := models.FindScheduledTaskByID(s.db.Querier, id)
+	if err != nil {
+		return err
+	}
+
+	if dbTask.Running {
+		return fmt.Errorf("task is running")
```
artemgavrilov, line 174: Minor 
```suggestion
		return errors.New("task is running")
```

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func (s *Service) loadFromDB() error {
+	s.mx.Lock()
+	defer s.mx.Unlock()
+
+	disabled := false
+	dbTasks, err := models.FindScheduledTasks(s.db.Querier, models.ScheduledTasksFilter{
+		Disabled: &disabled,
+	})
```
artemgavrilov, line 212: Minor
```suggestion
	dbTasks, err := models.FindScheduledTasks(s.db.Querier, models.ScheduledTasksFilter{
		Disabled: pointer.ToBool(false),
	})
```

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,330 @@
...
+// Service is responsive for executing tasks and storing them to DB.
+type Service struct {
+	db        *reform.DB
+	scheduler *gocron.Scheduler
+	l         *logrus.Entry
+	tasks     map[string]context.CancelFunc
+	taskMx    sync.RWMutex
+	jobs      map[string]*gocron.Job
+	jobsMx    sync.RWMutex
+	mx        sync.Mutex
```
artemgavrilov, line 42: I would suggest to sort fields a different way: https://dmitri.shuralyov.com/idiomatic-go#mutex-hat
```suggestion
	db        *reform.DB
	l         *logrus.Entry
	
	mx        sync.Mutex
	scheduler *gocron.Scheduler

	taskMx    sync.RWMutex
	tasks     map[string]context.CancelFunc

	jobsMx    sync.RWMutex
	jobs      map[string]*gocron.Job
```

models/database.go
```diff
@@ -517,6 +517,27 @@ var databaseSchema = [][]string{
...
+			id VARCHAR NOT NULL,
+			cron_expression VARCHAR NOT NULL CHECK (cron_expression <> ''),
+			type VARCHAR NOT NULL CHECK (type <> ''),
+			start_at TIMESTAMP,
+			last_run TIMESTAMP,
+			next_run TIMESTAMP,
+			data JSONB,
+			disabled BOOLEAN,
+			succeeded INTEGER,
+			failed INTEGER,
```
artemgavrilov, line 531: I would suggest to sort fields a different way

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+}
+
+func (s *Service) convertDBTask(dbTask *models.ScheduledTask) (Task, error) {
+	var task Task
+	switch dbTask.Type {
+	case models.ScheduledPrintTask:
+		task = NewPrintTask(dbTask.Data.Print.Message)
+	default:
+		return task, fmt.Errorf("unknown task type: %s", dbTask.Type)
```
artemgavrilov, line 338: Minor
```suggestion
		return task, errors.Errorf("unknown task type: %s", dbTask.Type)
```

models/database.go
Dasio: We don't use it. I just thought It'll be good to have such counters. I can delete it.

artemgavrilov: After one year it can be something like:
```
succeeded: 352
failed: 4
```
I'm not sure that this information can be somehow useful. Probably such counters make sense for a single run with retries, but most likely it will be part of job retries. If there is no need for this now, then let's remove it.

services/scheduler/scheduler.go
artemgavrilov: I need a few minutes to figure out the problem. 🤔

artemgavrilov: It looks like a task should be added to the scheduler(library) inside a db transaction. For example, if we will have two concurrent `Reload calls` it may look like this:
```
load_from_db_1 ---> load_from_db_2 ---> update_scheduller_2 ---> update_scheduller_1
```

In this situation, we will have the new version in DB, but the old version is the scheduler.
So it looks that this pair `load_from_db, update_scheduler` should be always wrapped into TX, no?

artemgavrilov: |  if it should reload when task is actually run. But from my point of view, I don't think it's necessary.
The current implementation of `Reload` allows already triggered tasks to finish their jobs, am I right?

Dasio: Yes, you can't reload if task is running.

artemgavrilov: Well, it sounds good to me.

artemgavrilov: Oh, what do you mean by `can't`?

Dasio: Right now there is
```
	if dbTask.Running {
		return errors.New("task is running")
	}
```
But probably it's not necessary. If task is running it will finish with old data.

Dasio: With artem we discussed that it will be better have DB update and scheduler internal update inside one transaction. So I rename `Reload` to `Update` and now it also updates DB and then reload it immediately. 

So let's say I have already running task. I'll update it. Running task is still running with old data. But future tasks will be changed.

I don't see point now, to select DB task inside wrap function.

models/scheduled_tasks_helpers.go
```diff
@@ -0,0 +1,234 @@
...
+	return nil
+}
+
+func checkUniqueScheduledTaskID(q *reform.Querier, id string) error {
+	if id == "" {
+		panic("empty schedule task ID")
+	}
+
+	location := &ScheduledTask{ID: id}
+	switch err := q.Reload(location); err {
```
oter, line 226:
```suggestion
	st := &ScheduledTask{ID: id}
	switch err := q.Reload(st); err {
```

services/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,76 @@
...
+func setup(t *testing.T) *Service {
+	t.Helper()
+	sqlDB := testdb.Open(t, models.SkipFixtures, nil)
+	db := reform.NewDB(sqlDB, postgresql.Dialect, reform.NewPrintfLogger(t.Logf))
+	return New(db)
+}
+func TestService(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	svc := setup(t)
```
oter, line 44: let's add a rollback to clean up the db after the test finishes.

oter (approved):
LGTM

services/scheduler/scheduler.go
BupycHuk: Why would you update function? If we generate PrintTask inside function based on task id, then everything will work. We just should update cron.

models/scheduled_tasks_helpers.go
```diff
@@ -0,0 +1,234 @@
...
+// ScheduledTasksFilter represents filters for scheduled tasks.
+type ScheduledTasksFilter struct {
+	Disabled   *bool
+	Types      []ScheduledTaskType
+	ServiceID  string
+	LocationID string
+}
+
+// FindScheduledTasks returns all scheduled tasks satisfying filter.
+func FindScheduledTasks(q *reform.Querier, filters ScheduledTasksFilter) ([]*ScheduledTask, error) {
```
BupycHuk, line 58: Can we write a few unit tests for this function?

models/scheduled_tasks_helpers.go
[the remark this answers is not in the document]
BupycHuk: then we should remove it

models/scheduled_tasks_helpers.go
BupycHuk: And for other functions in this file

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,330 @@
...
+		j := s.scheduler.Cron(dbTask.CronExpression).SingletonMode()
+		if !dbTask.StartAt.IsZero() {
+			j = j.StartAt(dbTask.StartAt)
+		}
+		scheduleJob, err := j.Tag(dbTask.ID).Do(fn)
+		if err != nil {
+			return err
+		}
+
+		s.jobs[dbTask.ID] = scheduleJob
```
BupycHuk, line 237: can we extract it to separate method and just call it in `Add`?

models/scheduled_tasks_helpers_test.go
```diff
@@ -0,0 +1,158 @@
...
+		createParams2.Disabled = true
+		task2, err := models.CreateScheduledTask(findTX.Querier, createParams2)
+		require.NoError(t, err)
+
+		type testCase struct {
+			filter models.ScheduledTasksFilter
+			ids    []string
+		}
+
+		tests := []testCase{
```
BupycHuk, line 126: What about other fields in `ScheduledTasksFilter`?

services/scheduler/scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+
+func (s *Service) addDBTask(dbTask *models.ScheduledTask) error {
+	task, err := s.convertDBTask(dbTask)
+	if err != nil {
+		return err
+	}
+
+	if dbTask.Disabled {
+		return nil
+	}
```
BupycHuk, line 199: I think it should be done before convert.

BupycHuk (approved):

models/scheduled_tasks_helpers_test.go
Dasio: Rigth now there is only one task type (in this PR), so I don't know to test it.

BupycHuk: You are testing model, so you can pass any value to `Types` even if they are not in our logic. e.g.: `ScheduledTaskType("writesomethingintest")`
Don't forget about `LocationID` and `ServiceID`

Dasio: I can't because there is check for type. So I can't create it by our helper.
I added more tests here https://github.com/percona/pmm-managed/pull/778/commits/d342a83962f2f6df488f9bea9d7aa582f0e2ff6f

PR #797, opened 2021-07-02
PMM-8219 Blank PMM version in offline mode.

JiriCtvrtka:
[PMM-8219](https://jira.percona.com/browse/PMM-8219)
FE part: https://jira.percona.com/browse/PMM-8397

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1899

pmm: https://github.com/percona/pmm/pull/762
---
- [x] Tests passed.
- [ ] Feature build pass.
- [x] (Re)requested review.
- [ ] Update API dependency.


services/server/server.go
```diff
@@ -247,13 +270,13 @@ func (s *Server) CheckUpdates(ctx context.Context, req *serverpb.CheckUpdatesReq
 
 	if req.Force {
 		if err := s.supervisord.ForceCheckUpdates(ctx); err != nil {
-			return nil, err
+			return s.noInternetResponse(ctx)
```
BupycHuk, line 273: We shouldn't ignore error message.
Can we return error message and installed version together?
If we can't we should at least log the error and maybe add new flag to return only installed version.

BupycHuk (changes requested):

services/server/server.go
JiriCtvrtka: I can modify struct to have field for error where we can mention that PMM is offline, but this changes also needs some changes on FE.

About flag is it really needed? Like if we will modify FE then we can do it on FE. What do you think?

BupycHuk: Flag looks more proper because it's consistent with other out APIs.

askomorokhov (approved):

BupycHuk (approved):

PR #800, opened 2021-07-13
PMM-5016 Enforce relationships between agents and services/nodes

Dasio:
[PMM-5016](https://jira.percona.com/browse/PMM-5016)

In ticket is said, that enforce DB checks. But it seems complex for me and what about cases when there is already invalid combination.

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1916

- API tests: https://github.com/Percona-Lab/pmm-api-tests/pull/226
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/agent_helpers.go
```diff
@@ -637,6 +637,63 @@ type CreateAgentParams struct {
...
+			return nil
+		}
+	default:
+		return nil
+	}
+	return errors.Errorf("invalid combination of node type %s and agent type %s", nodeType, agentType)
+}
+
+func verifyServiceAndAgent(serviceType ServiceType, agentType AgentType) error {
+	switch agentType {
```
oter, line 665: maybe try to use a map for describing allowed relationships between agents and services? e.g.
```
m := map[AgentType]*ServiceType{
	PMMAgentType: nil,
	MySQLdExporterType: ServiceTypePointer(MySQLServiceType),
...
}
```

models/agent_helpers.go
```diff
@@ -637,6 +637,63 @@ type CreateAgentParams struct {
...
+	case PMMAgentType:
+	case NodeExporterType:
+	case RDSExporterType:
+	case AzureDatabaseExporterType:
+	case VMAgentType:
```
BupycHuk, line 670: I think we can remove these cases and use map mentioned by @oter, because we shouldn't try to bind service for these agents.
For all cases out of that map we can return error.

models/agent_helpers.go
```diff
@@ -637,6 +637,63 @@ type CreateAgentParams struct {
 	DisableCollectors              []string
 }
 
+func verifyNodeAndAgent(nodeType NodeType, agentType AgentType) error {
```
BupycHuk, line 640: I think we can make these two functions return just boolean and return error in `CreateAgent`, in that case code will be simpler

BupycHuk (changes requested):

artemgavrilov (approved):
LGTM besides already existing comments

oter (approved):

models/agent_helpers.go
Dasio: Right now if it's out of map, I return as valid. Should I switch to false? Not sure what is better approach.
if we add some new agent type, allow it by default or forbid it by default.

models/agent_helpers.go
```diff
@@ -637,6 +637,55 @@ type CreateAgentParams struct {
...
+		GenericNodeType:             allowAll,
+		ContainerNodeType:           allowAll,
+		RemoteNodeType:              ExternalExporterType,
+		RemoteRDSNodeType:           RDSExporterType,
+		RemoteAzureDatabaseNodeType: AzureDatabaseExporterType,
+	}
+
+	allowed, ok := allow[nodeType]
+	if !ok {
+		return true
```
BupycHuk, line 652:
```suggestion
		return false
```

models/agent_helpers.go
```diff
@@ -637,6 +637,55 @@ type CreateAgentParams struct {
...
+		QANMongoDBProfilerAgentType:         MongoDBServiceType,
+		PostgresExporterType:                PostgreSQLServiceType,
+		QANPostgreSQLPgStatMonitorAgentType: PostgreSQLServiceType,
+		QANPostgreSQLPgStatementsAgentType:  PostgreSQLServiceType,
+		ExternalExporterType:                ExternalServiceType,
+	}
+
+	allowed, ok := allow[agentType]
+	if !ok {
+		return true
```
BupycHuk, line 683:
```suggestion
		return false
```

models/agent_helpers.go
```diff
@@ -637,6 +637,55 @@ type CreateAgentParams struct {
...
+}
+
+func compatibleServiceAndAgent(serviceType ServiceType, agentType AgentType) bool {
+	const forbidden = "forbidden"
+	allow := map[AgentType]ServiceType{
+		PMMAgentType:                        forbidden,
+		NodeExporterType:                    forbidden,
+		RDSExporterType:                     forbidden,
+		AzureDatabaseExporterType:           forbidden,
+		VMAgentType:                         forbidden,
```
BupycHuk, line 669: let's remove them.

BupycHuk (approved):

PR #802, opened 2021-07-15
PMM-8430 Backup retention

Dasio:
[PMM-8430](https://jira.percona.com/browse/PMM-8430)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/1918

- API: https://github.com/percona/pmm/pull/764 
---
- [x] Tests passed.
- [x] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


services/backup/retention_service.go
```diff
@@ -0,0 +1,97 @@
...
+	return &RetentionService{
+		l:          logrus.WithField("component", "management/backup/retention"),
+		db:         db,
+		removalSVC: removalSVC,
+	}
+}
+
+// EnforceRetention enforce retention on provided scheduled backup task
+// it removes any old successful artifacts below retention threshold.
+func (s *RetentionService) EnforceRetention(ctx context.Context, scheduleID string) error {
```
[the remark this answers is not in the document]
Dasio: I think that splitting this to 2 functions wouldn't benefit for readability.

services/backup/retention_service.go
```diff
@@ -0,0 +1,97 @@
...
+	var artifacts []*models.Artifact
+	var retention uint32
+
+	txErr := s.db.InTransaction(func(tx *reform.TX) error {
```
BupycHuk, line 51: in both places where you call it, it's already in transaction.

services/backup/retention_service.go
```diff
@@ -0,0 +1,97 @@
...
+	if txErr != nil {
+		return txErr
+	}
+
+	if retention == 0 || int(retention) > len(artifacts) {
+		return nil
+	}
+
+	for _, artifact := range artifacts[retention:] {
+		if err := s.removalSVC.DeleteArtifact(ctx, artifact.ID, true); err != nil {
```
BupycHuk, line 90: if we are calling this service it's better to call this method in a separate gourutine. What do you think?

Dasio: 1 goroutine per artifact. Or just single goroutine for everything?
Also I think if one failed, it shouldn't return error immediately and try remove other ones.

BupycHuk: 1 goroutine per scheduled task

Dasio: well, there is always only one schedueld task. You probably mean, per artifact?

BupycHuk: We call `handleResult` in the same goroutine where we handle all responses from pmm-agent.
Then we call `EnforceRetention` in the same goroutine where we call `handleResult`, so we block whole channel while we are deleting artifacts.

Dasio: Ah, got it :+1:

BupycHuk (approved):

services/agents/registry.go
```diff
@@ -312,23 +314,39 @@ func (r *Registry) handleJobResult(l *logrus.Entry, result *agentpb.JobResult) {
...
+			artifact, err := models.UpdateArtifact(t.Querier, res.Result.MySQLBackup.ArtifactID, models.UpdateArtifactParams{
 				Status: models.BackupStatusPointer(models.SuccessBackupStatus),
 			})
 			if err != nil {
 				return err
 			}
+
+			if artifact.Type == models.ScheduledArtifactType {
+				go func() {
+					if err := r.retentionService.EnforceRetention(ctx, artifact.ScheduleID); err != nil {
```
oter, line 326: This call has no querier argument, so in my opinion it shouldn't be inside of the transaction. Because in case of transaction rollback this call shouldn't happen.

services/backup/retention_service.go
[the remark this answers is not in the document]
oter: you could extract retention retrieving e.g lines 48-62

oter (approved):
LGTM besides existing comments.

artemgavrilov (approved):

services/agents/registry.go
Dasio: any suggestion how to properly do it? new switch outside transaction?
But then I'll need again to select artifact.

Dasio: Or maybe pass querier to it and remove DB from retention service?

PR #807, opened 2021-07-26
PMM-8460: software versions cache

oter:
[PMM-8460](https://jira.percona.com/browse/PMM-8460)

API PR: https://github.com/percona/pmm/pull/766
agent PR: https://github.com/percona/pmm-agent/pull/283

- [x] FB: https://github.com/Percona-Lab/pmm-submodules/pull/1941

---
- [x] Tests passed.
- [x] Feature build pass.
- [x] (Re)requested review.
- [x] Update API dependency.


models/database.go
```diff
@@ -542,6 +542,25 @@ var databaseSchema = [][]string{
...
 	},
+	41: {
+		`CREATE TYPE software_version AS (
+  			name VARCHAR NOT NULL CHECK (name <> ''),
+  			version VARCHAR NOT NULL CHECK (version <> '')
+		);`,
+		`CREATE TABLE service_software_versions (
+			id VARCHAR NOT NULL,
+			service_id VARCHAR NOT NULL CHECK (service_id <> ''),
+			software_versions ARRAY software_version,
```
oter, line 553: looks like ARRAY is not supported by the reform lib, so probably need to change to JSONB

Dasio: in docs i see:
```
pay_by_quarter  integer ARRAY,
```

models/software_version_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+// FindServicesSoftwareVersions returns services software versions.
+func FindServicesSoftwareVersions(q *reform.Querier, limit *int) ([]*ServiceSoftwareVersions, error) {
+	var args []interface{}
+	var limitStatement string
+	if limit != nil {
+		limitStatement = " LIMIT $1"
+		args = append(args, *limit)
+	}
+
+	const orderByField = "check_at"
```
artemgavrilov, line 148: Does it make sense to have it as const? I mean why not just `fmt.Sprintf("ORDER BY check_at %s"...`?

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,284 @@
...
+			return errors.Errorf("pmmAgent not found for service")
+		}
+		results.AgentID = pmmAgents[0].AgentID
+
+		checkAt := time.Now().UTC().Add(serviceCheckInterval) // TODO
+		if _, err := models.UpdateServiceSoftwareVersions(tx.Querier, servicesVersions[0].ID,
+			models.UpdateServiceSoftwareVersionsParams{CheckAt: &checkAt},
+		); err != nil {
+			return err
+		}
```
artemgavrilov, line 167: It seems that we should update this timestamp after the actual versions update, not before.

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,284 @@
...
+			ca, err := s.updateVersions()
+			if err != nil {
+				s.l.Warn(err)
+			}
+
+			checkAfter = ca
+		case <-syncTicker.C:
+			if err := s.syncServices(); err != nil {
+				s.l.Warn(err)
+			}
```
artemgavrilov, line 268: Mmmm.... why do we need to update services every hour?

models/database.go
```diff
@@ -549,6 +549,25 @@ var databaseSchema = [][]string{
...
+			check_at TIMESTAMP,
+
+			created_at TIMESTAMP NOT NULL,
+			updated_at TIMESTAMP NOT NULL,
+
+			PRIMARY KEY (id),
+			UNIQUE (service_id),
+			FOREIGN KEY (service_id) REFERENCES services (service_id)
```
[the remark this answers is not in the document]
Dasio: foreign key doesn't make field unique.

Anyway if it's 1-1 relation ship. Isn't better to use service_id as primary key?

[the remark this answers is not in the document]
oter: service_id used as primary key now.

models/software_version.go
```diff
@@ -0,0 +1,80 @@
...
+	Version string       `reform:"version"`
+}
+
+// ServiceSoftwareVersions represents service software versions.
+//reform:service_software_versions
+type ServiceSoftwareVersions struct {
+	ID               string            `reform:"id,pk"`
+	ServiceID        string            `reform:"service_id"`
+	SoftwareVersions []SoftwareVersion `reform:"software_versions"`
+	CheckAt          time.Time         `reform:"check_at"`
```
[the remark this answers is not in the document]
oter: when the version will be checked next time

models/software_version_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+
+// CreateServiceSoftwareVersionsParams are params for creating a new service software versions entry.
+type CreateServiceSoftwareVersionsParams struct {
+	ServiceID string
+	Versions  []SoftwareVersion
+	CheckAt   time.Time
+}
+
+// Validate validates params used for creating an service software versions entry.
+func (p *CreateServiceSoftwareVersionsParams) Validate() error {
```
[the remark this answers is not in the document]
oter: validation added

models/software_version_helpers.go
oter: there was additional ordering by ID. Constant removed.

services/versioncache/versioncache.go
oter: yeah, it's too frequent. Probably better to check every day?

models/service_helpers.go
```diff
@@ -328,5 +343,9 @@ func RemoveService(q *reform.Querier, id string, mode RemoveMode) error {
 		panic(fmt.Errorf("unhandled RemoveMode %v", mode))
 	}
 
+	if err := deleteAssociatedSoftwareVersions(q, id); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
oter: done

services/versioncache/versioncache.go
oter: let's have a call regarding this

oter: solution: add clarification note

models/software_version.go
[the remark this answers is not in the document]
oter: rename CheckAt -> NextCheckAt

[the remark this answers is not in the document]
oter: done

services/versioncache/versioncache.go
oter: interval increased to one day

models/software_version.go
```diff
@@ -0,0 +1,101 @@
...
+	return sv, nil
+}
+
+// Scan implements database/sql.Scanner interface. Should be defined on the pointer.
+func (sv *SoftwareVersions) Scan(src interface{}) error {
+	switch s := src.(type) {
+	case SoftwareVersions:
+		*sv = s
+	default:
+		return errors.Errorf("error")
```
Dasio, line 60: let's give better message :D

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// prepareUpdateVersions checks if there is any service that needs software versions update in the cache and
+// shifts the next check time for this service.
+func (s *Service) prepareUpdateVersions() (*prepareResults, error) {
+	results := &prepareResults{CheckAfter: minCheckInterval}
+
+	if err := s.db.InTransaction(func(tx *reform.TX) error {
+		limit := 1
+		servicesVersions, err := models.FindServicesSoftwareVersions(tx.Querier, &limit)
```
Dasio, line 126: you can use pointer.ToInt.. as in other places in our code base

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil {
+			return err
+		}
+		if len(servicesVersions) == 0 {
+			results.CheckAfter = serviceCheckInterval
+
+			return nil
+		}
+		if servicesVersions[0].NextCheckAt.After(time.Now().UTC()) {
```
Dasio, line 135: `.UTC()` shouldn't have any effect

Dasio (approved):
LGTM

models/software_version_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, errors.WithStack(err)
+	}
+}
+
+// FindServicesSoftwareVersions returns services software versions.
+func FindServicesSoftwareVersions(q *reform.Querier, limit *int) ([]*ServiceSoftwareVersions, error) {
+	var args []interface{}
+	var limitStatement string
+	if limit != nil {
+		limitStatement = " LIMIT $1"
```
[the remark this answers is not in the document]
oter: done

models/software_version.go
```diff
@@ -0,0 +1,101 @@
...
+	QpressSoftwareName     SoftwareName = "qpress"
+)
+
+// SoftwareVersion represents version of the given software.
+type SoftwareVersion struct {
+	Name    SoftwareName `reform:"name"`
+	Version string       `reform:"version"`
+}
+
+type SoftwareVersions []SoftwareVersion
```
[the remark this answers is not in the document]
oter: comment added

services/versioncache/versioncache.go
oter: yes, you are right, removed.

services/versioncache/versioncache.go
oter: done

models/software_version.go
oter: done

models/database.go
```diff
@@ -550,13 +550,9 @@ var databaseSchema = [][]string{
...
 	},
 	43: {
-		`CREATE TYPE software_version AS (
-  			name VARCHAR,
-  			version VARCHAR
-		);`,
 		`CREATE TABLE service_software_versions (
 			service_id VARCHAR NOT NULL CHECK (service_id <> ''),
-			software_versions software_version ARRAY,
+			software_versions JSONB,
```
Dasio, line 555: it doesn't work after all?

oter: inserts with empty array worked, but if it has some entries it require to define Scan() and Value() functions for the array element type, I decided to use JSONB, because it's easier to use.. plus, we use it in the other places.

models/software_version_helpers.go
```diff
@@ -0,0 +1,199 @@
...
+	return versions, nil
+}
+
+// DeleteServiceSoftwareVersions removes entry from the DB by service ID.
+func DeleteServiceSoftwareVersions(q *reform.Querier, serviceID string) error {
+	if _, err := FindServiceSoftwareVersions(q, serviceID); err != nil {
+		return err
+	}
+
+	if err := q.Delete(&ServiceSoftwareVersions{ServiceID: serviceID}); err != nil {
```
BupycHuk, line 195: this method deletes only one record by primary key, so probably it won't work.

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+	case *agents.Qpress:
+		softwareName = models.QpressSoftwareName
+	default:
+		return "", errors.Errorf("invalid software type %T", software)
+	}
+
+	return softwareName, nil
+}
+
+func (s *Service) updateVersions() (time.Duration, error) {
```
BupycHuk, line 222: so updateVersions updates version only for one service.
It definitely good to write here.

BupycHuk (changes requested):
I would suggest to update software versions once per day for all services. So we shouldn't take care of `CheckAfter` value for each service.
Current implementation looks a little bit overcomplicated. I spent few hours to review whole PR :)

models/software_version_helpers.go
oter: but why? I have a test for this method :)

models/software_version_helpers.go
```diff
@@ -0,0 +1,199 @@
...
+		case XtrabackupSoftwareName:
+		case XbcloudSoftwareName:
+		case QpressSoftwareName:
+		default:
+			return errors.Wrapf(ErrInvalidArgument, "invalid software name %q", sv.Name)
+		}
+	}
+
+	return nil
+}
```
artemgavrilov, line 114:
```suggestion
type UpdateServiceSoftwareVersionsParams struct {
	NextCheckAt      *time.Time
	SoftwareVersions []SoftwareVersion
}

// Validate validates params used for updating a service software versions entry.
func (u *UpdateServiceSoftwareVersionsParams) Validate() error {
	for _, sv := range u.SoftwareVersions {
		switch sv.Name {
		case MysqldSoftwareName:
		case XtrabackupSoftwareName:
		case XbcloudSoftwareName:
		case QpressSoftwareName:
		default:
			return errors.Wrapf(ErrInvalidArgument, "invalid software name %q", sv.Name)
		}
	}

	return nil
}
```

services/agents/versioner.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Software interface.
+type Software interface {
+	isSoftware()
+}
+
+// Mysqld represents mysqld software.
+type Mysqld struct{}
+
+func (*Mysqld) isSoftware() {}
```
artemgavrilov, line 51: Why do we need this as types, why not just string constants?

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+type Service struct {
+	cfg      Config
+	db       *reform.DB
+	l        *logrus.Entry
+	v        Versioner
+	updateCh chan struct{}
+}
+
+// New creates new service.
+func New(cfg Config, db *reform.DB, v Versioner) *Service {
```
artemgavrilov, line 72: With the current approach, you can't specify only one parameter from the config. And why do we need this config at all? The server will have only one version service, so probably better let to change it with env variables (if it makes sense to change them). Take a look how it's made in `checks.Service`

oter: I'm using that config in the test, so test won't execute for hours.

services/agents/versioner.go
oter: later for the mongodb I will need couple of fields for requesting db version.

artemgavrilov: Ah, got it!

services/versioncache/versioncache.go
artemgavrilov: In tests you can change `versioner.Service` fields directly because tests are in the same package.

models/software_version_helpers.go
```diff
@@ -0,0 +1,199 @@
...
+// FindServicesSoftwareVersionsFilter represents a filter for finding service software versions.
+type FindServicesSoftwareVersionsFilter struct {
+	Limit *int
+}
+
+// FindServicesSoftwareVersions returns all services software versions.
+func FindServicesSoftwareVersions(q *reform.Querier, filter FindServicesSoftwareVersionsFilter) ([]*ServiceSoftwareVersions, error) {
+	var args []interface{}
+	var tail strings.Builder
+	tail.WriteString("ORDER BY next_check_at ")
```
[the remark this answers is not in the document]
oter: the ORDER BY uses ASC by default, should I set it explicitly?

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+
+//go:generate mockery -name=Versioner -case=snake -inpkg -testonly
+
+// Versioner contains method for retrieving versions of different software.
+type Versioner interface {
+	GetVersions(pmmAgentID string, softwares []agents.Software) ([]agents.Version, error)
+}
+
+// Service is responsible for caching service software versions in the DB.
+type Service struct {
```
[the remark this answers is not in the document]
oter: in the other places we use this naming approach, so I access this service via `versioncache.Service`.

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+	})
+
+	return err
+}
+
+type prepareResults struct {
+	ServiceID   string
+	CheckAfter  time.Duration
+	NeedsUpdate bool
+	AgentID     string
```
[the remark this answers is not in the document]
oter: yup, renamed

services/versioncache/versioncache.go
oter: previously this config was represented as a set of const variables.

models/software_version_helpers.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if err := q.Update(row); err != nil {
+		return nil, errors.Wrap(err, "failed to update service software versions")
+	}
+
+	return row, nil
+}
+
+// FindServiceSoftwareVersions returns service software versions entry by given service ID if found, ErrNotFound if not.
+func FindServiceSoftwareVersions(q *reform.Querier, serviceID string) (*ServiceSoftwareVersions, error) {
```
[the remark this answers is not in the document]
oter: done

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+
+	svs := make([]models.SoftwareVersion, 0, len(softwares))
+	for i, software := range softwares {
+		name, err := softwareName(software)
+		if err != nil {
+			return s.cfg.MinCheckInterval, err
+		}
+
+		if versions[i].Error != "" {
+			s.l.Warnf("failed to get version of %q software: %s", name, versions[i].Error)
```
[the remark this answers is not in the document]
oter: no, because there could be the case when the error happens in some of the above statements, so we will have a constant loop of trying to update software versions for the same service (e.g. if it's broken).

services/versioncache/versioncache.go
oter: done

models/software_version_helpers.go
```diff
@@ -0,0 +1,199 @@
...
+
+	for _, sv := range p.SoftwareVersions {
+		switch sv.Name {
+		case MysqldSoftwareName:
+		case XtrabackupSoftwareName:
+		case XbcloudSoftwareName:
+		case QpressSoftwareName:
+		default:
+			return errors.Wrapf(ErrInvalidArgument, "invalid software name %q", sv.Name)
+		}
```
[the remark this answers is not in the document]
oter: this case will go directly to the default case, should I distinguish empty value too?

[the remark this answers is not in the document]
artemgavrilov: It's really minor comment, typically we have separate case for empry values just to provide more clear error message. In general, it's up to you.

services/versioncache/versioncache.go
oter: reworked to use default variables

models/software_version_helpers.go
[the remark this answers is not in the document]
oter: this is definitely the error because of the engineer, so normally it should not happen, so probably will leave as is.

models/software_version_helpers.go
oter: done 👍

services/management/mysql.go
```diff
@@ -183,5 +185,7 @@ func (s *MySQLService) Add(ctx context.Context, req *managementpb.AddMySQLReques
 	}
 
 	s.state.RequestStateUpdate(ctx, req.PmmAgentId)
+	s.vc.SyncAndUpdate()
```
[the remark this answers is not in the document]
oter: Yes, I can, but my current approach has better consistency, because it syncs available services and schedules versions updates for missing ones.

artemgavrilov (approved):

models/software_version_helpers.go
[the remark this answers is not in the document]
artemgavrilov: Yep, I think better to keep ordering explicit when it matters

models/software_version_helpers.go
```diff
@@ -0,0 +1,199 @@
...
+// CreateServiceSoftwareVersionsParams are params for creating a new service software versions entry.
+type CreateServiceSoftwareVersionsParams struct {
+	ServiceID        string
+	ServiceType      ServiceType
+	SoftwareVersions []SoftwareVersion
+	NextCheckAt      time.Time
+}
+
+// Validate validates params used for creating a service software versions entry.
+func (p *CreateServiceSoftwareVersionsParams) Validate() error {
```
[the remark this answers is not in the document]
oter: this is a validation, no business logic here

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,303 @@
...
+		softwareName = models.QpressSoftwareName
+	default:
+		return "", errors.Errorf("invalid software type %T", software)
+	}
+
+	return softwareName, nil
+}
+
+// updateVersions updates software versions for one service.
+func (s *Service) updateVersions() (time.Duration, error) {
```
BupycHuk, line 205: let's rename this method to make it clear.

services/versioncache/versioncache.go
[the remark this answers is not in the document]
BupycHuk: Yeah, correct, let's do it in error check above.

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,303 @@
...
+
+	softwares := []agents.Software{&agents.Mysqld{}, &agents.Xtrabackup{}, &agents.Xbcloud{}, &agents.Qpress{}}
+	versions, err := s.v.GetVersions(r.PMMAgentID, softwares)
+	if err != nil {
+		return minCheckInterval, err
+	}
+	if len(versions) != len(softwares) {
+		return minCheckInterval, errors.Errorf("slices length mismatch: versions len %d != softwares len %d",
+			len(versions), len(softwares))
+	}
```
BupycHuk, line 223: this check is not backward compatible, in case we add new softwares, but do not update client there always will be length mismatch.

services/versioncache/versioncache.go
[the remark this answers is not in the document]
BupycHuk: I mean that versions for the service won't be updated until next day, it's better to check it again earlier.

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,303 @@
...
+func New(db *reform.DB, v Versioner) *Service {
+	return &Service{
+		db:       db,
+		l:        logrus.WithField("component", "version-cache"),
+		v:        v,
+		updateCh: make(chan struct{}, 1),
+	}
+}
+
+func (s *Service) syncServices() error {
```
BupycHuk, line 64: I don't think that we need this method at all.
We can create ServiceSoftwareVersions in the moment of migration using `insert into items_ver (item_id, name, item_group)
select item_id, name, item_group from items where item_id=2;
` construction and when we create a service. 
if we will remove ServiceSoftwareVersion for service the moment we delete service we won't have to do it here.

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,303 @@
...
+type prepareResults struct {
+	ServiceID   string
+	CheckAfter  time.Duration
+	NeedsUpdate bool
+	PMMAgentID  string
+}
+
+// prepareUpdateVersions checks if there is any service that needs software versions update in the cache and
+// shifts the next check time for this service.
+func (s *Service) prepareUpdateVersions() (*prepareResults, error) {
```
BupycHuk, line 126: `findServiceForTheNextCheck`?

services/versioncache/versioncache.go
BupycHuk: and the same in `GetVersions`

services/management/mysql.go
BupycHuk, line 188: maybe it's better to call `RequestSoftwareVersionsUpdate` for better consistency in names.

models/software_version_helpers.go
BupycHuk: Oh, I see, we have one record for each service.

BupycHuk (commented):
If we have 1-1 relationship between ServiceSoftwareVersions and service maybe it's better to keep SoftwareVersions as field of Service like MySQLOptions and other fields? What do you think?

services/versioncache/versioncache.go
oter: returning filled `Error` field instead of the `Unknown` error: https://github.com/percona/pmm-agent/pull/283/commits/ecfa0dbc846ca5ff42d038823f385f37ba35e8e2

services/management/mysql.go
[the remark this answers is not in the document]
oter: reworked

oter: renamed

services/versioncache/versioncache.go
oter: syncServices removed, inserting on service insertion

services/versioncache/versioncache.go
oter: renamed to `findServiceForUpdate`

services/versioncache/versioncache.go
oter: renamed to `updateVersionsForNextService`

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+			}
+		}
+
+		return nil
+	})
+
+	return err
+}
+
+type prepareResults struct {
```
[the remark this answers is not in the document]
oter: renamed

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,320 @@
...
+		results.AgentID = pmmAgents[0].AgentID
+
+		// shift the next check time for this service, so, in case of versions fetch error,
+		// it will not loop in trying, but will continue with other services.
+		nextCheckAt := time.Now().UTC().Add(s.cfg.ServiceCheckInterval)
+		if _, err := models.UpdateServiceSoftwareVersions(tx.Querier, servicesVersions[0].ServiceID,
+			models.UpdateServiceSoftwareVersionsParams{NextCheckAt: &nextCheckAt},
+		); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
oter: so now I'm setting a short interval of 4h right after finding and setting to 24h interval on successful versions update.

services/versioncache/versioncache.go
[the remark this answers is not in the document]
oter: updated

BupycHuk (changes requested):
migration is missed, and SSV should be created only for MySQL service, it's critical ones. 
Other comments are minor for better readability.

models/software_version_helpers.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, errors.WithStack(err)
+	}
+}
+
+// FindServicesSoftwareVersionsFilter represents a filter for finding service software versions.
+type FindServicesSoftwareVersionsFilter struct {
+	Limit *int
+}
+
+// FindServicesSoftwareVersions returns all services software versions.
```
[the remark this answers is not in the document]
oter: clarified

models/service_helpers.go
```diff
@@ -254,6 +255,17 @@ func AddNewService(q *reform.Querier, serviceType ServiceType, params *AddDBMSSe
 		return nil, errors.WithStack(err)
 	}
 
+	const serviceVersionsFirstCheckDelay = 30 * time.Second
```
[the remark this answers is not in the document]
oter: done

models/service_helpers.go
```diff
@@ -254,6 +255,17 @@ func AddNewService(q *reform.Querier, serviceType ServiceType, params *AddDBMSSe
...
+	if _, err := CreateServiceSoftwareVersions(q, CreateServiceSoftwareVersionsParams{
```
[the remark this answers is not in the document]
oter: thanks, fixed

models/database.go
```diff
@@ -553,6 +553,20 @@ var databaseSchema = [][]string{
...
+			next_check_at TIMESTAMP,
+
+			created_at TIMESTAMP NOT NULL,
+			updated_at TIMESTAMP NOT NULL,
+
+			PRIMARY KEY (service_id),
+			FOREIGN KEY (service_id) REFERENCES services (service_id) ON DELETE CASCADE
+		);`,
+	},
 }
```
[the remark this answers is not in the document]
oter: added, thanks

models/service_helpers.go
```diff
@@ -254,6 +255,17 @@ func AddNewService(q *reform.Querier, serviceType ServiceType, params *AddDBMSSe
...
+		ServiceType:      serviceType,
+		SoftwareVersions: []SoftwareVersion{},
+		// add a small duration ahead, so first software versions check will happen when agent established a connection.
+		NextCheckAt: time.Now().Add(serviceVersionsFirstCheckDelay),
+	}); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
 	return row, nil
 }
```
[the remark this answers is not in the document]
oter: added

models/software_version.go
```diff
@@ -0,0 +1,92 @@
...
+// SoftwareVersions represents slice of SoftwareVersion.
+type SoftwareVersions []SoftwareVersion
+
+// Value implements database/sql/driver.Valuer interface. Should be defined on the value.
+func (sv SoftwareVersions) Value() (driver.Value, error) { return jsonValue(sv) }
+
+// Scan implements database/sql.Scanner interface. Should be defined on the pointer.
+func (sv *SoftwareVersions) Scan(src interface{}) error { return jsonScan(sv, src) }
+
+// ServiceSoftwareVersions represents service software versions.
```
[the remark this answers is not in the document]
oter: mentioned

services/versioncache/versioncache.go
```diff
@@ -0,0 +1,241 @@
...
+
+			return nil
+		}
+		if servicesVersions[0].NextCheckAt.After(time.Now()) {
+			results.CheckAfter = time.Until(servicesVersions[0].NextCheckAt)
+			if results.CheckAfter < minCheckInterval {
+				results.CheckAfter = minCheckInterval
+			}
+
+			return nil
```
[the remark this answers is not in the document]
oter: done

services/versioncache/versioncache.go
```diff
@@ -62,16 +62,16 @@ func New(db *reform.DB, v Versioner) *Service {
 }
 
 type service struct {
-	ServiceID   string
-	CheckAfter  time.Duration
-	NeedsUpdate bool
-	PMMAgentID  string
+	ServiceID     string
+	WaitNextCheck time.Duration
```
BupycHuk, line 66: I meant `NeedsUpdate` :)

BupycHuk (approved):

PR #811, opened 2021-08-03
PMM-8269 MongoDB Point-in-Time backups 

artemgavrilov:
[PMM-8269](https://jira.percona.com/browse/PMM-8269)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2001
---
- [ ] https://github.com/percona/pmm/pull/769
- [ ] https://github.com/percona/pmm-agent/pull/285
---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.


models/database.go
```diff
@@ -549,6 +549,10 @@ var databaseSchema = [][]string{
 		`ALTER TABLE agents
 		ADD COLUMN agent_password VARCHAR CHECK (agent_password <> '')`,
 	},
+	43: {
+		`ALTER TABLE artifacts
+      ADD COLUMN mode VARCHAR NOT NULL CHECK (mode <> '') DEFAULT 'snapshot'`,
```
Dasio, line 554: below it's said that we should avoid default values, so let's drop it after?

services/backup/backup_service.go
```diff
@@ -212,6 +241,40 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
...
+	if len(agents) == 0 {
+		return errors.Errorf("cannot find pmm agent for service %s", serviceID)
+	}
+	pmmAgentID := agents[0].AgentID
+	res, err := models.CreateActionResult(s.db.Querier, pmmAgentID)
+	if err != nil {
+		return err
+	}
+
+	DSN, agent, err := models.FindDSNByServiceIDandPMMAgentID(s.db.Querier, serviceID, pmmAgentID, "")
```
Dasio, line 263: wrap it to transaction?

models/database.go
artemgavrilov: Yep, added.

services/backup/backup_service.go
artemgavrilov: Yep, done

services/scheduler/scheduler.go
```diff
@@ -81,9 +85,12 @@ type AddParams struct {
 // Add adds task to scheduler and save it to DB.
 func (s *Service) Add(task Task, params AddParams) (*models.ScheduledTask, error) {
 	var scheduledTask *models.ScheduledTask
-	var err error
+	errTx := s.db.InTransactionContext(s.db.Querier.Context(), &sql.TxOptions{Isolation: sql.LevelSerializable}, func(tx *reform.TX) error {
```
artemgavrilov, line 88: Important stuff! This transaction is correct only on a serializable level. We need it because of `checkPreconditions` function, where we are checking if there other enabled scheduled backups for that service.

So simplified tx is:
```
// Chech existing tasks
SELECT * FROM scheduled_tasks WHERE ; 
// If the result is empty then we can enable a new task;
INSERT INTO tasks …;
```

On other isolation levels it's possible to get two enable PITR tasks.

artemgavrilov:
I will add more tests soon, but in general, it's ready for review.

services/backup/backup_service.go
```diff
@@ -226,6 +289,58 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
 	return restoreID, nil
 }
 
+// SwitchMongoPITR switches Point-in-Time recovery feature for mongoDB with given serviceID.
+func (s *Service) SwitchMongoPITR(ctx context.Context, serviceID string, enabled bool) error {
+	var res *models.ActionResult
+	var pmmAgentID, DSN string
```
oter, line 295: minor: `DSN` should be `dsn`

services/backup/backup_service.go
```diff
@@ -226,6 +289,58 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
...
+	var agent *models.Agent
+	var service *models.Service
+
+	errTX := s.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
+		var err error
+		service, err = models.FindServiceByID(s.db.Querier, serviceID)
```
oter, line 301: should use `tx` in the transactions.

services/backup/backup_service.go
```diff
@@ -226,6 +289,58 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
 	return restoreID, nil
 }
 
+// SwitchMongoPITR switches Point-in-Time recovery feature for mongoDB with given serviceID.
+func (s *Service) SwitchMongoPITR(ctx context.Context, serviceID string, enabled bool) error {
```
oter, line 293: is this func idempotent?

services/management/backup/backups_service.go
```diff
@@ -137,8 +156,16 @@ func (s *BackupsService) ScheduleBackup(ctx context.Context, req *backupv1beta1.
 		var task scheduler.Task
 		switch svc.ServiceType {
 		case models.MySQLServiceType:
+			if req.Mode != backupv1beta1.BackupMode_SNAPSHOT {
+				return status.Errorf(codes.Unimplemented, "unimplemented backup mode for mySQL: %s", req.Mode.String())
```
oter, line 160: maybe `MySQL`? also for the `mongoDB`

services/scheduler/scheduler.go
```diff
@@ -95,7 +102,7 @@ func (s *Service) Add(task Task, params AddParams) (*models.ScheduledTask, error
 			return err
 		}
 
-		if err := s.addDBTask(scheduledTask); err != nil {
+		if err = s.addDBTask(scheduledTask); err != nil {
```
oter, line 105: addDBTask call shouldn't be in the transaction since it doesn't accept tx.

oter (changes requested):

services/scheduler/scheduler.go
artemgavrilov: `addDBTask` doesn't use DB at all, it changes the in-memory state.  There is rollback functionality for this state here :https://github.com/percona/pmm-managed/blob/9a5d465d7cdccc48f1403ef916babce209deb1a6/services/scheduler/scheduler.go#L112-L118 but most likely it will not work properly. 

Anyway, all this stuff is not part of the current ticket, so it should be refactored in different ticket/pr

services/backup/backup_service.go
artemgavrilov: Nope, it switches PITR option on the target service node.

services/backup/backup_service.go
artemgavrilov: done

services/backup/backup_service.go
artemgavrilov: 🤦‍♂️🤦‍♂️🤦‍♂️🤦‍♂️ Done

services/management/backup/backups_service.go
artemgavrilov: Done

models/artifact_helpers.go
```diff
@@ -137,6 +137,23 @@ func FindArtifactByID(q *reform.Querier, id string) (*Artifact, error) {
...
+	if name == "" {
+		return nil, errors.New("provided artifact name is empty")
+	}
+	artifact := new(Artifact)
+	err := q.FindOneTo(artifact, "name", name)
+	switch err {
+	case nil:
+		return artifact, nil
+	case reform.ErrNoRows:
+		return nil, status.Errorf(codes.NotFound, "Artifact with name %q not found.", name)
```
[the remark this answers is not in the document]
artemgavrilov: Done

Dasio (approved):

models/database.go
```diff
@@ -601,6 +601,13 @@ var databaseSchema = [][]string{
 		`ALTER TABLE artifacts
       ADD COLUMN mode VARCHAR NOT NULL CHECK (mode <> '') DEFAULT 'snapshot'`,
 		`ALTER TABLE artifacts ALTER COLUMN mode DROP DEFAULT`,
+		`UPDATE scheduled_tasks set data = jsonb_set(data::jsonb, '{mysql_backup, data_model}', '"physical"') WHERE type = 'mongodb_backup'`,
```
Dasio, line 604: WHERE type = 'mongodb_backup'` ?

oter (approved):

models/database.go
artemgavrilov: Nice catch, fixed.

services/scheduler/scheduler.go
BupycHuk: so, let's add it as comment

BupycHuk (commented):
Can we run several pbm instances on the same node for different services?

models/artifact_helpers.go
```diff
@@ -137,6 +137,23 @@ func FindArtifactByID(q *reform.Querier, id string) (*Artifact, error) {
 	}
 }
 
+// FindArtifactByName returns artifact by given name if found, ErrNotFound if not.
+func FindArtifactByName(q *reform.Querier, name string) (*Artifact, error) {
+	if name == "" {
+		return nil, errors.New("provided artifact name is empty")
```
atymchuk, line 143:
```suggestion
		return nil, errors.New("provided backup artifact name is empty")
```

models/artifact_helpers.go
```diff
@@ -137,6 +137,23 @@ func FindArtifactByID(q *reform.Querier, id string) (*Artifact, error) {
...
+	}
+	artifact := new(Artifact)
+	err := q.FindOneTo(artifact, "name", name)
+	switch err {
+	case nil:
+		return artifact, nil
+	case reform.ErrNoRows:
+		return nil, errors.Wrapf(ErrNotFound, "artifact with name %q not found.", name)
```
atymchuk, line 151:
```suggestion
		return nil, errors.Wrapf(ErrNotFound, "backup artifact with name %q not found.", name)
```

services/backup/backup_service.go
```diff
@@ -142,6 +185,26 @@ func (s *Service) PerformBackup(ctx context.Context, params PerformBackupParams)
...
+		ServiceID: service.ServiceID,
+		Mode:      models.PITR,
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, task := range tasks {
+		if task.ID != scheduleID {
+			return status.Errorf(codes.FailedPrecondition, "Can't perform backup because service %s has configured scheduled "+
```
atymchuk, line 200: I suggest to turn it to this:
"Can't make a backup because service %s already contains scheduled PITR backups. Please disable them if you want to make another backup."

services/backup/backup_service.go
```diff
@@ -226,6 +289,58 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
...
+
+	errTX := s.db.InTransactionContext(ctx, nil, func(tx *reform.TX) error {
+		var err error
+		service, err = models.FindServiceByID(tx.Querier, serviceID)
+		if err != nil {
+			return err
+		}
+
+		if service.ServiceType != models.MongoDBServiceType {
+			return errors.Errorf("Point-in-Time recovery feature available only for mongoDB services,"+
```
atymchuk, line 307:
```suggestion
			return errors.Errorf("Point-in-Time recovery feature is only available for mongoDB services,"+
```

services/backup/backup_service.go
```diff
@@ -226,6 +289,58 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
...
+				"given service id: %s, service type: %s", serviceID, service.ServiceType)
```
atymchuk, line 308:
```suggestion
				"current service id: %s, service type: %s", serviceID, service.ServiceType)
```

services/scheduler/scheduler.go
```diff
@@ -298,32 +308,92 @@ func (s *Service) convertDBTask(dbTask *models.ScheduledTask) (Task, error) {
...
+			Disabled:  pointer.ToBool(false),
+			ServiceID: serviceID,
+		})
+		if err != nil {
+			return err
+		}
+
+		for _, task := range tasks {
+			if task.ID != scheduleID {
+				return status.Error(codes.FailedPrecondition, "Scheduled PITR backup can be enabled only if there is no other scheduled backups enabled.")
```
atymchuk, line 379: Enabling can be confused with `toggling something on/off`. I suggest we use the word `create` or `register`, i.e.
```suggestion
				return status.Error(codes.FailedPrecondition, "A scheduled PITR backup can be created only if there are no other scheduled backups.")
```

services/scheduler/scheduler.go
```diff
@@ -298,32 +308,92 @@ func (s *Service) convertDBTask(dbTask *models.ScheduledTask) (Task, error) {
...
+			Disabled:  pointer.ToBool(false),
+			ServiceID: serviceID,
+			Mode:      models.PITR,
+		})
+		if err != nil {
+			return err
+		}
+
+		if len(tasks) != 0 {
+			return status.Error(codes.FailedPrecondition, "Scheduled snapshot backup can be enabled only if there is no enabled PITR backup.")
```
atymchuk, line 394:
```suggestion
			return status.Error(codes.FailedPrecondition, "A scheduled snapshot backup can be created only if there are no other PITR backups.")
```

atymchuk (changes requested):

services/scheduler/scheduler.go
artemgavrilov: Done

models/artifact_helpers.go
artemgavrilov: Done

services/backup/backup_service.go
artemgavrilov: Done

services/scheduler/scheduler.go
artemgavrilov: Done partially. This check is about enabling, because user can create multiple disabled PITR backups

services/backup/backup_service.go
artemgavrilov: Done

services/scheduler/scheduler.go
artemgavrilov: Done partially.
Same, users can have many PITR backups if they are disabled.

models/artifact_helpers.go
artemgavrilov: Done

services/backup/backup_service.go
artemgavrilov: Done partially, a left `has` instead of `contains`. IMHO this word fits better.

artemgavrilov:
> Can we run several pbm instances on the same node for different services?

@BupycHuk `pbm-agent` should be started on the same node with MongoDB, but `pbm` itself is CLI tool and we can run it from anywhere. But right now we can't track parallel `pbm` invocations because they affect cluster, not a single service.

I just realized that user can harm pbm state right now because he can start two different backups for two different services, but members of one cluster. So this problem should be solved by adding cluster concept, and allowing to start mongo backups only for clusters.

atymchuk (approved):

oter (approved):

PR #825, opened 2021-09-14
PMM-8706: Add specific error codes in case of incompatible service

oter:
[PMM-8706](https://jira.percona.com/browse/PMM-8706)

API PR: https://github.com/percona/pmm/pull/777

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2016

---
- [x] Tests passed.
- [x] Feature build pass.
- [x] (Re)requested review.
- [x] Update API dependency.


Dasio (approved):

services/backup/backup_service_test.go
```diff
@@ -97,25 +98,198 @@ func TestStartBackup(t *testing.T) {
...
+		}
+		mockedVersioner.On("GetVersions", *agent.PMMAgentID, softwares).Return(versions1, nil).Once()
+		artifactID, err := backupService.PerformBackup(ctx, PerformBackupParams{
+			ServiceID:  pointer.GetString(agent.ServiceID),
+			LocationID: locationRes.ID,
+			Name:       "test_backup",
+		})
+		assert.NoError(t, err)
+
+		assert.NoError(t, err)
```
artemgavrilov, line 161: Duplication, also let's use `require` for errors.

services/management/backup/backups_service.go
```diff
@@ -83,14 +112,45 @@ func (s *BackupsService) StartBackup(ctx context.Context, req *backupv1beta1.Sta
...
-		return nil, err
+		return nil, convertBackupError(err)
 	}
 
 	return &backupv1beta1.StartBackupResponse{
 		ArtifactId: artifactID,
 	}, nil
 }
 
+func convertRestoreBackupError(restoreError error) error {
```
artemgavrilov, line 123: Can't we merge this func and `convertBackupError` into one?

oter: In my opinion these 2 convert functions are different, it is better to not have just one convert function that covers all the cases, because we will not be able to
 - cover in the tests the cases that is not used in the specific function
 - know what errors the function could return

services/backup/backup_service.go
```diff
@@ -278,8 +292,8 @@ func (s *Service) RestoreBackup(ctx context.Context, serviceID, artifactID strin
 
 		if params.ServiceType == models.MySQLServiceType && params.DBVersion != "" {
 			if params.DBVersion != dbVersion {
-				return errors.Errorf("incompatible service: artifact db version %q != db version %q",
-					params.DBVersion, dbVersion)
+				return fmt.Errorf("%w: incompatible service: artifact db version %q != db version %q",
```
BupycHuk, line 295: maybe it's better to use `errors.Wrap` ?
Error message here will be like `incompatible service: incompatible service: artifact db version %q != db version %q`, so it's better to not  duplicate that we already have in wrapped error.

BupycHuk: Here and in other places

BupycHuk (changes requested):

BupycHuk (approved):

services/backup/backup_service.go
oter: done

PR #856, opened 2021-10-29
PMM-8982: Migrate unfinished logs

oter:
[PMM-8982](https://jira.percona.com/browse/PMM-8982)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2119

---
- [x] Tests passed.
- [x] Feature build pass.
- [x] (Re)requested review.


BupycHuk (approved):
looks good to me, what about cases when pmm-agent has outdated version?

oter:
> looks good to me, what about cases when pmm-agent has outdated version?

The outdated pmm-agent should not affect the bug I fixed. My fix is about the logs functionality implementation which is not present in the previous version of pmm-managed and added in the following version, and we forgot to mark old jobs that they have no logs at all.

PR #860, opened 2021-11-09
Increase github jobs timeout

oter:
Jobs started to fail constantly because of jobs execution timeout, so the quick fix is to increase the timeout to 30m.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ on:
 jobs:
   test:
     name: Test
-    timeout-minutes: 20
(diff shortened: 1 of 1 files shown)


BupycHuk (approved):

artemgavrilov (approved):
But that's not ok that our CI requires > 20 minutes.

oter:
> But that's not ok that our CI requires > 20 minutes.

yes, agree, that's why I filed a ticket - https://jira.percona.com/browse/PMM-9128 

Some tests could be run in parallel, and some steps probably could be improved.

PR #862, opened 2021-11-09
PMM-9063 SSO

JiriCtvrtka:
[PMM-9063](https://jira.percona.com/browse/PMM-9063)

---
- [x] Tests passed.
- [ ] Feature build pass.
- [x] (Re)requested review.
- [ ] Update API dependency.


models/percona_sso_model_helpers.go
```diff
@@ -17,17 +17,104 @@
...
+			return nil, err
+		}
+		if err := json.Unmarshal(bodyBytes, &accessToken); err != nil {
+			return nil, err
+		}
+
+		accessToken.ExpiresAt = timeBeforeRequest.Add(time.Duration(accessToken.ExpiresIn) * time.Second)
+		sso.AccessToken = accessToken
+
+		if err := q.Insert(sso); err != nil {
```
jprukner, line 102: As we talked on Slack, it's not clear this function inserts anything.
Actually what about updating? Do we insert a new row if the token is expired?

After thinking about it I would suggest moving the insert out of this function into `InsertPerconaSSODetails` and using here `q.UpdateColumn(...)` to update the access token when we get it. What do you think? @JiriCtvrtka

jprukner (changes requested):

models/percona_sso_model_helpers.go
JiriCtvrtka: Fixed.

models/percona_sso_model_helpers_test.go
```diff
@@ -38,26 +40,62 @@ func setupDB(t *testing.T) (*reform.DB, func()) {
...
+		ssoDetails, err = models.GetPerconaSSODetails(ctx, db.Querier)
+		assert.Error(t, err)
+		assert.Nil(t, ssoDetails)
+		// See https://github.com/percona/pmm-managed/pull/852#discussion_r738178192
+		ssoDetails, err = models.GetPerconaSSODetails(ctx, db.Querier)
+		assert.Error(t, err)
+		assert.Nil(t, ssoDetails)
+	})
+
+	t.Run("WrongCredentials", func(t *testing.T) {
```
jprukner, line 86: Is this wrong credentials or wrong issuer URL?

models/percona_sso_model_helpers_test.go
```diff
@@ -38,26 +40,62 @@ func setupDB(t *testing.T) (*reform.DB, func()) {
...
+		}
+		err := models.InsertPerconaSSODetails(ctx, db.Querier, InsertSSODetails)
+		require.NoError(t, err)
+		ssoDetails, err := models.GetPerconaSSODetails(ctx, db.Querier)
+		require.NoError(t, err)
+		assert.NotNil(t, ssoDetails)
+		assert.Equal(t, expectedSSODetails.ClientID, ssoDetails.ClientID)
+		assert.Equal(t, expectedSSODetails.ClientSecret, ssoDetails.ClientSecret)
+		assert.Equal(t, expectedSSODetails.IssuerURL, ssoDetails.IssuerURL)
+		assert.Equal(t, expectedSSODetails.Scope, ssoDetails.Scope)
```
jprukner, line 74: Can we check that access_token has been fetched?

jprukner (changes requested):

models/percona_sso_model_helpers.go
```diff
@@ -17,32 +17,128 @@
...
+	h.Add("Accept", "application/json")
+	h.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	timeBeforeRequest := time.Now()
+	client := &http.Client{}
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
JiriCtvrtka: Done.

models/percona_sso_model_helpers_test.go
JiriCtvrtka: It is wrong credentials. Wrong client ID and client secret.

models/percona_sso_model_helpers_test.go
JiriCtvrtka: Check added that access token and properties are not nil.

models/percona_sso_model_helpers.go
```diff
@@ -17,32 +17,128 @@
...
+			return nil, errors.Wrap(err, "failed to insert Percona SSO Details")
+		}
+		details.AccessToken = refreshedToken
+	}
+
+	return details, nil
+}
+
+// GetPerconaSSOAccessToken returns PerconaSSOAccessToken if there is GetPerconaSSODetails, error otherwise.
+func GetPerconaSSOAccessToken(ctx context.Context, q *reform.Querier) (*PerconaSSOAccessToken, error) {
```
palash25, line 60: why do we have a separate getter method if `ssoDetails.AccessToken ` is an exported field?

jprukner: I would suggest to return string only, so this method has a purpose. Without it, we'd have to write `ssoDetails.AccessToken.AccessToken` to get the string token we actually need. What do you guys think?

palash25: but even just the string `AccessToken` is an exported field, if we want a method to get this value then we should make these fields unexported don;t you think?

jprukner: Well then we can't use reform. Because it can not insert into unexported field.

jprukner: Or use some proxy struct to fetch it.

models/percona_sso_model.go
```diff
@@ -17,33 +17,67 @@
...
 func (s *PerconaSSODetails) BeforeInsert() error {
 	now := Now()
 	s.CreatedAt = now.UTC()
 	return nil
 }
 
+// BeforeUpdate implements reform.BeforeUpdater interface.
+func (s *PerconaSSODetails) BeforeUpdate() error {
+	now := Now()
+	s.CreatedAt = now.UTC()
```
maxkondr, line 75:
```suggestion
	s.CreatedAt = Now().UTC()
```

models/percona_sso_model_helpers.go
```diff
@@ -17,32 +17,128 @@
...
+
+	return nil, fmt.Errorf("get access token of Percona SSO Details failed, status code: %d", res.StatusCode)
+}
+
+func (sso *PerconaSSODetails) isAccessTokenExpired() bool {
+	if sso == nil || sso.AccessToken == nil {
+		return true
+	}
+
+	return time.Now().After(sso.AccessToken.ExpiresAt)
```
maxkondr, line 118: I would add one more condition:
```suggestion
	return time.Now().After(sso.AccessToken.ExpiresAt.Add(-Minute * 5))
```

to reserve some window to perform API call to Portal (and verify this token on Portal side).

So that the logic is the following: refresh access token BEFORE it actually expires.

models/percona_sso_model_helpers.go
```diff
@@ -17,32 +17,128 @@
...
+	ssoDetails, err := GetPerconaSSODetails(ctx, q)
+	if err != nil {
+		return nil, err
+	}
+
+	return ssoDetails.AccessToken, nil
+}
+
+func (sso *PerconaSSODetails) refreshAndGetAccessToken(ctx context.Context, q *reform.Querier) (*PerconaSSOAccessToken, error) {
+	url := fmt.Sprintf("%s%s", sso.IssuerURL, sso.Scope)
```
maxkondr, line 70: 1) this is not enough, please check https://confluence.percona.com/display/PMM/Single+Sign-On+-+PMM+Server+integration+with+Portal+and+Okta#SingleSignOnPMMServerintegrationwithPortalandOkta-ObtainandUseAccessToken

```
curl -X POST '<ISSUER URL>/token?grant_type=client_credentials&scope=<portal_auth_scope>' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'accept: application/json' \
--header 'Authorization: Basic base64(<portal_auth_client_id:portal_auth_client_secret>)'
```

`/token` and `grant_type=client_credentials` are missed.

2) use encoded url params

models/percona_sso_model_helpers.go
```diff
@@ -17,32 +17,128 @@
...
+		sso.AccessToken = accessToken
+
+		if err := q.UpdateColumns(sso, "access_token"); err != nil {
+			return nil, err
+		}
+
+		return accessToken, nil
+	}
+
+	return nil, fmt.Errorf("get access token of Percona SSO Details failed, status code: %d", res.StatusCode)
```
maxkondr, line 110: I would add error log output with response body - it contains the reason of reject

maxkondr (changes requested):

models/percona_sso_model_helpers.go
JiriCtvrtka: Removed whole separated method to get access token.

jprukner (approved):

palash25 (approved):

percona-csalguero (approved):

palash25 (approved):

maxkondr (approved):

PR #863, opened 2021-11-09
PMM-9125 Enable stream parse only for postgres_exporter.

BupycHuk:
[PMM-9125](https://jira.percona.com/browse/PMM-9125)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2128

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	github.com/percona-platform/dbaas-api v0.0.0-20211006103141-a68316e565d2
 	github.com/percona-platform/saas v0.0.0-20211101203847-f65c32bc8770
 	github.com/percona/pmm v0.0.0-20211103155323-14c4ca50e3da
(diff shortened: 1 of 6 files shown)


percona-csalguero (approved):

artemgavrilov (approved):
From the code point of view LGTM. But I do not fully understand fix in general. Why do we leave this option for PG, how does it affects other exporters, etc? Also, what is the root cause of that ticket at all?

BupycHuk:
@artemgavrilov memory usage increased for postgres exporter last month because of thousands of metrics generated by postgres_exporter and we fixed it by enabling stream parse. But this fix increased memory usage for all other exporters. That's why we keep this option for postgres exporter and remove for other exporters.

PR #867, opened 2021-11-16
PMM-9155 fix failing API tests

ritbl:
[PMM-9155](https://jira.percona.com/browse/PMM-9155)

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2144

---
- [x] Tests passed.
- [ ] Feature build pass.
- [x] (Re)requested review.
- [x] Update API dependency.

I ran all tests locally against devcontainer, all green.

### DIFF
--- a/api-tests/inventory/services_test.go
+++ b/api-tests/inventory/services_test.go
@@ -733,6 +733,8 @@ func TestMongoDBService(t *testing.T) {
 }
 
 func TestPostgreSQLService(t *testing.T) {
+	const defaultPostgresDBName = "postgres"
+
 	t.Run("Basic", func(t *testing.T) {
 		t.Parallel()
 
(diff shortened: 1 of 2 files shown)


api-tests/inventory/services_test.go
```diff
@@ -757,11 +759,12 @@ func TestPostgreSQLService(t *testing.T) {
...
 				Postgresql: &services.AddPostgreSQLServiceOKBodyPostgresql{
-					ServiceID:   serviceID,
-					NodeID:      genericNodeID,
-					Address:     "localhost",
-					Port:        5432,
-					ServiceName: serviceName,
+					ServiceID:    serviceID,
+					NodeID:       genericNodeID,
+					Address:      "localhost",
+					DatabaseName: defaultPostgresDBName,
```
BupycHuk, line 765: what about services that were added before these changes, will they return `defaultPostgresDBName`?

ritbl: good question. 

TL;DR;
empty string will be returned for previously added databases

Long version
ToAPIService (pmm-managed/services/converters.go) which is called from API, will [return](https://github.com/percona/pmm-managed/pull/845/files#diff-824e770879964798b306ebce4e2a780a2bccea6bfa56cdcdbe73cdfc5f42f833R144) return whatever is stored in services.database_name column, for the existing endpoints it will [hold empty string](https://github.com/percona/pmm-managed/pull/845/files#diff-23d49d8e5032dd24381d8eba1389cb7c696fb893b73c9069c0435d5528661dc3R655).
In the API it is optional, so it is [backwards compatible](https://github.com/percona/pmm-managed/pull/845/files#diff-a87a12cc58d36aa06d6a438685288dc80e4dfebf6cca8c1ea00ba310575e91a7R241).

BupycHuk: TBH it brings inconsistency in API response, so if we change default value then we should do it for old records in DB (add migration) or not set default value for new records.

ritbl: I see your point. I will provide another PR with DB migration for pg services.

BupycHuk: let's not create one more PR and fix it here.

PR #877, opened 2021-12-01
PMM-9228 Disconnect PMM from Portal

JiriCtvrtka:
**Diff coverage will be cover with API tests.**

[PMM-9228](https://jira.percona.com/browse/PMM-9228)

- [x] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2205

PMM: https://github.com/percona/pmm/pull/802
---
- [x] Tests passed.
- [ ] Feature build pass.
- [x] (Re)requested review.
- [ ] Update API dependency.


services/platform/platform.go
```diff
@@ -111,16 +115,53 @@ func (s *Service) Connect(ctx context.Context, req *platformpb.ConnectRequest) (
...
+
+	err = s.disconnect(nCtx, &disconnectPMMParams{
+		PMMServerID: settings.PMMServerID,
+		AccessToken: ssoDetails.AccessToken.AccessToken,
+	})
+	if err != nil {
+		return nil, err // this is already a status error
+	}
+
+	err = models.DeletePerconaSSODetails(s.db.Querier)
```
jprukner, line 152: What do you think about reversing the order? If the deletion fails here, we can't reconnect.
If we delete the sso and then disconnect. When disconnect fails, we can rollback SSO details deleting.

go.sum
```diff
@@ -455,8 +455,10 @@ github.com/percona-platform/dbaas-api v0.0.0-20211201151251-014259873599 h1:dDlc
 github.com/percona-platform/dbaas-api v0.0.0-20211201151251-014259873599/go.mod h1:zgb9gTJusc8Jv2zRNAlWV0/XQ8IK+hwHMc9lIqfK0tM=
 github.com/percona-platform/saas v0.0.0-20211101203847-f65c32bc8770 h1:ZfS8TnnctQn0ZMB2YV7Q46LbN9odamb/nsyBJWFNaRY=
 github.com/percona-platform/saas v0.0.0-20211101203847-f65c32bc8770/go.mod h1:jJRyGMxxDJaSiU7AaHNS+8j1TFQQhX6lcYp8s0t8Knc=
-github.com/percona/pmm v0.0.0-20211203124433-9a6e4ff18a5a h1:KwQbiiRgZNx1fJQBZcpnzdppmFQeev8cip8gC0ImiA8=
-github.com/percona/pmm v0.0.0-20211203124433-9a6e4ff18a5a/go.mod h1:OmWayvQAavtvlzLkvpea5tAqaWGGNNyG+xj4MJUsNm4=
+github.com/percona/pmm v0.0.0-20211202192401-a3cf14b2ce6b h1:nN9YtzlQAnCKlaFDoOvvTdHquFhQi3VfV/yhZ3MFibI=
+github.com/percona/pmm v0.0.0-20211202192401-a3cf14b2ce6b/go.mod h1:OmWayvQAavtvlzLkvpea5tAqaWGGNNyG+xj4MJUsNm4=
+github.com/percona/pmm v0.0.0-20211206095108-ce2e6a6ab00d h1:SfZ1UiYkeiFpvLqP2tlYSZj8Cc1rXI70P9yLrUpri/0=
+github.com/percona/pmm v0.0.0-20211206095108-ce2e6a6ab00d/go.mod h1:OmWayvQAavtvlzLkvpea5tAqaWGGNNyG+xj4MJUsNm4=
```
oksana-grishchenko, line 461: seems like duplication here, try `go mod tidy`

JiriCtvrtka:
@oksana-grishchenko @jprukner Ok I will reverse order, do we want do this also for connect? Because right now it has same order like disconnect.

services/platform/platform.go
oksana-grishchenko: > @oksana-grishchenko @jprukner Ok I will reverse order, do we want do this also for connect? Because right now it has same order like disconnect.

Seems like `connect` doesn't have any restore actions. In case of failure it only returns an error and log it, so I guess the order is less matter in that case. 
I just would like to highlight that in the ideal world the consistent state should be restored and in that case the order matters. I don't know if this restore process after failed `disconnect` (as well as `connect`) is in the scope of the current task.

jprukner (approved):
Not sure if tx.Rollback(), tx.Commit would simpler for the deleting SSO details. But this works, too.

JiriCtvrtka:
For Tx I need pass db, in this case it means rewrite several methods. So in case of fail insert it again seems easier here.

services/platform/platform.go
```diff
@@ -111,16 +116,81 @@ func (s *Service) Connect(ctx context.Context, req *platformpb.ConnectRequest) (
...
+			s.l.Errorf("%s %s", rollbackFailed, e)
+		}
+		if e := models.InsertPerconaSSODetails(s.db.Querier, &models.PerconaSSODetailsInsert{
+			ClientID:     ssoDetails.ClientID,
+			ClientSecret: ssoDetails.ClientSecret,
+			IssuerURL:    ssoDetails.IssuerURL,
+			Scope:        ssoDetails.Scope,
+		}); e != nil {
+			s.l.Errorf("%s %s", rollbackFailed, e)
+		}
```
oksana-grishchenko, line 174: Just wondering, why inserting again instead of use a transaction and rollback it in case of fail? My concern is that in a general case another insert may lead to the changed internal ids, which means a new record instead of the old one, so it may break some logic that counting on the old ids. 
It may be not the case here, I'm just curious what do you think about it.

oksana-grishchenko (approved):

services/platform/platform.go
JiriCtvrtka: Another insert into SSO table doesnt break anything since we dont using incremental ID here, content of table will be same. About TX itseft then we need pass DB instead Querier and it require several methods to rewrite. Also not all steps of disconnect are in DB. So it will only affect only one from 3 steps.

services/platform/platform.go
```diff
@@ -110,16 +115,78 @@ func (s *Service) Connect(ctx context.Context, req *platformpb.ConnectRequest) (
...
+		return nil, status.Error(codes.Aborted, "PMM server is not connected to Portal")
+	}
+
+	settings, err := models.GetSettings(s.db)
+	if err != nil {
+		s.l.Errorf("Failed to fetch PMM server ID and address: %s", err)
+		return nil, status.Error(codes.Internal, internalServerError)
+	}
+
+	if err := s.CleanSupervisordConfigurations(ctx); err != nil {
```
palash25, line 141: i don't see the need for a separate method for this, `CleanSupervisordConfiguration` calls `models.GetSettings` (which has already been called above)  and then `s.supervisord.UpdateConfiguration(settings, nil)`  why not just call `s.supervisord.UpdateConfiguration` here instead of a having a separate method, i don't see it being reused anywhere else.

services/platform/platform.go
```diff
@@ -174,37 +246,69 @@ func (s *Service) connect(ctx context.Context, params *connectPMMParams) (*ssoDe
...
 		s.l.Errorf("Failed to decode response into SSO details: %s", err)
-		return nil, status.Error(codes.Internal, "Internal server error")
+		return nil, status.Error(codes.Internal, internalServerError)
 	}
 	return response.SSODetails, nil
 }
+
+func (s *Service) disconnect(ctx context.Context, params *disconnectPMMParams) error {
+	endpoint := fmt.Sprintf("https://%s/v1/orgs/inventory/%s:disconnect", s.host, params.PMMServerID)
+	client := http.Client{Timeout: s.platformAPITimeout}
```
palash25, line 286: this is the same for all requests, i think we can make this a field of Service struct and reuse everywhere. what do you think?

services/platform/platform.go
JiriCtvrtka: Yes right.

services/platform/platform.go
JiriCtvrtka: Sounds good, changed for both requests.

services/platform/platform.go
```diff
@@ -174,37 +233,67 @@ func (s *Service) connect(ctx context.Context, params *connectPMMParams) (*ssoDe
...
 		s.l.Errorf("Failed to decode response into SSO details: %s", err)
-		return nil, status.Error(codes.Internal, "Internal server error")
+		return nil, status.Error(codes.Internal, internalServerError)
 	}
 	return response.SSODetails, nil
 }
+
+func (s *Service) disconnect(ctx context.Context, params *disconnectPMMParams) error {
+	endpoint := fmt.Sprintf("https://%s/v1/orgs/inventory/%s:disconnect", s.host, params.PMMServerID)
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, endpoint, nil)
```
oksana-grishchenko, line 272: Since this API method updated to be custom, it uses POST instead of DELETE

JiriCtvrtka: Updated.

services/platform/platform.go
jprukner: Should not we first insert and then update configuration?

JiriCtvrtka: Order changed.

jprukner (approved):

oksana-grishchenko (approved):

palash25 (approved):

maxkondr (approved):

services/grafana/client.go
```diff
@@ -508,6 +508,39 @@ func (c *Client) IsReady(ctx context.Context) error {
...
 
+const grpcGatewayCookie = "grpcgateway-cookie"
+
+type currentUser struct {
+	AccessToken string `json:"access_token"`
+}
+
+var errCookieIsNotSet = errors.Errorf("cookie %q is not set", grpcGatewayCookie)
+
+func (c *Client) GetCurrentUserAccessToken(ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
JiriCtvrtka: Right, thanks.

services/platform/platform.go
```diff
@@ -34,9 +33,15 @@ import (
 	"github.com/percona/pmm/api/platformpb"
 
 	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/services"
 	"github.com/percona/pmm-managed/utils/envvars"
 )
 
+const (
+	internalServerError = "Internal server error"
```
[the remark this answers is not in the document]
JiriCtvrtka: Changed.

PR #881, opened 2021-12-03
PMM-9050 parseSAASHost: avoid including the port if it's passed

jprukner:
[PMM-9050](https://jira.percona.com/browse/PMM-9050)

This is a fix for a bug that makes pmm-managed crash when port is part of the SAAS host. We now can handle such state. Parsing out the host only.

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2178

---
- [ ] Tests passed.
- [ ] Feature build pass.
- [ ] (Re)requested review.
- [ ] Update API dependency.

### DIFF
--- a/utils/envvars/parser.go
+++ b/utils/envvars/parser.go
@@ -19,6 +19,7 @@ package envvars
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


percona-csalguero (approved):

utils/envvars/parser.go
```diff
@@ -218,10 +219,18 @@ func parseSAASHost(v string) (string, error) {
 		logrus.Infof("Using default SaaS host %q.", defaultSaaSHost)
 		return defaultSaaSHost, nil
 	}
-	if strings.Contains(v, ":") {
-		return "", errors.New("saas host can't have port set: it's signed for domain without the port")
+	if strings.HasPrefix(v, ":") {
```
artemgavrilov, line 222: What is the point to accept `host:port` format if we throw away `port` part?

jprukner: Because it's everywhere in tests with ports. Previously we failed to start pmm-managed in those scenarios.

artemgavrilov: In that case, we should modify our tests and use strict validation here. It may be really confusing for someone it the future when you set port but it's ignored.

Can you please create ticket for that?

PR #972, opened 2022-01-03
PMM-9400 Added indexstats list

percona-csalguero:
[PMM-9400](https://jira.percona.com/browse/PMM-9400) Missing metrics for indexStats and collStats when all collectors enabled

In order to make it work correctly, we need to pass `--mongodb.indexstats-colls=` also, not only `--mongodb.collstats-colls=`

- [ ] Build: https://github.com/Percona-Lab/pmm-submodules/pull/2265

- [ ] Tests passed.
- [ ] Feature build pass.

PR #975, opened 2022-01-10
Pmm 9400 cherry pick on 2.26

percona-csalguero:


denisok (changes requested):
it doesn't looks like  https://github.com/percona/pmm-managed/pull/972/files
you probably get some more, like 1.7 go as well

PR #982, opened 2022-01-20
PMM-9413 Fix deadlock in agent addition/removal logic

ShashankSinha252:
[PMM-9413](https://jira.percona.com/browse/PMM-9413) : In PMM server, locks are used to update list of agents. Agent addition or removal is performed after lock is acquired. Due to programming error, we hit a scenario where we perform removal of an agent while adding another agent. This leads to deadlock, as goroutine waits to acquire a lock that it already holds.

This change fixes the bug. We now acquire lock for agent addition only after agent removal is performed.

Related builds :
- https://github.com/Percona-Lab/pmm-submodules/pull/2291


oksana-grishchenko (approved):
So since under the hood the `Kick` method also calls this `r.rw.Lock()` it led to a deadlock when there was an already registered agent with the same ID. 

Now we use the concurrent safe way to read - `r.IsConnected`, the method `r.Kick` is also concurrent safe, so the `r.rw.Lock()` can be called later and not lead to a deadlock. 

Looks good

JiriCtvrtka (approved):

BupycHuk (approved):

PR #988, opened 2022-01-30
PMM-8308 track process exec path

qwest812:
https://jira.percona.com/browse/PMM-8308
fb https://github.com/Percona-Lab/pmm-submodules/pull/2307
pmm https://github.com/percona/pmm/pull/821
pmm-admin https://github.com/percona/pmm-admin/pull/176
pmm-agent  https://github.com/percona/pmm-agent/pull/314


docker-compose.yml
```diff
@@ -34,6 +34,7 @@ services:
       - 15432:5432
     volumes:
       - .:/root/go/src/github.com/percona/pmm-managed
+      - ../pmm:/root/go/src/github.com/percona/pmm
```
BupycHuk, line 37: let's revert it, it might bring new problems.

models/database.go
```diff
@@ -683,6 +683,10 @@ var databaseSchema = [][]string{
 		`ALTER TABLE ia_templates
 			DROP COLUMN tiers`,
 	},
+	57: {
+		`ALTER TABLE agents
+		ADD COLUMN process_exec_path VARCHAR`,
```
BupycHuk, line 688: please check that we won't have issues with long paths

BupycHuk (changes requested):
Please write a few tests. 
If we included showing exec path in front-end into the tasks then we should expose it in API.

docker-compose.yml
qwest812: I reverted it

services/converters.go
```diff
@@ -216,13 +216,18 @@ func ToAPIAgent(q *reform.Querier, agent *models.Agent) (inventorypb.Agent, erro
 		}
 		serviceID = service.ServiceID
 	}
+	var agentProcess string
+	if agent.ProcessExecPath != nil {
+		agentProcess = *agent.ProcessExecPath
+	}
```
BupycHuk, line 222: this block can be replaced by `pointer.GetString`

services/converters.go
```diff
@@ -220,9 +220,10 @@ func ToAPIAgent(q *reform.Querier, agent *models.Agent) (inventorypb.Agent, erro
 	switch agent.AgentType {
 	case models.PMMAgentType:
 		return &inventorypb.PMMAgent{
-			AgentId:      agent.AgentID,
-			RunsOnNodeId: pointer.GetString(agent.RunsOnNodeID),
-			CustomLabels: labels,
+			AgentId:         agent.AgentID,
+			RunsOnNodeId:    pointer.GetString(agent.RunsOnNodeID),
+			CustomLabels:    labels,
+			ProcessExecPath: *agent.ProcessExecPath,
```
BupycHuk, line 226: please use pointer lib to get value instead of dereference in all places

qwest812: fixed

BupycHuk (approved):

PR #989, opened 2022-01-31
PMM-9439 Added node_exporter tls config pass

tshcherban:
PMM-9439

- [submodules FB](https://github.com/Percona-Lab/pmm-submodules/pull/2290)


tshcherban:
We are bypassing that logic for node_exporter completely:
image
Is that intentionally?
With current PR I'm changing node exporter to use this code similar to another exporters.
Also another question, is there a good reason we are not using `t.Parallel()` in tests? Semms it can speedup things a bit
image
I've had a timeout issue when added some simple config tests and try to run them with `-race` enabled
@JiriCtvrtka @BupycHuk

BupycHuk:
In tests where we use DB it's better not to use `t.Parallel` in other cases fill free to add it.

services/agents/node.go
```diff
@@ -29,6 +29,35 @@ import (
...
+
+func getExporterEnv(exporter *models.Agent, pmmAgentVersion *version.Parsed) []string {
+	env := []string{}
+
+	// basic auth via env for older exporters
+	if pmmAgentVersion.Less(v2_26_99) { // <= 2.27
+		env = append(env, fmt.Sprintf("HTTP_AUTH=pmm:%s", exporter.GetAgentPassword()))
+	}
+
+	return env
```
JiriCtvrtka, line 58: What about dont create env var in beginning and in case of <= 2.27 return it immediately and in other cases just return []string{}.

models/agent_model.go
```diff
@@ -631,6 +638,18 @@ func (s Agent) TemplateDelimiters(svc *Service) *DelimiterPair {
 	return &tdp
 }
 
+func (s *Agent) buildWebConfigFile(password string) string {
+	buf, err := bcrypt.GenerateFromPassword([]byte(password), 14)
```
ritbl, line 642: `14` is magic var

models/agent_model.go
```diff
@@ -631,6 +638,18 @@ func (s Agent) TemplateDelimiters(svc *Service) *DelimiterPair {
...
+	if err != nil {
+		log.Fatal(err, "cannot encrypt basic auth password")
```
ritbl, line 644: It is not `encryption` it is hashing. `cannot calculate hash for password`

models/agent_model.go
```diff
@@ -631,6 +638,18 @@ func (s Agent) TemplateDelimiters(svc *Service) *DelimiterPair {
...
+	}
+
+	content := "basic_auth_users:" + "\n" +
```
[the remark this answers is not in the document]
tshcherban: replaced with format

services/agents/node.go
tshcherban: simplified

services/agents/node.go
```diff
@@ -128,13 +157,25 @@ func nodeExporterConfig(node *models.Node, exporter *models.Agent, agentVersion
 
 	sort.Strings(args)
 
+	env := getExporterEnv(exporter, agentVersion)
+	files := getExporterFiles(exporter, agentVersion)
+
+	for k := range files {
+		switch k {
```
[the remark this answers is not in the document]
tshcherban: cant be sure about future, for now simplified

models/agent_model.go
tshcherban: done

models/agent_model.go
tshcherban: currently can move it to some local const, but maybe we should introduce global configurable value for that?

ritbl (approved):
LGTM

models/agent_model.go
ritbl: I think moving to const is good for now

PR #1001, opened 2022-02-18
PMM-9439 node_exporter reforking PoC

tshcherban:
PMM-9439

Build: SUBMODULES-2290


services/agents/node.go
```diff
@@ -29,6 +29,33 @@ import (
 	"github.com/percona/pmm-managed/utils/collectors"
 )
 
+var v2_26_99 = version.MustParse("2.26.99")
```
BupycHuk, line 32: I suggest to put all logic related to some version in one place. so it's easier to manage it.

models/agent_model.go
```diff
@@ -631,6 +641,15 @@ func (s Agent) TemplateDelimiters(svc *Service) *DelimiterPair {
 	return &tdp
 }
 
+func (s *Agent) buildWebConfigFile(password string) string {
+	buf, err := bcrypt.GenerateFromPassword([]byte(password), passwordHashCost)
+	if err != nil {
+		log.Fatal(err, "cannot calculate hash for password")
+	}
+
+	return fmt.Sprintf("basic_auth_users:\n    pmm: %s\n", string(buf))
```
BupycHuk, line 650: let's use templates instead of fmt.Sprintf, so it's easier to read.

services/agents/node.go
tshcherban: you mean declarations only? because for all version checks it would be quite a lot of changes, almost every entity within that package has some backward-compatibility checks

tshcherban: resolve it temporary and move to another ticket

models/agent_model.go
tshcherban: Do you mean those templates?
`tmpl, err := template.New("test").Parse("{{.Count}} items are made of {{.Material}}")
if err != nil { panic(err) }
err = tmpl.Execute(os.Stdout, sweaters)
if err != nil { panic(err) }`

tshcherban: Or move that 
> basic_auth_users:\n    pmm: %s\n

to external file and load as a resource?

BupycHuk: please check how we use it in QAN, supervisord or scrape configs. it makes your template easier to read.
from what I see I can't understand quickly that it has some structure and might think that pmm is a value for `basic_auth_users`
```
basic_auth_users:
    pmm: {{ buf }}
```

services/agents/node.go
BupycHuk: Ok, now I remember why I wrote it.
We have multiple checks of pmmAgentVersion here in code, so it'll easier to read if we have one place with all code related to some version there.

services/agents/node_test.go
```diff
@@ -23,15 +23,98 @@ import (
...
+				"--web.listen-address=:{{ .listen_port }}",
+			},
+			Env: []string{},
+		}
+
+		require.Equal(t, expected.Env, actual.Env)
+		content, exist := actual.TextFiles["webConfigPlaceholder"]
+		if !exist {
+			t.Error("Expected 'webConfigPlaceholder' in text files")
+		}
```
BupycHuk, line 84: please use `assert` or `require`

services/agents/node_test.go
```diff
@@ -23,15 +23,98 @@ import (
...
+
+		var cfg WebConfig
+		err := yaml.Unmarshal([]byte(content), &cfg)
+		if err != nil {
+			t.Fatal(err)
+		}
```
BupycHuk, line 90: please use `require.NoError(err)`

services/agents/node_test.go
```diff
@@ -23,15 +23,98 @@ import (
...
+
+		require.NotEmpty(t, cfg.BasicAuthUsers.Pmm, "WebConfig file should contain a secret for 'pmm' user")
+
+		buff := []byte(cfg.BasicAuthUsers.Pmm)
+
+		err = bcrypt.CompareHashAndPassword(buff, []byte(exporter.AgentID))
+		if err != nil {
+			t.Fatal(err)
+		}
```
BupycHuk, line 99: please use `require.NoError`

services/agents/node_test.go
```diff
@@ -23,15 +23,98 @@ import (
...
+			TemplateLeftDelim:  "{{",
+			TemplateRightDelim: "}}",
+			Args: []string{
+				"--web.listen-address=:{{ .listen_port }}",
+			},
+			Env: []string{},
+		}
+
+		require.Equal(t, expected.Env, actual.Env)
+		content, exist := actual.TextFiles["webConfigPlaceholder"]
```
BupycHuk, line 81: let's make sure that content is always the same.

BupycHuk (changes requested):

services/agents/node.go
BupycHuk: @tshcherban this one is missed

go.mod
```diff
@@ -144,3 +144,5 @@ require (
 	gopkg.in/ini.v1 v1.66.4 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace golang.org/x/crypto v0.0.0-20220313003712-b769efc7c000 => github.com/percona-lab/crypto v0.0.0-20220413143500-70ef7e9f1813
```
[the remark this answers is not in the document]
tshcherban: approved with team

models/agent_model.go
```diff
@@ -631,6 +635,46 @@ func (s Agent) TemplateDelimiters(svc *Service) *DelimiterPair {
...
+	hashedPassword := HashPassword(password, salt)
+
+	var configBuffer bytes.Buffer
+	if tmpl, err := template.New("webConfig").Parse(webConfigTemplate); err != nil {
+		log.Fatalln(err)
+	} else if err = tmpl.Execute(&configBuffer, hashedPassword); err != nil {
+		log.Fatalln(err)
+	}
+
+	config := configBuffer.String()
```
YaroslavPodorvanov, line 665: log.Fatalln(err) contains os.Exit(1) so clean up looks like
```go
	tmpl, err := template.New("webConfig").Parse(webConfigTemplate)

	if err != nil {
		log.Fatalln(err)
	}

	var configBuffer bytes.Buffer
	if err = tmpl.Execute(&configBuffer, hashedPassword); err != nil {
		log.Fatalln(err)
	}

	config := configBuffer.String()
```

But up to you

services/agents/agents.go
```diff
@@ -61,3 +64,22 @@ func pathsBase(agentVersion *version.Parsed, tdpLeft, tdpRight string) string {
 
 	return tdpLeft + " .paths_base " + tdpRight
 }
+
+// ensureAuthParams updates agent start parameters to contain prometheus webconfig.
+func ensureAuthParams(exporter *models.Agent, params *agentpb.SetStateRequest_AgentProcess, agentVersion *version.Parsed, minAuthVersion *version.Parsed) {
+	if params.Env == nil {
+		params.Env = []string{}
+	}
```
YaroslavPodorvanov, line 72: why you do this?
```go
var env []string
env = append(env, "some)
```
work fine, no need init

BupycHuk (approved):

models/agent_model.go
tshcherban: passed error up to correct handler

models/agent_model.go
```diff
@@ -631,6 +635,49 @@ func (s Agent) TemplateDelimiters(svc *Service) *DelimiterPair {
 	return &tdp
 }
 
+// HashPassword func to calculate password hash. Public and overridable for testing purposes.
+var HashPassword = func(password, salt string) (string, error) {
```
YaroslavPodorvanov, line 639: func HashPassword (password, salt string)

services/agents/agents.go
```diff
@@ -60,3 +63,24 @@ func pathsBase(agentVersion *version.Parsed, tdpLeft, tdpRight string) string {
 
 	return tdpLeft + " .paths_base " + tdpRight
 }
+
+// ensureAuthParams updates agent start parameters to contain prometheus webconfig.
+func ensureAuthParams(exporter *models.Agent, params *agentpb.SetStateRequest_AgentProcess, agentVersion *version.Parsed, minAuthVersion *version.Parsed) error {
+	if agentVersion.Less(minAuthVersion) {
+		params.Env = append(params.Env, fmt.Sprintf("HTTP_AUTH=pmm:%s", exporter.GetAgentPassword()))
+	} else {
+		if params.TextFiles == nil {
```
YaroslavPodorvanov, line 72: move init to after `wcf, err`

YaroslavPodorvanov (approved):

JiriCtvrtka (approved):

PR #1003, opened 2022-02-22
PMM-9320 Unable to add MongoDB service due to + in the password

mihailo-chehivsky:
PMM-9320

### DIFF
--- a/models/agent_model.go
+++ b/models/agent_model.go
@@ -436,7 +436,7 @@ func (s *Agent) DSN(service *Service, dialTimeout time.Duration, database string
 			Path:     path,
 			RawQuery: q.Encode(),
 		}
-		plusPlaceholder := "__plus__" // There is a risk of error, when the password contains this string. Can change it to a more unique string.
+		const plusPlaceholder = "__plus__" // There is a risk of error, when the password contains this string. Can change it to a more unique string.
 		switch {
 		case password != "":
 			password = strings.ReplaceAll(password, "+", plusPlaceholder)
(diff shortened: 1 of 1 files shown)


models/agent_model.go
```diff
@@ -436,15 +436,18 @@ func (s *Agent) DSN(service *Service, dialTimeout time.Duration, database string
 			Path:     path,
 			RawQuery: q.Encode(),
 		}
+		plusPlaceholder := "__plus__" // There is a risk of error, when the password contains this string. Can change it to a more unique string.
```
ritbl, line 439: why not change to a constant?

ritbl:
@mihailo-chehivsky in general I don't like this workaround. is possible to escape the password?
if not I'd rather consider using `hash(hash(pwd)+salt)` instead of a password. How have you tested it?

ritbl (changes requested):
see my comments

ritbl:
@mihailo-chehivsky it is not an escape, what about other potential issues (with other characters). What about my suggestion?

mihailo-chehivsky:
> @mihailo-chehivsky it is not an escape, what about other potential issues (with other characters). What about my suggestion?
> 
> > Tested adding a new instance MongoDB using a password contains "+" on development environment.
> 
> so you run devcontainer? did you test via pmm-admin or via UI?

vi UI

mihailo-chehivsky:
@ritbl 
I guess this bug relates to "+" character only. 
Similar issue described here: https://stackoverflow.com/questions/2678551/when-should-space-be-encoded-to-plus-or-20

PR #1007, opened 2022-02-28
PMM-9391 Added out-of-the-box MySQL Replication Alerts.

PavelKhripkov:
PMM-9391

Build: SUBMODULES-2359


idoqo (approved):

artemgavrilov (approved):
LGTM. But if we have time, let's replace `go-bindata` with native embed: https://pkg.go.dev/embed

idoqo (approved):

PR #1010, opened 2022-03-08
Fix CI cache

artemgavrilov:
Fuzzdata left after tests caused error during cache process. This PR fixes it.

Original issue: https://github.com/actions/cache/issues/753

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,7 @@ jobs:
       - name: Cache
         run: |
           docker exec pmm-managed-server go clean -testcache
+          docker exec --workdir=/root/go/src/github.com/percona/pmm-managed pmm-managed-server find . -type d -name fuzzdata -exec rm -r {} +
           rm -fr ~/.cache/go-build
           docker cp pmm-managed-server:/root/.cache/go-build ~/.cache/go-build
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -216,8 +217,8 @@ jobs:
       - name: Cache
         run: |
           docker exec pmm-managed-server go clean -testcache
+          docker exec --workdir=/root/go/src/github.com/percona/pmm-managed pmm-managed-server find . -type d -name fuzzdata -exec rm -r {} +
           rm -fr ~/.cache/go-build
-          mkdir -p ~/.cache
```
denisok, line 220: why this is deleted? maybe there was some case when there is no cache dir and thus we need to create it ?

artemgavrilov: It was not present in the same step in other job. These cache steps are identical, so I can't find a reason why we need it only here.

artemgavrilov: @BupycHuk maybe you know?

BupycHuk: it's bit different from other repos, because we use docker here, that's why it is required.

BupycHuk: but looks like it's not required

BupycHuk (approved):

PR #1013, opened 2022-03-09
PMM-8434 PMM public address from env

percona-csalguero:
[PMM-8434](https://jira.percona.com/browse/PMM-8434) PMM Server Public Address could be passed as environment variable

This patch allows setting the value for the PMM public address from the environment. It will be used in cases like un K8s where when the container is started, the DNS server is already started and we know the server address.

Build: [SUBMODULES-2372](https://github.com/Percona-Lab/pmm-submodules/pull/2372)


BupycHuk:
please update PR description

JiriCtvrtka (approved):

artemgavrilov (approved):

PR #1019, opened 2022-03-18
Advisors V2

artemgavrilov:
* PMM-9469 Support multy-query checks

Build: SUBMODULES-0

### DIFF
--- a/services/checks/checks.go
+++ b/services/checks/checks.go
@@ -70,7 +70,7 @@ const (
 	prometheusSubsystem = "checks"
 
 	alertsPrefix        = "/stt/"
-	maxSupportedVersion = 1
+	maxSupportedVersion = 2
 )
 
 // pmm-agent versions with known changes in Query Actions.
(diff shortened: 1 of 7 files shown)


services/checks/checks.go
```diff
@@ -594,6 +627,11 @@ func (s *Service) minPMMAgentVersion(t check.Type) *version.Parsed {
 	case check.MongoDBGetDiagnosticData:
 		return pmmAgent2_27_0
 
+	case check.MetricsRange:
+		fallthrough
+	case check.MetricsInstant:
+		return pmmAgentAny
```
BupycHuk, line 633: looks like returning nil will be ok and work faster

services/checks/checks.go
```diff
@@ -64,6 +69,8 @@ const (
 	scriptExecutionTimeout = 5 * time.Second  // time limit for running pmm-managed-starlark
 	resultCheckInterval    = time.Second
 
+	pmmVictoriaMetricsAddress = "http://127.0.0.1:9090/prometheus"
```
BupycHuk, line 72: we have victoriaMetricsURLF in main.go, let's reuse it in case we extract VM outside of PMM in future.

services/checks/checks.go
artemgavrilov: Yep, done

BupycHuk (approved):

oksana-grishchenko (approved):

PR #1049, opened 2022-04-01
PMM-9265: Telemetry V2

ritbl:
PMM-0

Build: SUBMODULES-0


docker-compose.yml
```diff
@@ -15,6 +22,12 @@ services:
       - ENABLE_ALERTING=1
       - ENABLE_BACKUP_MANAGEMENT=1
 
+    extra_hosts:
+      - host.docker.internal:host-gateway
```
ritbl, line 26: for linux support

docker-compose.yml
```diff
@@ -26,14 +39,23 @@ services:
       memlock: 67108864
 
     ports:
-      - 127.0.0.1:80:80
-      - 443:443
+      - ${PMM_PORT_HTTP:-80}:80
+      - ${PMM_PORT_HTTPS:-443}:443
```
ritbl, line 43: Portal must run on 443, so I need have option to run PMM on different port

main.go
```diff
@@ -22,11 +22,15 @@ import (
 	"database/sql"
 	_ "expvar" // register /debug/vars
 	"fmt"
+	"github.com/joho/godotenv"
+	"github.com/percona/pmm-managed/services/config"
+	"github.com/percona/pmm-managed/services/telemetry_v2"
```
BupycHuk, line 27: please fix imports

config/telemetry/default.yml
```diff
@@ -0,0 +1,37 @@
+telemetry:
+  - id: MySQLVersion
```
BupycHuk, line 2: i thought that we sign each telemetry field, aren't we?

.dev.devcontainer.pmm-managed.yaml.example
```diff
@@ -0,0 +1,81 @@
+services:
```
BupycHuk, line 1: why do we need one more configuration file, even two?

.devcontainer/nginx/pmm.conf
```diff
@@ -0,0 +1,220 @@
+  upstream managed-grpc {
```
BupycHuk, line 1: what is the difference between this file and one in pmm-server repository? It'll make us to handle two versions of nginx config and make development environment different from a real one.

services/config/config.go
```diff
@@ -0,0 +1,113 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+package config
+
+import (
+	_ "embed"
+	"github.com/percona/pmm-managed/services/platform"
+	"github.com/percona/pmm-managed/services/telemetry"
+	"github.com/percona/pmm-managed/services/telemetry_v2"
```
BupycHuk, line 23: please fix imports

services/config/config.go
```diff
@@ -0,0 +1,113 @@
...
+func NewService() *Service {
+	l := logrus.WithField("component", "config")
+
+	return &Service{
+		l: l,
+	}
+}
+
+func (s *Service) Load() error {
+	configPath, present := os.LookupEnv(ENV_CONFIG_PATH)
```
BupycHuk, line 69: We are trying to get rid of dependency from file system, but it increases dependency.

services/config/request.go
```diff
@@ -0,0 +1,21 @@
+package config
+
+import (
+	"github.com/percona/pmm-managed/utils/saasreq"
```
BupycHuk, line 4: please fix imports everywhere. 
we should have 3 blocks:
1. go libraries
2. third party libraries
3. packages from the same repo

services/management/ia/templates_service.go
```diff
@@ -406,7 +406,8 @@ func (s *TemplatesService) downloadTemplates(ctx context.Context) ([]alert.Templ
 	}
 
 	endpoint := fmt.Sprintf("https://%s/v1/check/GetAllAlertRuleTemplates", s.host)
-	bodyBytes, err := saasreq.MakeRequest(ctx, http.MethodPost, endpoint, accessToken, nil)
+	bodyBytes, err := saasreq.MakeRequest(ctx, http.MethodPost, endpoint, accessToken, nil,
+		&saasreq.SaasRequestOptions{SkipTlsVerification: false})
```
BupycHuk, line 410: since we added this value into config, shouldn't we use it?

services/telemetry/config.go
```diff
@@ -0,0 +1,64 @@
...
+	RetryBackoffStr     string        `yaml:"retry_backoff"`
+	RetryBackoffEnv     string        `yaml:"retry_backoff_env"`
+	RetryBackoff        time.Duration `yaml:"-"`
+	SendTimeoutStr      string        `yaml:"send_timeout"`
+	SendTimeout         time.Duration `yaml:"-"`
+	RetryCount          int           `yaml:"retry_count"`
+}
+
+func (c *Config) Init() error {
+	reportingInterval, err := time.ParseDuration(c.Reporting.IntervalStr)
```
BupycHuk, line 41: it's weird that we parse IntervalStr here, but parse IntervalEnv in another place

services/telemetry_v2/config.go
```diff
@@ -0,0 +1,261 @@
...
+		result[each.Column] = each
+	}
+	return result
+}
+
+type ReportingConfig struct {
+	SkipTlsVerification bool          `yaml:"skip_tls_verification"`
+	SendOnStart         bool          `yaml:"send_on_start"`
+	IntervalStr         string        `yaml:"interval"`
+	IntervalEnv         string        `yaml:"interval_env"`
```
BupycHuk, line 109: looks like this field is ignored

services/telemetry_v2/ds_vm.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+	"github.com/sirupsen/logrus"
+	"time"
+)
+
+type dsVm struct {
```
BupycHuk, line 13: what does dsVM stands for?

services/telemetry_v2/ds_qandb_select.go
```diff
@@ -0,0 +1,56 @@
...
+		return nil, err
+	}
+	return &dsQanDbSelect{
+		l:      l,
+		config: config,
+		db:     db,
+	}, nil
+}
+
+func openQANDBConnection(dsn string, enabled bool) (*sql.DB, error) {
```
BupycHuk, line 39: let's not connect to clickhouse directly and request all required data from qan-api

services/telemetry_v2/ds_vm.go
BupycHuk: dataSourceVictoriaMetrics?
let's use full name for struct type

services/telemetry_v2/telemetry.go
```diff
@@ -0,0 +1,306 @@
...
+		var e error
+		if settings, e = models.GetSettings(tx); e != nil {
+			return e
+		}
+
+		if settings.Telemetry.Disabled {
+			return errors.New("disabled via settings")
+		}
+		if settings.Telemetry.UUID == "" {
+			settings.Telemetry.UUID, e = generateUUID()
```
BupycHuk, line 163: in case we enable both telemetries, what will happen?

services/telemetry_v2/telemetry.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			return models.SaveSettings(tx, settings)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	serverID, err := hex.DecodeString(settings.Telemetry.UUID)
```
BupycHuk, line 175: Do we really need to decode uuid?

BupycHuk (changes requested):

.dev.devcontainer.pmm-managed.yaml.example
ritbl: I run pmm-managed in docker container (devcontainer) - `.dev.devcontainer.pmm-managed.yaml`
and locally - `.dev.pmm-managed.yaml`

these are example configs

.devcontainer/nginx/pmm.conf
BupycHuk: `auth_request` is an important part of the PMM development to be sure what works and what doesn't.
`/prometheus` should be alias for `/victoriametrics`

config/telemetry/default.yml
ritbl: no, as of now we agreed to sign the whole file.

main.go
ritbl: yeah, missed that

services/config/config.go
ritbl: yeah, I totally agree. it is by design.

services/telemetry/config.go
ritbl: yeah, I need `time.Duration` not string. And I need ENV name, because it is in the spec...

services/telemetry_v2/config.go
ritbl: correct, added overrides

services/telemetry_v2/ds_qandb_select.go
ritbl: it was a requirement. we need to use separate credentials and dedicated connection

services/telemetry/config.go
BupycHuk: I mean that it's better to keep whole logic into one place

services/telemetry_v2/ds_vm.go
ritbl: > dataSourceVictoriaMetrics?
yes

> let's use full name for struct type
yeah, totally make sense

.devcontainer/nginx/pmm.conf
ritbl: it is needed for local development only, it can be used in docker-compose
```
    volumes:
      - .devcontainer/nginx/pmm.conf:/etc/nginx/conf.d/pmm.conf:ro
```

services/telemetry_v2/telemetry.go
ritbl: they are independent, it will be sent 2 time.

in telemeteryD we support V1 and V2, so it will be stored in old and new format 2 times

services/telemetry_v2/telemetry.go
ritbl: it comes from the original version...

services/telemetry_v2/telemetry.go
BupycHuk: but we might generate it twice and store it to DB twice and override each other.

services/telemetry/config.go
ritbl: These are independent services and configs, unifying it will make code a lot more complicated.
Implementation for V1 will be removed, so we will have no duplication.

services/telemetry_v2/telemetry.go
ritbl: I see, we can have race with V1 and V2. As a consequence on the first run there is a risk of having V1 with one ID and V2 with another.
I think V1 or V2 can be enabled, but not both.

services/telemetry/config.go
BupycHuk: They overwrite the same value.
https://github.com/percona/pmm-managed/pull/1049/files#diff-7bc66d203ca22000f3fa14a2f1db4d3f013baf12a3b7930f4ad52d1f82b135bbL107-L119

I don't talk about telemetry v1 and v2, I talk about updating the same value from different places.

services/telemetry_v2/telemetry.go
BupycHuk: Do we have check for safety from that?

ritbl: I removed V1, I think it is not needed.

.devcontainer/nginx/pmm.conf
BupycHuk: and we have to support that, local environment should be as close as possible to production

main.go
```diff
@@ -35,6 +36,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/joho/godotenv"
+
+	"github.com/percona/pmm-managed/services/config"
+	"github.com/percona/pmm-managed/services/telemetry"
```
BupycHuk, line 42: please fix imports, now we have 5 blocks

main.go
```diff
@@ -221,7 +226,7 @@ func runGRPCServer(ctx context.Context, deps *gRPCServerDeps) {
 	dbaasv1beta1.RegisterLogsAPIServer(gRPCServer, managementdbaas.NewLogsService(deps.db, deps.dbaasClient))
 	dbaasv1beta1.RegisterComponentsServer(gRPCServer, managementdbaas.NewComponentsService(deps.db, deps.dbaasClient, deps.versionServiceClient))
 
-	platformService, err := platform.New(deps.db, deps.supervisord, deps.grafanaClient)
+	platformService, err := platform.New(deps.db, deps.supervisord, deps.grafanaClient, cfg)
```
BupycHuk, line 229: let's create this service in `main()`, so we have consistency in service creation.

services/telemetry_v2/config.go
```diff
@@ -0,0 +1,272 @@
...
+		TrustedPublicKeys       []string             `yaml:"trusted_public_keys"`
+		trustedPublicKeysParsed []minisign.PublicKey `yaml:"-"`
+	} `yaml:"signing"`
+	telemetry    []TelemetryConfig `yaml:"-"`
+	Endpoints    EndpointsConfig   `yaml:"endpoints"`
+	SaasHostname string            `yaml:"saas_hostname"`
+	DataSources  struct {
+		VM           *DataSourceVictoriaMetrics `yaml:"VM"`
+		QANDB_SELECT *DSConfigQAN               `yaml:"QANDB_SELECT"`
+		PMMDB_SELECT *DSConfigPMMDB             `yaml:"PMMDB_SELECT"`
```
BupycHuk: these fields doesn't look good.

main.go
```diff
@@ -610,6 +632,28 @@ func main() {
...
+		pmmdb.DSN.Host = *postgresAddrF
+		pmmdb.DSN.DB = *postgresDBNameF
+		q := make(url.Values)
+		q.Set("sslmode", "disable")
+		pmmdb.DSN.Params = q.Encode()
+
+		return nil
+	}); err != nil {
+		l.Panicf("Failed to update config: %+v", err)
+	}
```
BupycHuk, line 655: it just complicates logic
```suggestion
		ds := cfg.Config.Services.Telemetry.DataSources
		pmmdb := ds.PmmDBSelect

		pmmdb.Credentials.Username = *postgresDBUsernameF
		pmmdb.Credentials.Password = *postgresDBPasswordF
		pmmdb.DSN.Scheme = "postgres" //TODO: should be configurable
		pmmdb.DSN.Host = *postgresAddrF
		pmmdb.DSN.DB = *postgresDBNameF
		q := make(url.Values)
		q.Set("sslmode", "disable")
		pmmdb.DSN.Params = q.Encode()
```

services/telemetry/config.go
```diff
@@ -0,0 +1,312 @@
...
+	} `yaml:"datasources"`
+	Reporting ReportingConfig `yaml:"reporting"`
+}
+
+// FileConfig telemetry config.
+type FileConfig struct {
+	Telemetry []Config `yaml:"telemetry"`
+}
+
+// EndpointsConfig telemetry config.
```
BupycHuk, line 61: Could you add better description, please?

services/telemetry/config.go
```diff
@@ -0,0 +1,312 @@
...
+	SaasHostname string          `yaml:"saas_hostname"` //nolint:tagliatelle
+	DataSources  struct {
+		VM          *DataSourceVictoriaMetrics `yaml:"VM"`
+		QanDBSelect *DSConfigQAN               `yaml:"QANDB_SELECT"`
+		PmmDBSelect *DSConfigPMMDB             `yaml:"PMMDB_SELECT"`
+	} `yaml:"datasources"`
+	Reporting ReportingConfig `yaml:"reporting"`
+}
+
+// FileConfig telemetry config.
```
BupycHuk, line 56: Could you add better description, please?

services/telemetry/config.go
```diff
@@ -0,0 +1,312 @@
...
+type FileConfig struct {
+	Telemetry []Config `yaml:"telemetry"`
+}
+
+// EndpointsConfig telemetry config.
+type EndpointsConfig struct {
+	Report string `yaml:"report"`
+}
+
+// ReportEndpointURL reporting endpoint URL.
```
BupycHuk, line 66:
```suggestion
// ReportEndpointURL returns reporting endpoint URL.
```

services/telemetry/config.go
BupycHuk: please improve it in cases below too.

services/telemetry/ds_vm.go
```diff
@@ -0,0 +1,104 @@
...
+	"time"
+
+	pmmv1 "github.com/percona-platform/saas/gen/telemetry/events/pmm"
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+	"github.com/sirupsen/logrus"
+)
+
+type dataSourceVictoriaMetrics struct {
```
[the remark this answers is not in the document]
BupycHuk: and let's use dataSource in other cases too for consistency

main.go
BupycHuk: in this case we don't have to pass `cfg`

BupycHuk (changes requested):

.devcontainer/nginx/pmm.conf
ritbl: okay I can remove from PR and keep it local

main.go
ritbl: it should be fixed with tools, right? `make format` doesn't change import

main.go
ritbl: I don't want to make this change now, this method is blocking goroutine will be blocked till a shutdown.
and this service needs gRPCServer as a dependency

main.go
```diff
@@ -557,6 +562,23 @@ func migrateDB(ctx context.Context, sqlDB *sql.DB, dbName, dbAddress, dbUsername
 }
 
 func main() {
+	if _, err := os.Stat(".env"); err == nil {
```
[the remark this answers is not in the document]
ritbl: no, because app can run without containers. It can drastically speed up the process

.dev.devcontainer.pmm-managed.yaml.example
BupycHuk: having multiple environments is not a good practice. We created devcontainer to have the same environment for all of us.

main.go
ritbl: it had a purpose, it needed to be executed at a specific time. as of now, it is not needed. will simplify it

services/telemetry/config.go
ritbl: To be honest, I see zero value here, it is self-explanatory. All these structs here are used to construct config.

services/telemetry/config.go
ritbl: all these comments are useless, I'd create one config struct, but I'd need to use anonymous structs everywhere.

services/telemetry/config.go
ritbl: If you are asking about these 2 places - it is intended. We take value from the config, which can be overridden if ENV var is set. To be consistent with requirements

services/telemetry/ds_vm.go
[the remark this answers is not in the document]
ritbl: sure

.dev.devcontainer.pmm-managed.yaml.example
ritbl: okay, I will remove it from the PR and leave it for my personal productivity

main.go
BupycHuk: then let's make `cfg` part of `deps`

Makefile.devcontainer
```diff
@@ -75,6 +75,7 @@ FILES = $(shell find . -type f -name '*.go')
 
 format:                         ## Format source code.
 	goimports -local github.com/percona/pmm-managed -l -w $(FILES)
+	gofumpt -l -w $(FILES)
```
artemgavrilov, line 78: Let's invoke first `gofumpt` and then `goimports`

artemgavrilov (commented):
I'm in the middle of review process. Continue reviewing.

.golangci-required.yml
```diff
@@ -27,7 +27,6 @@ linters:
     - goimports
     - govet
     - ineffassign
-    - nolintlint
```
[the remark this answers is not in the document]
ritbl: it is useless, we have noline all over the codebase

Makefile.devcontainer
ritbl: sure

cmd/pmm-sign-config/main.go
```diff
@@ -0,0 +1,72 @@
...
+package main
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+
+	"aead.dev/minisign"
```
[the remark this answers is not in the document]
ritbl: This version cannot be used in scripts, because password is asked 2 times

services/telemetry/config.go
BupycHuk: I left the comment before you moved it here. So it what I wanted 👍🏼

main.go
ritbl: ok

.golangci-required.yml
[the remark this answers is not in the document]
BupycHuk: Yeah, but nolintlint checks if we use them correctly

services/supervisord/maintail_test.go
```diff
@@ -68,8 +68,8 @@ func TestParseEvent(t *testing.T) {
 			{Time: time.Date(2019, 8, 8, 17, 9, 57, 854000000, time.UTC), Type: stopping, Program: "pmm-managed"},
 			{Time: time.Date(2019, 8, 8, 17, 9, 59, 854000000, time.UTC), Type: stopping, Program: "pmm-managed"},
 			{Time: time.Date(2019, 8, 8, 17, 10, 0, 863000000, time.UTC), Type: stopped, Program: "pmm-managed"},
-			{Time: time.Date(2019, 8, 8, 17, 10, 01, 932000000, time.UTC), Type: starting, Program: "pmm-managed"},
-			{Time: time.Date(2019, 8, 8, 17, 10, 03, 6000000, time.UTC), Type: running, Program: "pmm-managed"},
+			{Time: time.Date(2019, 8, 8, 17, 10, 0o1, 932000000, time.UTC), Type: starting, Program: "pmm-managed"},
+			{Time: time.Date(2019, 8, 8, 17, 10, 0o3, 6000000, time.UTC), Type: running, Program: "pmm-managed"},
```
BupycHuk, line 72:
```suggestion
			{Time: time.Date(2019, 8, 8, 17, 10, 1, 932000000, time.UTC), Type: starting, Program: "pmm-managed"},
			{Time: time.Date(2019, 8, 8, 17, 10, 3, 6000000, time.UTC), Type: running, Program: "pmm-managed"},
```

services/config/request.go
```diff
@@ -0,0 +1,38 @@
...
+
+package config
+
+import (
+	"net/http"
+
+	"github.com/percona/pmm-managed/utils/saasreq"
+)
+
+func (c *Config) configureSaasReqEnrichment() {
```
BupycHuk: let's keep structure and its methods in the same file.

services/config/pmm-managed.yaml
```diff
@@ -0,0 +1,44 @@
+services:
```
BupycHuk, line 1: because of deadlines we have to accept this type of configs, but once we cut release branch it should be reworked.

services/telemetry/config.go
```diff
@@ -0,0 +1,313 @@
...
+		Params string
+	} `yaml:"-"`
+	Credentials struct {
+		Username string
+		Password string
+	} `yaml:"-"`
+	SeparateCredentials struct {
+		Username string `yaml:"username"`
+		Password string `yaml:"password"`
+	} `yaml:"separate_credentials"` //nolint:tagliatelle
```
BupycHuk, line 107: we can get use only `Credentials` instead of two separate fields, we can check for `UseSeparateCredentials` in main() where we fill values for Credentials

services/telemetry/datasource_pmmdb_select.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return &dsPmmDBSelect{
+		l:      l,
+		config: config,
+		db:     db,
+	}, nil
+}
+
+func openPMMDBConnection(config DSConfigPMMDB) (*sql.DB, error) {
```
BupycHuk, line 61: looks like this function duplicates `models.OpenDB`

BupycHuk (changes requested):

services/telemetry/datasource_pmmdb_select.go
BupycHuk: as far as I see difference in two lines of code which can be overriden by passing new params or by changing them once `db` is returned.

services/telemetry/datasources.go
```diff
@@ -0,0 +1,129 @@
...
+	localCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	tx, err := db.BeginTx(localCtx, &sql.TxOptions{})
+	if err != nil {
+		return nil, err
+	}
+	// to minimize risk of modifying DB
+	defer tx.Rollback() //nolint:errcheck
+
+	rows, err := db.Query("SELECT " + config.Query) //nolint:gosec,rowserrcheck
```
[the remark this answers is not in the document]
artemgavrilov: That's good note

services/telemetry/config.go
```diff
@@ -0,0 +1,313 @@
...
+// ServiceConfig telemetry config.
+type ServiceConfig struct {
+	l                          *logrus.Entry
+	Enabled                    bool   `yaml:"enabled"`
+	LoadDefaults               bool   `yaml:"load_defaults"`                //nolint:tagliatelle
+	ConfigLocation             string `yaml:"config_location"`              //nolint:tagliatelle
+	DisableSigningVerification bool   `yaml:"disable_signing_verification"` //nolint:tagliatelle
+	Signing                    struct {
+		TrustedPublicKeys       []string             `yaml:"trusted_public_keys"` //nolint:tagliatelle
+		trustedPublicKeysParsed []minisign.PublicKey `yaml:"-"`
```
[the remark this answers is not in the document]
ritbl: with `Unmarshaler` interface, you will need to duplicate struct and it will involve a lot of boilerplate. Here is an [example](https://github.com/percona-platform/grafana/blob/45eb4a0adf663976269b133626d196fbd1528a75/pkg/services/ngalert/api/tooling/definitions/alertmanager.go#L478).
Additionally, you are limited with what is passed, you cannot have logger, for example.
In this case, initialization is an explicit call - `Init`.

I have checked, time.Duration can be deserialized with now. Simplified impl.

services/telemetry/config.go
ritbl: we need it to be able to provide custom credentials

services/telemetry/config.go
```diff
@@ -0,0 +1,313 @@
...
+	SeparateCredentials struct {
+		Username string `yaml:"username"`
+		Password string `yaml:"password"`
+	} `yaml:"separate_credentials"` //nolint:tagliatelle
+}
+
+// Config telemetry config.
+type Config struct {
+	ID      string `yaml:"id"`
+	Source  string `yaml:"source"`
```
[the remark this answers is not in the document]
ritbl: what custom type? kinda enum?

services/telemetry/config.go
```diff
@@ -0,0 +1,313 @@
...
+		defaultConfigBytes := []byte(defaultConfig)
+		if err := yaml.Unmarshal(defaultConfigBytes, &fileCfg); err != nil {
+			return nil, errors.Wrap(err, "cannot unmashal default config")
+		}
+		fileConfigs = append(fileConfigs, fileCfg)
+	}
+
+	if err := c.validateConfig(fileConfigs); err != nil {
+		c.l.Errorf(err.Error())
+	}
```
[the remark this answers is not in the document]
ritbl: yes

services/telemetry/config.go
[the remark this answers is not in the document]
artemgavrilov: Yep, something like this: https://github.com/percona-platform/saas/blob/90f261fe195c7caabf59875b3b97d213869966e1/pkg/check/check.go#L119-L133

services/telemetry/config.go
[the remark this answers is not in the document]
artemgavrilov: I meant this: 
```
{
...
TrustedPublicKeys PubKeys `yaml:"trusted_public_keys"`
...
}

type PubKeys []minisign.PublicKey

func (pk PubKeys) UnmarshalYAML(value *Node) error {
// Unmarshall logic
}

```


BTW, let's not do that now.

services/telemetry/config.go
```diff
@@ -0,0 +1,313 @@
...
+
+	return c.merge(fileConfigs), nil
+}
+
+func (c *ServiceConfig) merge(cfgs []FileConfig) []Config {
+	var result []Config
+	ids := make(map[string]bool)
```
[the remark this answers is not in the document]
ritbl: Don't agree, with bool a bit more clear. Simple problem - simple solution.

if I'd use map[string]struct{}
setting a value will look like

```
ids := make(map[string]struct{})
ids["<some-id>"] = struct{}{}
```

compare that to 

```
ids := make(map[string]bool)
ids["<some-id>"] = true
```

Of course we can create a set Set type alias and create methods that hide all this mechanics. Again, simple problem - simple solution.

services/config/config.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v3"
+
+	"github.com/percona/pmm-managed/services/platform"
+	"github.com/percona/pmm-managed/services/telemetry"
+)
+
+const (
+	envConfigPath     = "PERCONA_PMM_CONFIG_PATH"
+	defaultConfigPath = "/etc/percona/pmm/pmm-managed.yml"
```
denisok, line 35: who and how creates this conf file ? are there changes to pmm-update or pmm-server?

services/telemetry/config.go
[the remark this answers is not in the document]
artemgavrilov: Using empty structs is idiomatic way to create sets in go. In your code you never check actual value, so it will works the same even if you will assign `false`. Main problem with this approach that at first sight its no clear is values in the map are important or not (i.e. is it set or map of bools).

services/config/config.go
ritbl: At the moment it is an extension point that we can use in the future. At the moment it can be used while developing (e.g. in devcontainer) via docker volume mount.

The idea is to have options to override config on different levels:
1. env variable pointing to a config file
2. default location of a config file
3. embedded config

1 - highest priority

services/telemetry/config.go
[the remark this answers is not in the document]
ritbl: What you are describing is not an objective statement. But rather a personal opinion, I do agree that value here is not that important. And it is not being used. We can use `map[string]*struct{}` so that we can set `nil` as a value. But it is also not an ideal solution either, because we risk having null dereference error. Plus, it doesn't contribute to clarity.

Again, I don't want to spend days and make another 5 lines to be ideal to everyone.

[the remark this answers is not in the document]
artemgavrilov: Here is another benefit of using empty structs for sets: That's how we do in PMM.

services/telemetry/config.go
```diff
@@ -0,0 +1,313 @@
...
+	for _, cfg := range cfgs {
+		for _, each := range cfg.Telemetry {
+			_, exist := ids[each.ID]
+			if exist {
+				return errors.Errorf("telemetry config ID duplication: %s", each.ID)
+			}
+			ids[each.ID] = true
+		}
+	}
+	return nil
```
artemgavrilov, line 214: This fuction validates only duplicates, am rigth? Also it will return error on the first duplicate, and all other will not be logged. I guess we should also validate that ID, query, summary is not empty, source value is valid, etc...

services/telemetry/datasources.go
```diff
@@ -0,0 +1,129 @@
...
+
+func fetchMetricsFromDB(ctx context.Context, l *logrus.Entry, timeout time.Duration, db *sql.DB, config Config) ([]*pmmv1.ServerMetric_Metric, error) {
+	localCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	tx, err := db.BeginTx(localCtx, &sql.TxOptions{})
+	if err != nil {
+		return nil, err
+	}
+	// to minimize risk of modifying DB
+	defer tx.Rollback() //nolint:errcheck
```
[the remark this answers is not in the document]
ritbl: I know, but it can change. And it potentially could be handled by the driver.
Still, from PMM standpoint we use SQL interface, not a native one

utils/saasreq/request.go
```diff
@@ -32,22 +32,30 @@ import (
 	"github.com/percona/pmm-managed/utils/logger"
 )
 
-var dialTimeout time.Duration
+var (
+	dialTimeout time.Duration
+	// RequestEnricher extension point to modify request.
+	RequestEnricher func(method string, endpoint string, req *http.Request)
```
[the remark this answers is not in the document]
ritbl: most if these extension points were deleted, deleting this one

PR #1050, opened 2022-04-01
Platform: Add missing migrations for new connect flow

palash25:
PMM-0

Build: [FB](https://github.com/Percona-Lab/pmm-submodules/pull/2420)


models/database.go
```diff
@@ -695,6 +695,14 @@ var databaseSchema = [][]string{
 		`ALTER TABLE percona_sso_details
 			ADD COLUMN pmm_server_name VARCHAR`,
 	},
+	60: {
+		`ALTER TABLE percona_sso_details
+			DROP COLUMN client_id, 
+			DROP COLUMN client_secret,
```
BupycHuk, line 701: shouldn't we migrate value from client_id into pmm_managed_client_id? and the same for secret?

palash25: done @BupycHuk  , changed it to RENAME

oksana-grishchenko (commented):
What about Grafana configuration changes? According to [docs](https://confluence.percona.com/display/PMM/Single+Sign-On+-+PMM+Server+several+domains+support) it should use GrafanaClientId and don't use ClientSecret anymore. Will this change be introduced later?

palash25:
no i forgot about that, will add it

oksana-grishchenko (approved):

services/supervisord/supervisord.go
```diff
@@ -711,13 +711,13 @@ command =
         cfg:default.server.domain="{{ .PMMServerAddress }}"
         cfg:default.auth.generic_oauth.enabled=true
         cfg:default.auth.generic_oauth.name="Percona Account"
-        cfg:default.auth.generic_oauth.client_id="{{ .PerconaSSODetails.ClientID }}"
-        cfg:default.auth.generic_oauth.client_secret="{{ .PerconaSSODetails.ClientSecret }}"
+        cfg:default.auth.generic_oauth.client_id="{{ .PerconaSSODetails.GrafanaClientID }}"
```
BupycHuk, line 714: Do we need client_secret?

palash25: the grafana config in the new spec doesn't mention it  

https://confluence.percona.com/pages/viewpage.action?spaceKey=PMM&title=Single+Sign-On+-+PMM+Server+several+domains+support#SingleSignOnPMMServerseveraldomainssupport-Orgdchanges

and this ticket went through QA and was working fine so I guess it should be ok

BupycHuk (approved):

BupycHuk:
@palash25 what is the task number and link to FB?

palash25:
@BupycHuk this is the FB  https://github.com/Percona-Lab/pmm-submodules/pull/2420  i don't think anyone created a ticket for this

models/database.go
```diff
@@ -695,6 +695,22 @@ var databaseSchema = [][]string{
...
+			RENAME COLUMN client_secret TO pmm_managed_client_secret`,
+	},
+	62: {
+		`ALTER TABLE percona_sso_details
+			ADD COLUMN grafana_client_id VARCHAR NOT NULL`,
+	},
+	63: {
+		`ALTER TABLE percona_sso_details
+			ALTER COLUMN organization_id SET NOT NULL`,
+	},
```
palash25, line 713: added this new migration because we discovered a bug here https://github.com/percona/pmm-managed/pull/1050/files#diff-046eff8c6fd93bc9a6ace012834425228856532b375e8b7e218fea20f284a88dR131  where we were setting the organization_id field to null

when organization_id was first introduced the old flow of connect PMM had already been shipped to customers and we didn't want to break things for them so we didn't add the NOT NULL constraint to organization_id  but now since we are going to ship the new connect flow which is not going to be backwards compatible and we expect to break things i guess it wouldn't hurt to add this migration. Correct me if I am wrong @BupycHuk

oksana-grishchenko (approved):

services/platform/platform.go
```diff
@@ -367,6 +379,11 @@ func (s *Service) SearchOrganizationTickets(ctx context.Context, req *platformpb
 	h := r.Header
 	h.Add("Authorization", fmt.Sprintf("Bearer %s", userAccessToken))
 
+	// TODO remove before merging
```
artemgavrilov, line 382: ?

services/platform/platform.go
```diff
@@ -476,6 +496,11 @@ func (s *Service) SearchOrganizationEntitlements(ctx context.Context, req *platf
...
```
artemgavrilov, line 499: same

palash25:
> One question about the logs - some of them looks like debugging logs. Will they persist?

no i will remove them

artemgavrilov (approved):

models/database.go
```diff
@@ -692,8 +692,17 @@ var databaseSchema = [][]string{
 			WHERE 'mongo_db_tls_options' is not null AND jsonb_typeof(mongo_db_tls_options->'stats_collections') = 'string'`,
 	},
 	59: {
+		`DELETE FROM percona_sso_details WHERE organization_id = ''`,
```
BupycHuk, line 695: this migration makes no sense because empty organization_id won't break migration, but null in value will, so we should check for null value as well.

services/platform/platform.go
```diff
@@ -356,6 +362,10 @@ func (s *Service) SearchOrganizationTickets(ctx context.Context, req *platformpb
 		return nil, errGetSSODetailsFailed
 	}
 
+	if ssoDetails.OrganizationID == "" {
+		return nil, errOrgIDNotFound
+	}
```
BupycHuk, line 367: can we get this value from portal?

BupycHuk (changes requested):

models/database.go
palash25: changed to `IS NULL`

services/platform/platform.go
palash25: i realized we no longer need this condition since orgID is now not null so I removed it.

oksana-grishchenko (approved):

services/platform/platform.go
BupycHuk: it still can be empty value, but ok

BupycHuk (approved):

PR #1056, opened 2022-04-04
PMM-9377 upgrade to grpc-gateway v2

idoqo:
PMM-9377

grpc-gateway v2 upgrade that uses a custom error handler instead of a fork.

Build: SUBMODULES-2419

- [ ] https://github.com/percona/pmm/pull/861


BupycHuk (approved):

palash25 (approved):

api-tests/helpers.go
```diff
@@ -70,12 +70,14 @@ func AssertAPIErrorf(t TestingT, actual error, httpStatus int, grpcCode codes.Co
 	require.True(t, codeField.IsValid(), "Wrong response structure. There is no field Code in Payload.")
 	assert.Equal(t, int64(grpcCode), codeField.Int(), "gRPC status codes are not equal")
 
-	errorField := payload.Elem().FieldByName("Error")
-	require.True(t, errorField.IsValid(), "Wrong response structure. There is no field Error in Payload.")
+	errorField := payload.Elem().FieldByName("Message")
+	require.True(t, errorField.IsValid(), "Wrong response structure. There is no field Message in Payload.")
```
artemgavrilov, line 74: It's breaking change, no?

idoqo: No actually, the `Error` field is still present in the response, but it doesn't exist in the payload structure since it's autogenerated from grpc-gateway/openapi

idoqo: Yeah, it used to be.

artemgavrilov: @BupycHuk @idoqo So, that means that this is breaking change, right?

idoqo: I'm still not sure what this breaks, I think I'm missing something.

The way I understand it, these tests depend on the default HTTP clients generated by go-swagger, but the protobuf used to generate the responses no longer have the field (https://github.com/grpc-ecosystem/grpc-gateway/blob/aec6aa29864109e41408491319a859f190ec4040/protoc-gen-openapiv2/internal/genopenapi/template.go#L1326). 

Also, these responses are used in our OpenAPI spec, but not used directly by pmm-managed. The issue I see is we now have a field in actual response (`error`), that is not in the OpenAPI documentation but in this case, I think that is fine because it is not a field we want users to depend on in the first place.

palash25 (approved):

PR #1059, opened 2022-04-06
PMM-9415 fix first restore attempt never works

PavelKhripkov:
PMM-9415

Build: SUBMODULES-2402

- [ ] https://github.com/percona/pmm-agent/pull/347

When I was testing backup feature I realized I cannot use MySQL 8.0.27 with PXB 8.0.27. I changed the logic such a way it can work without need to modify the code every time new versions of MySQL and PXB are released. Now it works according to PXB documentation. Also I changed tests, because there was no negative tests and tests of extremums.


idoqo (approved):

BupycHuk (approved):
looks good to me, a few minor comments

artemgavrilov (approved):

PR #1060, opened 2022-04-10
PMM-9265: telemetry fixes and config update

ritbl:
PMM-9265

Build: SUBMODULES-0

### DIFF
--- a/services/telemetry/datasources.go
+++ b/services/telemetry/datasources.go
@@ -78,11 +78,11 @@ func (r *dataSourceRegistry) LocateTelemetryDataSource(name string) (DataSource,
 
 // DataSource telemetry data source.
 type DataSource interface {
-	FetchMetrics(ctx context.Context, config Config) ([]*pmmv1.ServerMetric_Metric, error)
+	FetchMetrics(ctx context.Context, config Config) ([][]*pmmv1.ServerMetric_Metric, error)
 	Enabled() bool
 }
 
-func fetchMetricsFromDB(ctx context.Context, l *logrus.Entry, timeout time.Duration, db *sql.DB, config Config) ([]*pmmv1.ServerMetric_Metric, error) {
+func fetchMetricsFromDB(ctx context.Context, l *logrus.Entry, timeout time.Duration, db *sql.DB, config Config) ([][]*pmmv1.ServerMetric_Metric, error) {
 	localCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 	tx, err := db.BeginTx(localCtx, &sql.TxOptions{})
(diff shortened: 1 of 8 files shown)


services/telemetry/datasources.go
```diff
@@ -92,13 +92,11 @@ func fetchMetricsFromDB(ctx context.Context, l *logrus.Entry, timeout time.Durat
 	// to minimize risk of modifying DB
 	defer tx.Rollback() //nolint:errcheck
 
-	rows, err := db.Query("SELECT " + config.Query) //nolint:gosec,rowserrcheck,sqlclosecheck
+	rows, err := tx.Query("SELECT " + config.Query) //nolint:gosec,rowserrcheck,sqlclosecheck
```
ritbl, line 95: Horrible mistake causing timeouts for each call

services/telemetry/telemetry.go
```diff
@@ -161,22 +161,38 @@ func (s *Service) prepareReport(ctx context.Context) (*reporter.ReportRequest, e
 			continue
 		}
 
+		// fetch metrics from DS
+		metricFetchStartTime := time.Now()
 		metrics, err := ds.FetchMetrics(ctx, telemetry)
+		metricFetchTook := time.Since(metricFetchStartTime)
+		s.l.Debugf("fetching [%s] took [%s]", telemetry.ID, metricFetchTook)
+		totalTime += metricFetchTook
```
artemgavrilov, line 169: Not for now: I think it can be just another PMM metric.

artemgavrilov (approved):

services/telemetry/datasources.go
```diff
@@ -109,19 +107,33 @@ func fetchMetricsFromDB(ctx context.Context, l *logrus.Entry, timeout time.Durat
...
 		}
 
+		var metric []*pmmv1.ServerMetric_Metric
 		for idx, column := range columns {
-			if _, ok := cfgColumns[column]; ok {
-				metrics = append(metrics, &pmmv1.ServerMetric_Metric{
-					Key:   column,
-					Value: *strs[idx],
-				})
+			var value string
```
BupycHuk, line 120:
```suggestion
			value := pointer.GetString(strs[idx])
```
it looks simpler

services/telemetry/telemetry.go
```diff
@@ -161,22 +161,38 @@ func (s *Service) prepareReport(ctx context.Context) (*reporter.ReportRequest, e
...
 			s.l.Debugf("failed to extract metric from datasource for [%s]:[%s]: %v", telemetry.Source, telemetry.ID, err)
 			continue
 		}
 
-		current.Metrics = append(current.Metrics, metrics...)
+		for _, each := range metrics {
+			telemetryMetric, err := s.makeMetric(ctx)
+			if err != nil {
+				s.l.Debugf("failed to make Metric %v", err)
+				continue telemetryLoop
```
BupycHuk, line 179: it can be just `break`

BupycHuk (approved):
Looks good to me, just a few minor comments.

services/telemetry/telemetry.go
ritbl: in need to jump to outer loop, break will not work here

BupycHuk: there are no operations after inner loop, so break will do the same.

PR #1065, opened 2022-04-12
PMM-9764 support checks list filtering by categories

idoqo:
PMM-9764

Build: SUBMODULES-2458

- [ ] https://github.com/percona/pmm/pull/864
- [ ] https://github.com/percona-platform/platform/pull/437
- [ ] https://github.com/percona-platform/saas/pull/64


services/management/checks.go
```diff
@@ -212,24 +212,62 @@ func (s *ChecksAPIService) ListSecurityChecks(ctx context.Context, req *manageme
...
-			Interval:    convertInterval(c.Interval),
-		})
+	checks := make([]*managementpb.SecurityCheck, 0, len(results))
+	if req != nil && req.FilterParams != nil {
+		if req.FilterParams.Category != nil {
+			// we only support filtering by category for now.
+			// later, other "keys" can be supported by "AND-ing" the filters i.e., the response from each filter
+			// can be passed to the next.
+			filterValues := req.FilterParams.Category.Value
+			switch filterValues.(type) {
```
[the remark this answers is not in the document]
PavelKhripkov: Actually we can follow the advice.

services/management/checks.go
```diff
@@ -212,24 +212,62 @@ func (s *ChecksAPIService) ListSecurityChecks(ctx context.Context, req *manageme
...
 
-	return &managementpb.ListSecurityChecksResponse{Checks: res}, nil
+	return &managementpb.ListSecurityChecksResponse{Checks: checks}, nil
+}
+
+func filterByCategory(checks map[string]check.Check, disabledChecks map[string]struct{}, categories []string) []*managementpb.SecurityCheck {
+	res := make([]*managementpb.SecurityCheck, 0, len(checks))
+
+	var disabled bool
+	for _, category := range categories {
```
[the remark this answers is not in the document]
idoqo: Yeah, just that I'm not sure we can gain much from that. Because we will end up processing all the checks (in all categories) only to later select the few categories we need. 

I think we cut that down with the current approach, as it ignores checks that don't match with the if check.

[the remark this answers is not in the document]
BupycHuk: Agree with @PavelKhripkov,
He means to do the same that was done for disabledChecks for categories: create map where key is category and value is empty struct. So you will be able just check if category of check exist in the map. And in case of more filters it will be simpler to add them too.

services/management/checks.go
```diff
@@ -212,24 +212,60 @@ func (s *ChecksAPIService) ListSecurityChecks(ctx context.Context, req *manageme
...
-		res = append(res, &managementpb.SecurityCheck{
-			Name:        c.Name,
-			Disabled:    disabled,
-			Summary:     c.Summary,
-			Description: c.Description,
-			Interval:    convertInterval(c.Interval),
-		})
+	checks := make([]*managementpb.SecurityCheck, 0, len(results))
+	if req != nil && req.FilterParams != nil {
+		if req.FilterParams.Category != nil {
```
BupycHuk, line 222: in case of FilterParams passed, but categories isn't passed in request we will return empty result. I'm not sure if it's a good idea.

services/management/checks.go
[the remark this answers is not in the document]
idoqo: Oh, got it.

services/management/checks.go
```diff
@@ -212,24 +212,64 @@ func (s *ChecksAPIService) ListSecurityChecks(ctx context.Context, req *manageme
...
+			res = append(res, &managementpb.SecurityCheck{
+				Name:        ch.Name,
+				Disabled:    disabled,
+				Summary:     ch.Summary,
+				Description: ch.Description,
+				Interval:    convertInterval(ch.Interval),
+				Category:    ch.Category,
+			})
+		}
+	}
```
BupycHuk, line 271: looks like duplication of this and block above https://github.com/percona/pmm-managed/pull/1065/files#diff-ae29f66284415f440fd7f8446d2c0c11735f3807133a9b2d9889c9a1a40d0101R235-R245 can be simplified using just by adding `or`.
```suggestion
	for _, ch := range checks {
		if _, ok := cm[ch.Category]; ok || len(cm) == 0 {
			_, disabled = disabledChecks[ch.Name]
			res = append(res, &managementpb.SecurityCheck{
				Name:        ch.Name,
				Disabled:    disabled,
				Summary:     ch.Summary,
				Description: ch.Description,
				Interval:    convertInterval(ch.Interval),
				Category:    ch.Category,
			})
		}
	}
```

idoqo: Yeah, we can either skip it or call it always (whether there's a category filter or not) and both will work.

I think skipping makes it clearer which filter gets applied or not.

BupycHuk: In other endpoints we return all values in cases of empty filters.
Even if we skip the function we can simplify using the same code style to not have duplication.
Could you write API tests for it?

api-tests/server/checks_test.go
```diff
@@ -84,15 +85,39 @@ func TestListSecurityChecks(t *testing.T) {
...
+			},
+		})
+		require.NoError(t, err)
+		assert.NotNil(t, resp)
+		assert.NotEmpty(t, resp.Payload.Checks)
+		for _, c := range resp.Payload.Checks {
+			assert.NotEmpty(t, c.Name, "%+v", c)
+			assert.NotEmpty(t, c.Summary, "%+v", c)
+			assert.NotEmpty(t, c.Description, "%+v", c)
+			assert.Equal(t, c.Category, "performance")
```
PavelKhripkov, line 106: I think we should add some checks with `category` field. Otherwise we get empty list to assert.

api-tests/server/checks_test.go
```diff
@@ -84,15 +85,39 @@ func TestListSecurityChecks(t *testing.T) {
...
-	require.NoError(t, err)
-	assert.NotNil(t, resp)
-	assert.NotEmpty(t, resp.Payload.Checks)
-	for _, c := range resp.Payload.Checks {
-		assert.NotEmpty(t, c.Name, "%+v", c)
-		assert.NotEmpty(t, c.Summary, "%+v", c)
-		assert.NotEmpty(t, c.Description, "%+v", c)
-	}
+	t.Run("with categories filter", func(t *testing.T) {
+		filter := `{"filter_params": {"category": {"string_values": {"values": ["performance"]}}}}'`
```
PavelKhripkov, line 89:
```suggestion
		filter := `{"filter_params": {"category": {"string_values": {"values": ["performance"]}}}}`
```

api-tests/server/checks_test.go
```diff
@@ -84,15 +85,41 @@ func TestListSecurityChecks(t *testing.T) {
...
-	require.NoError(t, err)
-	assert.NotNil(t, resp)
-	assert.NotEmpty(t, resp.Payload.Checks)
-	for _, c := range resp.Payload.Checks {
-		assert.NotEmpty(t, c.Name, "%+v", c)
-		assert.NotEmpty(t, c.Summary, "%+v", c)
-		assert.NotEmpty(t, c.Description, "%+v", c)
-	}
+	t.Run("with categories filter", func(t *testing.T) {
+		filter := `{"filter_params": {"category": {"string_values": {"values": ["performance"]}}}}`
```
PavelKhripkov, line 89:
```suggestion
		filter := `{"category": {"string_values": {"values": ["performance"]}}}`
```

api-tests/server/checks_test.go
```diff
@@ -84,15 +85,41 @@ func TestListSecurityChecks(t *testing.T) {
...
+		filterParams := &security_checks.ListSecurityChecksParamsBodyFilterParams{}
+		err := json.Unmarshal([]byte(filter), filterParams)
+		assert.NoError(t, err)
+
+		resp, err := managementClient.Default.SecurityChecks.ListSecurityChecks(&security_checks.ListSecurityChecksParams{
+			Body: security_checks.ListSecurityChecksBody{
```
PavelKhripkov, line 95:
```suggestion
			Context: pmmapitests.Context,
			Body: security_checks.ListSecurityChecksBody{
```

api-tests/server/checks_test.go
idoqo: Yeah, this is less complicated, but marshaling it to the request body will fail. `string_values` can't take arrays directly because we use `oneOf` in the `proto` definition and it doesn't allow repeated values (https://developers.google.com/protocol-buffers/docs/proto3#using_oneof),

api-tests/server/checks_test.go
```diff
@@ -84,15 +85,41 @@ func TestListSecurityChecks(t *testing.T) {
...
+		})
+		require.NoError(t, err)
+		assert.NotNil(t, resp)
+		assert.NotNil(t, resp.Payload)
+		if len(resp.Payload.Checks) != 0 {
+			for _, c := range resp.Payload.Checks {
+				assert.NotEmpty(t, c.Name, "%+v", c)
+				assert.NotEmpty(t, c.Summary, "%+v", c)
+				assert.NotEmpty(t, c.Description, "%+v", c)
+				assert.Equal(t, c.Category, "performance")
```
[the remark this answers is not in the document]
idoqo: At the moment, `string_values` can't take arrays directly because we use `oneOf` in the proto definition and it doesn't allow repeated values (https://developers.google.com/protocol-buffers/docs/proto3#using_oneof),

api-tests/server/checks_test.go
idoqo: At the moment, `string_values` can't take arrays directly because we use `oneOf` in the proto definition and it doesn't allow repeated values (https://developers.google.com/protocol-buffers/docs/proto3#using_oneof),

PavelKhripkov (approved):

BupycHuk:
please fix merge conflicts

BupycHuk (approved):

PR #1067, opened 2022-04-12
PMM-7 fix imports

BupycHuk:
PMM-7

Build: SUBMODULES-0


Makefile.devcontainer
```diff
@@ -75,6 +75,7 @@ FILES = $(shell find . -type f -name '*.go')
 
 format:                         ## Format source code.
 	gofumpt -l -w $(FILES)
+	gci write --Section Standard --Section Default --Section "Prefix(github.com/percona/pmm-managed)" .
 	goimports -local github.com/percona/pmm-managed -l -w $(FILES)
```
idoqo, line 79: Shouldn't we run `gci` after `goimports`? Not sure but I imagined gci will only order them and it makes sense to do that after goimports have fixed missing/unused imports or something.

```suggestion
        goimports -local github.com/percona/pmm-managed -l -w $(FILES)
	gci write --Section Standard --Section Default --Section "Prefix(github.com/percona/pmm-managed)" .
```

BupycHuk: good point

Makefile.devcontainer
```diff
@@ -75,7 +75,8 @@ FILES = $(shell find . -type f -name '*.go')
 
 format:                         ## Format source code.
 	gofumpt -l -w $(FILES)
-	goimports -local github.com/percona/pmm-managed -l -w $(FILES)
+        goimports -local github.com/percona/pmm-managed -l -w $(FILES)
+	gci write --Section Standard --Section Default --Section "Prefix(github.com/percona/pmm-managed)" .
```
artemgavrilov, line 79: Redundand tab space before goimports. But do we need goimports at all now?

artemgavrilov (approved):
Epic!

Makefile.devcontainer
idoqo: old comment so not if still valid, but looks like we do (https://github.com/daixiang0/gci/issues/1#issuecomment-668305417)

idoqo (approved):

Makefile.devcontainer
artemgavrilov: Aha, got it! Thanks!

artemgavrilov:
@BupycHuk Let's also add `gci` to the list of required linters

PR #1079, opened 2022-04-18
Changes from 2.27.0

BupycHuk:
Changes from 2.27.0


models/database.go
```diff
@@ -692,8 +692,17 @@ var databaseSchema = [][]string{
 			WHERE 'mongo_db_tls_options' is not null AND jsonb_typeof(mongo_db_tls_options->'stats_collections') = 'string'`,
 	},
 	59: {
+		`DELETE FROM percona_sso_details WHERE organization_id IS NULL`,
+	},
+	60: {
 		`ALTER TABLE percona_sso_details
-			ADD COLUMN pmm_server_name VARCHAR`,
+			RENAME COLUMN client_id TO pmm_managed_client_id;
+		ALTER TABLE percona_sso_details
```
artemgavrilov, line 700: Everything is fine here?

artemgavrilov (approved):

models/database.go
BupycHuk: doesn't look like that. it's not critical, but also is not a good approach for development.

artemgavrilov: Just to be sure: between releases we have all migrations in order. This mess is only in dev, right?

PR #1082, opened 2022-04-18
PMM-9854 refresh checks cache on platform connection

idoqo:
PMM-9854

Build: SUBMODULES-2461


services/checks/checks.go
```diff
@@ -390,8 +393,6 @@ func (s *Service) run(ctx context.Context, intervalGroup check.Interval, checkNa
 		return errors.WithStack(err)
 	}
 
-	s.CollectChecks(ctx)
```
BupycHuk, line 393: What was the reason to move it out from here?

idoqo: `runChecksLoop()` is called for each interval (standard, rare, frequent) so we are collecting all the checks three times.

Moving it out means that we only download checks once per execution. Also, now  `run` only cares about running the checks, not downloading them

BupycHuk (approved):

services/platform/platform.go
```diff
@@ -53,14 +53,21 @@ type supervisordService interface {
 	UpdateConfiguration(settings *models.Settings, ssoDetails *models.PerconaSSODetails) error
 }
 
+// type checksService is a subset of methods of checks.Service used by this package.
+// We use it instead of real type for testing and to avoid dependency cycle.
+type checksService interface {
+	CollectChecks(ctx context.Context)
+}
```
PavelKhripkov, line 60: Usually we keep interfaces in separate file `deps.go` in a package folder. Let's move there this one and other two)

services/checks/checks.go
```diff
@@ -349,8 +349,8 @@ func (s *Service) ToggleCheckAlert(ctx context.Context, alertID string, silence
 	return err
 }
 
-// runChecksGroup downloads and executes STT checks in synchronous way.
-// If intervalGroup is empty.
+// runChecksGroup downloads and executes checks that should run in the interval specified by intervalGroup.
```
PavelKhripkov, line 352:
```suggestion
// runChecksGroup downloads and executes Advisor checks that should run in the interval specified by intervalGroup.
```

PavelKhripkov (approved):

services/checks/checks.go
artemgavrilov: @idoqo But nothing changed. You still download checks on each group tick.

BupycHuk: it was reverted in this PR https://github.com/percona/pmm-managed/pull/1082/commits/b56ea0587d005eba95b17f0efde702602a35c6cd

artemgavrilov:
@idoqo What about addin `CollectChecks` to the Run function? Somewhere early, because we have 1 min sleep there and during this period checks page will be empty.

services/checks/checks.go
idoqo: Yeah yeah - forgot to update this comment but I ended up reverting it (sort of).

I realized it makes sense to re-download the checks on each interval because `Run()` only gets called when managed starts up, so if for instance, a user moves from paid to registered (or vice-versa), the checks we downloaded on startup remains in the cache.

So yeah, the only thing different now is moving `CollectChecks()` out of the run method.

BupycHuk (approved):

PR #1084, opened 2022-04-20
PMM-9890 more data in server info response

yurkovychv:
PMM-9890

Build: https://github.com/Percona-Lab/pmm-submodules/pull/2466

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/minio/minio-go/v7 v7.0.24
 	github.com/percona-platform/dbaas-api v0.0.0-20220110092915-5aacd784d472
 	github.com/percona-platform/saas v0.0.0-20220326163809-90f261fe195c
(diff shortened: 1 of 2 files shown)


services/platform/platform.go
```diff
@@ -564,18 +564,22 @@ func (s *Service) ServerInfo(ctx context.Context, req *platformpb.ServerInfoRequ
 	}
 
 	serverName := ""
+	var connectedToPortal bool
```
ritbl, line 567: nit: I'd write
`connectedToPortal := false`

but it just a personal preference. this way we are not relying no default the value (false), and just being more explicit

ritbl (approved):
LGTM

services/platform/platform.go
yurkovychv: changed, thanks

palash25 (approved):

BupycHuk (approved):

PR #1093, opened 2022-04-26
PMM-9362 Force disconnect

oksana-grishchenko:
PMM-9362 Force disconnect 

If for some reason a PMM can not obtain a user token (for default PMM admin user or when the SSO is broken), the PMM admin should be able to cleanup the PMM side of the connection - remove sso_details record and update supervisord configurations. 

Build: SUBMODULES-0

- [ ] https://github.com/percona/pmm/pull/876


oksana-grishchenko:
`CI / Test (1.18.x, ubuntu-latest, false) ` failure seems to be unrelated, I can also see it in other PRs in this repo

services/platform/platform.go
```diff
@@ -169,7 +169,7 @@ func (s *Service) Disconnect(ctx context.Context, req *platformpb.DisconnectRequ
 	err = s.disconnect(ctx, &disconnectPMMParams{
 		PMMServerID: settings.PMMServerID,
 	})
-	if err != nil {
+	if err != nil && !req.Force {
```
ritbl, line 172: probably it should be `|| req.Force`?

BupycHuk (commented):
BTW why don't we use transaction in disconnect?

services/platform/platform.go
oksana-grishchenko: I don't think so. This block restores the above deleted data.
In case of an error we need to restore it only when the request is not forced.
In case of `|| req.Force` the data will be restored even when there was no error at all, which is wrong. 

I agree it's hard to follow, but it's how the method is designed.

BupycHuk: I suggest to use transaction in `disconnect`, so it won't be so hard to follow.

ritbl: oh... I see. but that is super unclear. How about a bit of refactoring?

we can have:
1: extra variable:
```
recover := err != nil && !req.Force
if recover {
...
}
```

2: extract method:
```
if err != nil && !req.Force {
 recoverXyz() // here we need a good name, as it the scope grows
}
```

BupycHuk: Please check usage of transaction here https://github.com/percona/pmm-managed/blob/main/services/management/mysql.go#L58
It rollbacks changes in case of error.

artemgavrilov (approved):

artemgavrilov:
> BTW why don't we use transaction in disconnect?

I guess it's because disconnect is blocking network operation, and we don't wan't to hang transaction because of network problems.

services/platform/platform.go
BupycHuk: ok, sounds reasonable.

palash25 (approved):

PR #1096, opened 2022-04-26
PMM-9841 enable advisors by default

idoqo:
PMM-9841
SUBMODULES-2482

This turns on advisors by default in PMM. Instead of setting STTEnabled to true, it renames STTEnabled to STTDisabled (which is false by default). This is to avoid having to deal with boolean pointers (since we still need to detect "absent" vs "false" cases).

This doesn't need any corresponding frontend change to work but ideally, we could also update it to use the updated fields for consistency.

### DIFF
--- a/api-tests/server/settings_test.go
+++ b/api-tests/server/settings_test.go
@@ -433,7 +433,7 @@ func TestSettings(t *testing.T) {
 					assert.True(t, resg.Payload.Settings.SttEnabled)
 				})
 
-				t.Run("DisableSTTifTelemetryIsDisabled", func(t *testing.T) {
+				t.Run("DisableTelemetryWhileSTTEnabled", func(t *testing.T) {
 					res, err := serverClient.Default.Server.ChangeSettings(&server.ChangeSettingsParams{
 						Body: server.ChangeSettingsBody{
 							DisableTelemetry: true,
@@ -441,7 +441,7 @@ func TestSettings(t *testing.T) {
 						Context: pmmapitests.Context,
 					})
 					require.NoError(t, err)
-					assert.False(t, res.Payload.Settings.SttEnabled)
+					assert.True(t, res.Payload.Settings.SttEnabled)
 					assert.False(t, res.Payload.Settings.TelemetryEnabled)
 				})
 			})
@@ -462,7 +462,7 @@ func TestSettings(t *testing.T) {
 				resg, err := serverClient.Default.Server.GetSettings(nil)
 				require.NoError(t, err)
 				assert.False(t, resg.Payload.Settings.TelemetryEnabled)
-				assert.False(t, resg.Payload.Settings.SttEnabled)
+				assert.True(t, resg.Payload.Settings.SttEnabled)
 
 				t.Run("EnableSTTWhileTelemetryDisabled", func(t *testing.T) {
 					defer restoreSettingsDefaults(t)
--- a/api-tests/server/settings_test.go
+++ b/api-tests/server/settings_test.go
@@ -44,7 +44,7 @@ func TestSettings(t *testing.T) {
 		res, err := serverClient.Default.Server.GetSettings(nil)
 		require.NoError(t, err)
 		assert.True(t, res.Payload.Settings.TelemetryEnabled)
-		assert.False(t, res.Payload.Settings.SttEnabled)
+		assert.True(t, res.Payload.Settings.SttEnabled)
 		expected := &server.GetSettingsOKBodySettingsMetricsResolutions{
 			Hr: "5s",
 			Mr: "10s",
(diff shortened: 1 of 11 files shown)


models/settings.go
```diff
@@ -33,8 +33,8 @@ type MetricsResolutions struct {
 
 // SaaS contains settings related to the SaaS platform.
 type SaaS struct {
-	// Security Threat Tool enabled
-	STTEnabled bool `json:"stt_enabled"`
+	// Advisor checks disabled, false by default.
+	STTDisabled bool `json:"stt_disabled"`
```
artemgavrilov, line 37: We can't change it like this. We store this object in PMM internal DB and once user will update his PMM advisors will be enabled automatically. But ticket says that we should enable advisors only for the fresh installations.

artemgavrilov (changes requested):

models/settings.go
artemgavrilov: I think what we need is just to change default value that we have for `STTEnabled` field.

idoqo: Yeah, Donnie suggested in the comments to enable it for all installations while keeping the same steps to disable it.

Actually it seems we can't just change default for `STTEnabled`. Problem is that we call `fillDefauts` multiple times and setting `STTEnabled` to true there will make it impossible to disable Checks. We can work around that by using `*bool` instead of `bool` but that introduces more complexity imo.

artemgavrilov:
```
UPDATE settings SET settings = settings #- '{sass, stt_enabled}';
```

artemgavrilov (approved):

oksana-grishchenko (approved):

services/checks/checks_test.go
```diff
@@ -687,7 +664,7 @@ func TestGetFailedChecks(t *testing.T) {
 		settings, err := models.GetSettings(db)
 		require.NoError(t, err)
 
-		settings.SaaS.STTEnabled = true
+		settings.SaaS.STTDisabled = false
 		err = models.SaveSettings(db, settings)
 		require.NoError(t, err)
```
BupycHuk, line 669: probably this block can be removed.

services/checks/checks_test.go
```diff
@@ -750,14 +727,36 @@ func TestGetFailedChecks(t *testing.T) {
...
```
BupycHuk, line 732: and this one

api-tests/server/settings_test.go
```diff
@@ -433,15 +433,15 @@ func TestSettings(t *testing.T) {
 					assert.True(t, resg.Payload.Settings.SttEnabled)
 				})
 
-				t.Run("DisableTelemetryWhileSTTEnabled", func(t *testing.T) {
+				t.Run("DisableSTTifTelemetryIsDisabled", func(t *testing.T) {
```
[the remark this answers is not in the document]
artemgavrilov: I wonder why @idoqo  changed it 🤔. I'm reverting it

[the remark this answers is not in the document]
idoqo: IIRC, we mention in the docs and on the UI that checks download won't work with telemetry disabled.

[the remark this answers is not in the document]
artemgavrilov: Downloading - yes, and we have check for that there. But runtime itself should work.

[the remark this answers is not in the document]
artemgavrilov: https://github.com/percona/pmm-managed/blob/bcc3a33857617ae3b0f46f6fdc221f7430429633/services/checks/checks.go#L1030-L1039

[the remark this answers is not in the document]
BupycHuk: https://jira.percona.com/browse/PMM-8154
but we still have built-in checks.

services/checks/checks_test.go
artemgavrilov: Done

services/checks/checks_test.go
artemgavrilov: Removed

api-tests/server/settings_test.go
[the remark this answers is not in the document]
artemgavrilov: Reverted

models/settings_helpers_test.go
```diff
@@ -231,13 +231,14 @@ func TestSettings(t *testing.T) {
 			assert.True(t, errors.As(err, &errInvalidArgument))
 			assert.EqualError(t, err, `invalid argument: both enable_stt and disable_stt are present`)
 
+			// disable both Advisor Checks and Telemetry
 			ns, err = models.UpdateSettings(sqlDB, &models.ChangeSettingsParams{
-				EnableSTT:        true,
+				DisableSTT:       true,
 				DisableTelemetry: true,
```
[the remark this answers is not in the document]
artemgavrilov: Done, I guess

PR #1131, opened 2022-05-24
PMM-10068: fix PMMServerSTTEnabled query

ritbl:
[PMM-10068](https://jira.percona.com/browse/PMM-10068)

### DIFF
--- a/services/telemetry/config.default.yml
+++ b/services/telemetry/config.default.yml
@@ -10,7 +10,7 @@ telemetry:
 
   - id: PMMServerSTTEnabled
     source: PMMDB_SELECT
-    query: stt->'stt_enabled' AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
+    query: (CASE WHEN stt->'stt_disabled' = 'false' THEN 'true' ELSE 'false' END) AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
     summary: "PMM Server Security Thread Tool feature enabled/disabled"
     data:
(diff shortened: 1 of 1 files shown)


services/telemetry/config.default.yml
```diff
@@ -32,14 +32,6 @@ telemetry:
       - metric_name: "dbaas_enabled"
         column: "dbaas_enabled"
 
-  - id: PMMServerDBaaSEnabled
```
ritbl, line 35: I see that we have a duplicate

services/telemetry/config.default.yml
```diff
@@ -10,7 +10,7 @@ telemetry:
 
   - id: PMMServerSTTEnabled
     source: PMMDB_SELECT
-    query: stt->'stt_enabled' AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
+    query: (CASE WHEN stt->'stt_disabled' = 'false' THEN 'true' ELSE 'false' END) AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
```
ritbl, line 13: stt_enabled no longer exist

idoqo (approved):

pkadej (approved):

services/telemetry/config.default.yml
BupycHuk: can it be simplified like below?
```suggestion
    query: stt->'stt_disabled' != 'false' AS stt_enabled FROM settings s, jsonb_extract_path(s.settings, 'sass') AS stt
```

PR #1136, opened 2022-05-26
PMM-10078 Extract portal client, add dev env variables

artemgavrilov:
PMM-10078

Build: SUBMODULES-0


utils/envvars/parser.go
```diff
@@ -32,13 +32,14 @@ import (
...
+	defaultPlatformAddress = "https://check.percona.com"
+	envPlatformAddress     = "PERCONA_TEST_PLATFORM_ADDRESS"
+	envPlatformInsecure    = "PERCONA_TEST_PLATFORM_INSECURE"
+	envPlatformPublicKey   = "PERCONA_TEST_PLATFORM_PUBLIC_KEY"
 	// TODO REMOVE PERCONA_TEST_DBAAS IN FUTURE RELEASES.
 	envTestDbaas              = "PERCONA_TEST_DBAAS"
 	envEnableDbaas            = "ENABLE_DBAAS"
-	envPlatfromAPITimeout     = "PERCONA_PLATFORM_API_TIMEOUT"
+	envPlatformAPITimeout     = "PERCONA_PLATFORM_API_TIMEOUT"
 	defaultPlatformAPITimeout = 30 * time.Second
```
artemgavrilov, line 43: Here is new variables

utils/platform/client.go
```diff
@@ -0,0 +1,400 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+	"gopkg.in/reform.v1"
+
+	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/utils/envvars"
+	"github.com/percona/pmm-managed/utils/signatures"
+)
+
+// Client is HTTP Percona Platform client.
+// TODO: Replace this client with generated one https://jira.percona.com/browse/SAAS-956
```
artemgavrilov, line 48: https://jira.percona.com/browse/SAAS-956

services/checks/checks_test.go
```diff
@@ -57,20 +58,22 @@ func TestDownloadChecks(t *testing.T) {
 
 	sqlDB := testdb.Open(t, models.SkipFixtures, nil)
 	db := reform.NewDB(sqlDB, postgresql.Dialect, nil)
+	portalClient, err := platform.NewClient(db)
+	require.NoError(t, err)
+	portalClient.SetAddress(devPortalAddress)
+	portalClient.SetPublicKeys([]string{devPortalPublicKey})
```
BupycHuk, line 64: feels like these two parameters could be part of the constructor.

services/management/ia/templates_service.go
```diff
@@ -84,22 +82,23 @@ type TemplatesService struct {
...
 	if err != nil {
 		return nil, err
 	}
 
 	s := &TemplatesService{
 		db:                db,
 		l:                 l,
+		portalClient:      portalClient,
 		userTemplatesPath: templatesDir,
 		host:              host,
```
BupycHuk, line 103: looks like `host` is not used anywhere else.

BupycHuk: the same for `publicKeys`

services/platform/platform.go
```diff
@@ -586,62 +336,30 @@ func (s *Service) GetContactInformation(ctx context.Context, req *platformpb.Get
 		return nil, status.Error(codes.Aborted, "PMM server is not connected to Portal")
 	}
 
-	endpoint := fmt.Sprintf("https://%s/v1/orgs/%s", s.host, ssoDetails.OrganizationID)
+	resp, err := s.client.GetContactInformation(ctx, accessToken, ssoDetails.OrganizationID)
```
BupycHuk, line 339: error is ignored here

BupycHuk (approved):
Great improvement

services/management/ia/templates_service.go
artemgavrilov: Removed

services/checks/checks_test.go
artemgavrilov: The idea behind that is these methods needed only for tests.

oksana-grishchenko (approved):

services/checks/checks_test.go
```diff
@@ -38,12 +38,13 @@ import (
...
 	"github.com/percona/pmm-managed/services"
+	"github.com/percona/pmm-managed/utils/platform"
 	"github.com/percona/pmm-managed/utils/testdb"
 )
 
 const (
-	devChecksHost      = "check-dev.percona.com"
-	devChecksPublicKey = "RWTg+ZmCCjt7O8eWeAmTLAqW+1ozUbpRSKSwNTmO+exlS5KEIPYWuYdX"
+	devPortalAddress   = "https://check-dev.percona.com"
+	devPortalPublicKey = "RWTg+ZmCCjt7O8eWeAmTLAqW+1ozUbpRSKSwNTmO+exlS5KEIPYWuYdX"
```
Oneiroi, line 47: Keys should not be present in source, could you please lmk the reason this is present ?

services/management/ia/templates_service_test.go
```diff
@@ -29,11 +29,12 @@ import (
...
 
 	"github.com/percona/pmm-managed/models"
+	"github.com/percona/pmm-managed/utils/platform"
 	"github.com/percona/pmm-managed/utils/testdb"
 )
 
 const (
-	devPortalHost      = "check-dev.percona.com"
+	devPortalAddress   = "https://check-dev.percona.com"
 	devPortalPublicKey = "RWTg+ZmCCjt7O8eWeAmTLAqW+1ozUbpRSKSwNTmO+exlS5KEIPYWuYdX"
```
Oneiroi, line 38: Another key in the source, please lmk why this is present, thank you.

services/checks/checks_test.go
artemgavrilov: We have hardcoded public Percona Platform keys that we use for signatures verification of files that we receive from Portal. https://github.com/percona/pmm-managed/blob/05124a1d4ffae221ddd9d974e2c9f5d926f4afb2/utils/signatures/signatures.go#L27-L31

In tests we want to use dev environment instead of production, so we overriding prod keys with dev key. That let everyone to run this tests and check that everything is fine.

docker-compose.yml
```diff
@@ -14,7 +14,9 @@ services:
       - AWS_SECRET_KEY=${AWS_SECRET_KEY}
       - ENABLE_ALERTING=1
       - ENABLE_BACKUP_MANAGEMENT=1
-#      - PERCONA_TEST_SAAS_HOST=check.localhost
+#      - PERCONA_TEST_PLATFORM_ADDRESS=https://check.localhost
+#      - PERCONA_TEST_PLATFORM_INSECURE=1
+#      - PERCONA_TEST_PLATFORM_PUBLIC_KEY=<public key>
```
ritbl, line 19: nit: it is not clear if it is a PATH to a key

services/checks/checks.go
```diff
@@ -1422,40 +1407,12 @@ func (s *Service) downloadChecks(ctx context.Context) ([]check.Check, error) {
 		return nil, nil
 	}
 
-	s.l.Infof("Downloading checks from %s ...", s.host)
-
-	nCtx, cancel := context.WithTimeout(ctx, platformRequestTimeout)
+	nCtx, cancel := context.WithTimeout(ctx, portalRequestTimeout)
```
ritbl, line 1410: nit: naming of `nCtx` is not clear, what about `portalRequestCtx`?

docker-compose.yml
artemgavrilov: It's not a path, it's key itself

services/checks/checks.go
artemgavrilov: This a common naming in PMM projects, also scope of this variable is just few lines. I'm not sure that long var name will bring any clarity here.

ritbl: what does `n` stands for? what if we have more contexts? it will become n2Ctx?

artemgavrilov: Just `new context`. If you have several contexts of course you should use more meaningful names. But when your variable lives just few lines of code better to use short variable names.

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

utils/envvars/parser.go
```diff
@@ -32,13 +32,14 @@ import (
 )
 
 const (
-	defaultSaaSHost = "check.percona.com"
-	envSaaSHost     = "PERCONA_TEST_SAAS_HOST"
```
ritbl, line 36: Should it be backward compatible? ie we have new env var, but old one should be supported as well?
Otherwise, it will be hard to test, and it just a good practice - to have grace period and warn when old feature is used.

utils/platform/client.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// SetPublicKeys overrides Percona Platform public keys.
+func (c *Client) SetPublicKeys(keys []string) {
+	c.publicKeys = keys
+}
+
+// GetChecks download checks from Percona Platform. It also validates content and checks signatures.
+func (c *Client) GetChecks(ctx context.Context) ([]check.Check, error) {
+	const path = "/v1/check/GetAllChecks"
```
ritbl, line 102: how about removing to `endpoint`?

docker-compose.yml
ritbl: can you show example of how it is supposed to be used? the key is multiline. the user should cat it can send it?
I think, it can be a potential source of issues. ie when user set value in terminal (escaping and new line issues).

utils/envvars/parser.go
artemgavrilov: `PERCONA_TEST_*` variables are completely for internal usage. No need to have backward compatibility for them, we just replacing them at all places an once.

docker-compose.yml
artemgavrilov: It's minisign public key, it's one line, here is example:
https://github.com/percona/pmm-managed/blob/05124a1d4ffae221ddd9d974e2c9f5d926f4afb2/services/checks/checks_test.go#L46


This variable should not be used by anyone outside Percona. We print warnings in logs each time `PERCONA_TEST_*` variable used. It's completely for dev/qa needs.

ritbl: Imagine user creates `.env`, and sets env var this way:
      - PERCONA_TEST_PLATFORM_PUBLIC_KEY=${PERCONA_TEST_PLATFORM_PUBLIC_KEY}

how .env should look like?

image

artemgavrilov: Here is how it will look like for our dev environment:
```
PERCONA_TEST_PLATFORM_PUBLIC_KEY=RWTg+ZmCCjt7O8eWeAmTLAqW+1ozUbpRSKSwNTmO+exlS5KEIPYWuYdX
```

ritbl: > It's minisign public key, it's one line, here is example:
oh, it is very confusing, because it can be though as https cert
can you pls rename so that it will clear?

ritbl: `PUBLIC_KEY` part is confusing, can we rename it according to it's function? ie `PERCONA_TEST_PLATFORM_ENCRYPTION_KEY`

services/checks/checks_test.go
Oneiroi: This really needs to move to a mechanism for key management, hard-coding them in place, is fine for now, being as they are public keys based on your inclusion above, having multiple such keys is going to lead to manageability issues if nothing else.

Oneiroi (approved):
I've raised concerns over the inclusion of hard-coded keys; as these have been confirmed to only be the public key component this is fine to do. for now, however given there are multiple such keys in use (which appears to be dependant on version) there will soon be a need for a better key management approach, to avoid many such keys being present.

services/checks/checks_test.go
artemgavrilov: BTW refactored

services/checks/checks_test.go
```diff
@@ -211,7 +223,12 @@ func TestDisableChecks(t *testing.T) {
 	t.Run("disable same check twice", func(t *testing.T) {
 		sqlDB := testdb.Open(t, models.SkipFixtures, nil)
 		db := reform.NewDB(sqlDB, postgresql.Dialect, nil)
-		s, err := New(nil, nil, db, vmAddress)
+		portalClient, err := platform.NewClient(db)
+		require.NoError(t, err)
+		portalClient.SetAddress(devPortalAddress)
+		portalClient.SetPublicKeys([]string{devPortalPublicKey})
+
+		s, err := New(db, portalClient, nil, nil, vmAddress)
```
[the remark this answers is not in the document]
artemgavrilov: Done

docker-compose.yml
artemgavrilov: In fact it's not a new variable, it was `PERCONA_TEST_CHECKS_PUBLIC_KEY` for a long time. I renamed it to `PERCONA_TEST_PLATFORM_PUBLIC_KEY` because now it used not only for checks system but also for alerting.

It's _public_ key for signatures verification, not for encryption/decryption. May be `PERCONA_TEST_PLATFORM_VERIFICATION_KEY` will work better. BTW I will create a poll

PR #1149, opened 2022-06-13
api tests: Connect to real DB for tests

ShashankSinha252:
PMM-9029

Build: SUBMODULES-2565

### DIFF
--- a/api-tests-unmocked/api_poc_test.go
+++ b/api-tests-unmocked/api_poc_test.go
@@ -1,0 +1,300 @@
+// pmm-managed
+// Copyright (C) 2017 Percona LLC
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


api-tests-unmocked/api_poc_test.go
```diff
@@ -0,0 +1,295 @@
+// pmm-managed
```
BupycHuk, line 1: can we include this test into api-tests directory?

api-tests-unmocked/api_poc_test.go
```diff
@@ -0,0 +1,295 @@
...
+		logrus.Fatal("Agent response is nil")
+	}
+
+	agentCount := len(agentListResponse.Payload.PMMAgent)
+	if agentCount != 2 {
+		logrus.Fatalf("Expected only 2 PMM agents, found %d", agentCount)
+	}
+
+	for _, pmmAgent := range agentListResponse.Payload.PMMAgent {
+		if pmmAgent.AgentID == "pmm-server" {
```
BupycHuk, line 225: let's also check that this pmm-agent is connected.

api-tests-unmocked/api_poc_test.go
BupycHuk: with that we will be able to remove a lot of duplications

BupycHuk:
API tests are failing with
```
make: *** No rule to make target 'init'.  Stop.
```

